PR #21, opened 2024-04-11
Adding Authentication

gbenhaim:
1. Authenticate requests using oauth2-proxy (currently configured to use Keycloak since this is what our UI is using).

2. Since oauth2-proxy doesn't know how to parse the "Sec-Websocket-Protocol" header, transform the header to a standard bearer token using a lua function.

3. In order to use lua (which provides the base64 decode method), the ubi nginx image was changed to an openresty image. It provides nginx with the support of lua.

4. The change to the openresty image required some change to the deployment manifest of nginx.

5. Use the email address information of the logged user, as returned from oauth2-proxy for impersonation, registration and workspace listing.

6. Print nginx error log to STDERR for better debugging.

openresty:

- https://github.com/openresty/docker-openresty
- https://github.com/openresty/lua-nginx-module

oauth2-proxy:

- https://oauth2-proxy.github.io/oauth2-proxy/
- https://developer.okta.com/blog/2022/07/14/add-auth-to-any-app-with-oauth2-proxy

### DIFF
--- a/konflux-ci/ui/core/proxy/nginx.conf
+++ b/konflux-ci/ui/core/proxy/nginx.conf
@@ -12,6 +12,7 @@ events {
 http {
     log_format upstreamlog '[$time_local] $remote_addr - $remote_user - $server_name $host to: $proxy_host  $upstream_addr: $request $status upstream_response_time $upstream_response_time msec $msec request_time $request_time';
     access_log /dev/stderr upstreamlog;
+    error_log /dev/stderr;
 
     sendfile on;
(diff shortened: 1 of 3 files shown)


yftacherzog (commented):
Do we need to change the example users we have so that they provide the email which is mentioned as variable?

gbenhaim:
> Do we need to change the example users we have so that they provide the email which is mentioned as variable?

done

yftacherzog (approved):

Issue #59, opened 2024-05-12
Verify software dependencies in beginning of deployment scripts


yftacherzog:
Verify software dependencies (e.g. `openssl`) in beginning of deployment scripts and fail if not.

amisstea:
This comment is a test of the auto sync process from GitHub to Jira for an issue that's been moved between projects. It should be replicated as a comment on the jira issue.

PR #70, opened 2024-05-13
Verify software requirements in beginning of deployment scripts

avi-biton:
#59

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -3,10 +3,37 @@
 script_path="$(dirname -- "${BASH_SOURCE[0]}")"
 
 main() {
+    echo "Checking requirements" >&2
+    check_req
     echo "Deploying Konflux Dependencies" >&2
     deploy
     echo "Waiting for the dependencies to be ready" >&2
     "${script_path}/wait-for-all.sh"
+}
+
+check_req(){
+    # declare the requirements
+    local requirements=(kubectl openssl)
+    local uninstalled_requirements=()
+
+    # check if requirements are installed
+    for i in ${requirements[@]}; do
+        if ! command -v $i &> /dev/null; then
+                uninstalled_requirements+=("$i")
+        else
+                echo -e "$i is installed"
+        fi  
+    done
+
+    if (( ${#uninstalled_requirements[@]} == 0 )); then
+        echo -e "\nAll requirements are met\nContinue"
+    else
+        echo -e "\nSome requirements are missing, please install the following requirements first:"
+        for req in ${uninstalled_requirements[@]}; do
+                echo $req
+        done
+        exit 1
+    fi
 }
 
 deploy() {


deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
+    check_req
     echo "Deploying Konflux Dependencies" >&2
     deploy
     echo "Waiting for the dependencies to be ready" >&2
     "${script_path}/wait-for-all.sh"
 }
 
+check_req(){
+    # declare the requirements
+    requirements=(kind kubectl git openssl)
```
yftacherzog: I think we should not have kind here. This script runs after the part in which kind is involved. Also, I don't think we should couple this script to kind. 

About git, is it actually needed past this point? The scripts use it?

Maybe we should test that we access to kubeconfig and to a cluster. But maybe this is for another pr.

avi-biton: I was not so sure about `kind` but added it anyway.
We can decide what are the requirements we should check during the `deploy-deps.sh` script and edit it.
I think that for now it should be: `kubectl`, `openssl` and `podman` / `docker`

avi-biton: I removed `git` and `kind` from the requirements

deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
+    do
+        if ! command -v $i &> /dev/null; then
+                uninstalled_requirements+=("$i")
+        else
+                echo -e "$i is installed"
+    fi
+    done
+
+    # check if podman or docker are installed 
+    if ! command -v podman &> /dev/null && ! command -v docker &> /dev/null; then
```
gbenhaim: why do we need to check they are installed?

deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
+    uninstalled_requirements=()
+
+    # check if requirements are installed
+    for i in ${requirements[@]}
+    do
+        if ! command -v $i &> /dev/null; then
+                uninstalled_requirements+=("$i")
+        else
+                echo -e "$i is installed"
+    fi
```
gbenhaim: wrong indentation

deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
 }
 
+check_req(){
+    # declare the requirements
+    requirements=(kubectl openssl)
+    uninstalled_requirements=()
+
+    # check if requirements are installed
+    for i in ${requirements[@]}
+    do
```
gbenhaim: can we have the `do` in the same line with the for? (I wish we had a formmater for those things).

deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
+    check_req
     echo "Deploying Konflux Dependencies" >&2
     deploy
     echo "Waiting for the dependencies to be ready" >&2
     "${script_path}/wait-for-all.sh"
 }
 
+check_req(){
+    # declare the requirements
+    requirements=(kubectl openssl)
```
gbenhaim: please add local to the array declaration

deploy-deps.sh
```diff
@@ -3,12 +3,48 @@
...
+    uninstalled_requirements=()
```
gbenhaim: please add local to the array declaration

deploy-deps.sh
avi-biton: We checked it and when we are running the kind cluster creation we already need podman or docker, so the `deploy-deps.sh` should not check for it.
I will remove it

deploy-deps.sh
avi-biton: Done

deploy-deps.sh
avi-biton: Done

deploy-deps.sh
avi-biton: Done

deploy-deps.sh
avi-biton: Done

deploy-deps.sh
avi-biton: Removed

yftacherzog (approved):

gbenhaim (approved):

PR #73, opened 2024-05-16
fix: use kubectl apply when deploying registry

### LINKED ISSUE #72: Deploy-deps should use kubectl apply for the registry

yftacherzog:
We should use kubectl instead of create, so that the script could be executed multiple times.

closes #72


avi-biton (approved):
lgtm

PR #74, opened 2024-05-19
doc: add integration tests

yftacherzog:
This change explains how to add integration tests. The pipeline used does not support custom certificates at the moment, so this change also covers moving from the local registry to a public one.

### DIFF
--- a/test/resources/demo-users/user/ns2/ec-integration-test.yaml
+++ b/test/resources/demo-users/user/ns2/ec-integration-test.yaml
@@ -1,0 +1,25 @@
+apiVersion: appstudio.redhat.com/v1beta1
+kind: IntegrationTestScenario
+metadata:
+  annotations:
+    test.appstudio.openshift.io/kind: enterprise-contract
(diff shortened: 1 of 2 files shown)


test/resources/demo-users/user/ns2/ec-integration-test.yaml
```diff
@@ -0,0 +1,25 @@
...
+  contexts:
+    - description: Application testing
+      name: application
+  params:
+    - name: PUBLIC_KEY
+      value: "k8s://tekton-pipelines/public-key"
+  resolverRef:
+    params:
+      - name: url
+        value: 'https://github.com/gbenhaim/build-definitions'
```
avi-biton, line 20: Do we want to use Gal's repository?
Or is it in the meanwhile until the fix will be added to build-definition?

yftacherzog: Yes. Just until that task gets merged to build-definitions.

PR #76, opened 2024-05-19
fix: tekton operator overwrites configs

yftacherzog:
The tekton operator seems to overwrite whatever tekton configs that are in place when it is deployed. For that reason, we should wait for those configs to appear before deploying our own configs.

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -48,13 +48,15 @@ deploy_tekton() {
     # Operator
     kubectl apply -k "${script_path}/dependencies/tekton-operator"
     retry kubectl wait --for=condition=Ready -l app=tekton-operator -n tekton-operator pod --timeout=240s
+    # Wait for the operator to create configs for the first time before applying configs
+    kubectl wait --for=condition=Ready tektonconfig/config --timeout=360s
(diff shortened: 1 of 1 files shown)


deploy-deps.sh
```diff
@@ -48,13 +48,15 @@ deploy_tekton() {
     # Operator
     kubectl apply -k "${script_path}/dependencies/tekton-operator"
     retry kubectl wait --for=condition=Ready -l app=tekton-operator -n tekton-operator pod --timeout=240s
+    # Wait for the operator to create configs for the first time before applying configs
+    kubectl wait --for=condition=Ready TektonConfig/config --timeout=360s
```
avi-biton: I tested it with `240s` for several times and it worked.
Do we want to wait this long?

yftacherzog: Just because that was the number on the original wait line. It might be that @gbenhaim had some tries in which 240 was not enough.

avi-biton:
```suggestion
    kubectl wait --for=condition=Ready tektonconfig/config --timeout=360s
```
for consistency with the second call

yftacherzog: done

gbenhaim (approved):

PR #78, opened 2024-05-20
add retry to tektonconfig wait

avi-biton:
add retry in case tektonconfig is not yet created

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -49,7 +49,7 @@ deploy_tekton() {
     kubectl apply -k "${script_path}/dependencies/tekton-operator"
     retry kubectl wait --for=condition=Ready -l app=tekton-operator -n tekton-operator pod --timeout=240s
     # Wait for the operator to create configs for the first time before applying configs
-    kubectl wait --for=condition=Ready tektonconfig/config --timeout=360s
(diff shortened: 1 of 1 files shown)


yftacherzog:
The overall wait time could be very long here - 3 x 360s.

If I understand correctly, we're doing retries not for the case in which reconciliation is not done yet, but in case the configs were not created yet (i.e. reconciliation did not start yet).

If I'm trying to do a `wait` on configs that don't exist at all, the command returns immediately:

```bash
kubectl wait --for=condition=Ready tektonconfig/configZ --timeout=360s
Error from server (NotFound): tektonconfigs.operator.tekton.dev "configZ" not found
```

So in this case the retry will move to the next iteration immediately, which will mean the waiting time will not be very long, but if the configs exist and the operator is just unable to reconcile them, then the waiting time is going to be very long, which might be frustrating.

I'm not sure what's the right approach, but given that we did not see this `wait` failing yet, I think it might be that the condition we wait on just before (`--for=condition=Ready -l app=tekton-operator`) might already cover the creation of the configs.

All in all, I'm not sure we should apply this fix before we see the issue exists.

cc @gbenhaim

gbenhaim:
@yftacherzog I don't mind waiting with this PR.

avi-biton:
Closing it for now, we can reopen it if needed

PR #84, opened 2024-05-21
Add suggestions to README

ernesgonzalez33:
Added three suggestions on the README based on my experience while playing with the repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,8 +212,8 @@ git push origin HEAD
 ```
 
 4. Your terminal should now display a link for creating a new Pull Request in GitHub.
-   Click the link, make sure the PR is targeted against your fork's `main` branch and
-   not against the repository from which it was forked (i.e. `base repository` should
+   Click the link, **make sure the PR is targeted against your fork's `main` branch and
+   not against the repository from which it was forked** (i.e. `base repository` should
    reside under your user name).
 
    Finally, click "Create pull request".
@@ -556,6 +556,20 @@ In addition, you can configure additional `Identity providers` such as `Github`,
 `Google`, etc.. by clicking on the `Identity providers` tab on the left.
 
 ## Troubleshooting Common Issues
+
+### Using Podman with Kind while also having Docker Installed
+
+If you have docker installed, Kind will try to use it by default, so if you want
+it to use Podman, you can do it by creating the cluster with the following command:
+
+```bash
+KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --name konflux --config kind-config.yaml
+```
+
+### Unknown Field "replacements"
+
+If you get the following error: `error: json: unknown field "replacements"`, while executing
+any of the setup scripts, you will need to update your `kubectl`.
 
 ### PR changes are not Triggering Pipelines
 


README.md
```diff
@@ -35,6 +35,12 @@ From the root of this repository, run the following commands:
 kind create cluster --name konflux --config kind-config.yaml
 ```
 
+**Note:** Kind uses docker by default, so if you want to use podman, run the command the following way:
+
+```bash
+KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --name konflux --config kind-config.yaml
```
yftacherzog: At least when only podman is installed, this is not required. I.e. it uses podman automatically.

avi-biton: we can somehow reflect the fact that if both docker and podman are installed, kind will use docker as a default, and if we want it to run with podman, thats the command to run

ergonzal22: Hmmm. Interesting! I had the issue because I have both installed. I can drop it or rephrase the note. Whatever you two prefer!

yftacherzog: What I try to keep in mind is that the flow is already not short. I'd prefer we don't add things which are not required.

In this case, if it works, I don't mind if the users use podman or docker. We support both, so there's no incentive to make them lean toward podman.

yftacherzog: > Hmmm. Interesting! I had the issue because I have both installed. I can drop it or rephrase the note. Whatever you two prefer!

Did you see issues when using docker?

ergonzal22: I didn't use docker. I have it installed but not enabled. That's why I got the error and I looked for how to use podman with kind instead of docker

yftacherzog: Alright. Let's move it to the troubleshooting section at the end then.

avi-biton: Maybe we can add it to the troubleshoot

ergonzal22: Ok!

ergonzal22: Ok!

README.md
```diff
@@ -77,6 +83,9 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
 ./deploy-konflux.sh
 ```
 
+**Note:** If you get the following error: `error: json: unknown field "replacements"`,
+you will need to update your `kubectl`.
+
```
ergonzal22: Should I move this one to the troubleshooting section as well? @yftacherzog @avi-biton

yftacherzog: Sounds good!

README.md
```diff
@@ -557,6 +557,20 @@ In addition, you can configure additional `Identity providers` such as `Github`,
...
+### Using Podman with Kind while also having Docker installed
+
+If you have docker installed, Kind will try to use it by default, so if you want
+it to use podman, you can do it by creating the cluster with the followind command:
+
+```bash
+KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --name konflux --config kind-config.yaml
+```
+
+### Unknown field "replacements"
```
```suggestion
### Unknown Field "replacements"
```

README.md
```diff
@@ -557,6 +557,20 @@ In addition, you can configure additional `Identity providers` such as `Github`,
 
 ## Troubleshooting Common Issues
 
+### Using Podman with Kind while also having Docker installed
```
```suggestion
### Using Podman with Kind while also having Docker Installed
```

yftacherzog (approved):
lgtm. I commented regarding capitalization, just so we're consistent across the document.

yftacherzog (approved):

README.md
```diff
@@ -557,6 +557,20 @@ In addition, you can configure additional `Identity providers` such as `Github`,
 
 ## Troubleshooting Common Issues
 
+### Using Podman with Kind while also having Docker Installed
+
+If you have docker installed, Kind will try to use it by default, so if you want
+it to use podman, you can do it by creating the cluster with the followind command:
```
```suggestion
it to use Podman, you can do it by creating the cluster with the following command:
```

PR #94, opened 2024-05-23
chore(RHTAPWATCH-1018): test PVC creation pre-deployment

### LINKED ISSUE #58: Test PVC creation before deploying the rest of the manifests
Create a small pvc before deploying anything just to ensure the cluster has default storage cluster and provisioner configured and running.

Omeramsc:
Adding a check in deploy-deps to ensure PVCs are bound and the cluster has default storage cluster, before deploying all the resources.

fix #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@
   * [Unknown Field "replacements"](#unknown-field-replacements)
   * [PR changes are not Triggering Pipelines](#pr-changes-are-not-triggering-pipelines)
   * [Setup Scripts or Pipeline Execution Fail](#setup-scripts-or-pipeline-execution-fail)
+    + [Running out of Resources](#running-out-of-resources)
+    + [Unable to Bind PVCs](#unable-to-bind-pvcs)
   * [Release Fails](#release-fails)
     + [Common Issues](#common-issues)
 - [Repository Links](#repository-links)
@@ -70,7 +72,7 @@ The following applications are required on the host machine:
 
 ### Bootstrapping the Cluster
 
-From the root of this repository, run the following commands:
+From the root of this repository, run the setup scripts:
 
 1. Create a cluster
 
@@ -797,6 +799,8 @@ kubectl delete secret pipelines-as-code-secret -n pipelines-as-code
 
 ### Setup Scripts or Pipeline Execution Fail
 
+#### Running out of Resources
+
 If setup scripts fail or pipelines are stuck or tend to fail at relatively random
 stages, it might be that the cluster is running out of resources.
 
@@ -813,7 +817,7 @@ As last resort, you could restart the container running the cluster node. If you
 that, you'd have to, once more, increase the PID limit for that container as explained
 in the [cluster setup section](#bootstrapping-the-cluster).
 
-To restart the container ( if using Docker, replace `podman` with `docker`):
+To restart the container (if using Docker, replace `podman` with `docker`):
 
 ```bash
 podman restart konflux-control-plane
@@ -821,6 +825,16 @@ podman restart konflux-control-plane
 
 **Note:** It might take a few minutes for the UI to become available once the container
 is restarted.
+
+#### Unable to Bind PVCs
+The `deploy-deps` script includes a check to verify whether PVCs on the default storage
+class can be bind. If volume claims are unable to be fulfilled, the script will fail,
+displaying:
+`error: timed out waiting for the condition on persistentvolumeclaims/test-pvc`.
+
+If you are using Kind, try to [restart the container](#running-out-of-resources).
+Otherwise, ensure that PVCs (Persistent Volume Claims) can be allocated for the
+cluster's default storage class.
 
 ### Release Fails
 
--- a/dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
+++ b/dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
@@ -1,0 +1,34 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-pvc-ns
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: test-pvc
+  namespace: test-pvc-ns
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pvc-consumer
+  namespace: test-pvc-ns
+spec:
+  containers:
+  - name: test-pvc-consumer
+    image: registry:2
+    volumeMounts:
+    - name: my-volume
+      mountPath: /data
+  volumes:
+  - name: my-volume
+    persistentVolumeClaim:
+      claimName: test-pvc
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -5,6 +5,8 @@ script_path="$(dirname -- "${BASH_SOURCE[0]}")"
 main() {
     echo "Checking requirements" >&2
     check_req
+    echo "Testing PVC creation for default storage class" >&2
+    test_pvc_binding
     echo "Deploying Konflux Dependencies" >&2
     deploy
     echo "Waiting for the dependencies to be ready" >&2
@@ -42,6 +44,15 @@ deploy() {
     deploy_tekton
     deploy_keycloak
     deploy_registry
+}
+
+test_pvc_binding(){
+    local pvc_resources="${script_path}/dependencies/pre-deployment-pvc-binding"
+    echo "Creating PVC from '$pvc_resources' using the cluster's default storage class"
+    kubectl apply -k "$pvc_resources"
+    retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s
+    kubectl delete -k "$pvc_resources"
+    echo "PVC binding successfull"
 }
 
 deploy_tekton() {
--- a/dependencies/pre-deployment-pvc-binding/kustomization.yml
+++ b/dependencies/pre-deployment-pvc-binding/kustomization.yml
@@ -1,0 +1,5 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - pre-deployment-pvc-binding.yaml


dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - name: test-pvc-consumer
+    image: registry:2
+    volumeMounts:
+    - name: my-volume
+      mountPath: /data
+  volumes:
+  - name: my-volume
+    persistentVolumeClaim:
+      claimName: test-pvc
+
```
yftacherzog: extra newline

deploy-deps.sh
```diff
@@ -37,13 +37,36 @@ check_req(){
...
+    PVC_NAME=test-pvc
+    PVC_NAMESPACE=test-pvc-ns
+
+    kubectl apply -k "$PVC_RESOURCES"
+
+    timeout_start=$(date +%s)
+    while [[ $(kubectl get pvc "$PVC_NAME" -n "$PVC_NAMESPACE" -o jsonpath='{.status.phase}') != "Bound" ]]; do
+        if [[ "$(($(date +%s) - timeout_start))" -ge 60 ]]
+        then
+            echo "Couldn't bound `$PVC_NAME` PVC within the expected time. Exiting."
```
yftacherzog: `$PVC_NAME` is not being printed here.

yftacherzog: I think we should mention the path to where the PVC is defined and should mention this is trying to create a PVC using the cluster's default storage class.
Ultimately, what's important is that the user knows which storage class they need to be able to create PVCs for.

deploy-deps.sh
```diff
@@ -37,13 +37,36 @@ check_req(){
 }
 
 deploy() {
+    test_pvc_binding
```
yftacherzog: I think this should be called from within `check_req` or right after it in main.

dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
```diff
@@ -0,0 +1,36 @@
...
+metadata:
+  name: test-pvc
+  namespace: test-pvc-ns
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: standard
```
yftacherzog: I'm rethinking this, and I think we should drop the storage class from here. I think on Kind, we're using `standard` all over because this is the default storage class, but on other setups this might be different. We should let the storage class to be the cluster's default.

deploy-deps.sh
yftacherzog: (I edited the comment above)

deploy-deps.sh
Omeramsc: I agree, but I think `main` is better. fixed.

dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
Omeramsc: fixed

dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
Omeramsc: makes sense, fixed.

deploy-deps.sh
Omeramsc:
```
Waiting for PVC 'test-pvc' to be bound...
Couldn't bound 'test-pvc' PVC using the cluster's default storage class within the expected time. Exiting.
```
seem to get printed now 
fixed the rest

deploy-deps.sh
```diff
@@ -5,10 +5,12 @@ script_path="$(dirname -- "${BASH_SOURCE[0]}")"
...
-    echo "Deploying Konflux Dependencies" >&2
-    deploy
-    echo "Waiting for the dependencies to be ready" >&2
-    "${script_path}/wait-for-all.sh"
+    echo "Ensuring the cluster has default storage configured" >&2
+    test_pvc_binding
+    # echo "Deploying Konflux Dependencies" >&2
+    # deploy
+    # echo "Waiting for the dependencies to be ready" >&2
+    # "${script_path}/wait-for-all.sh"
```
yftacherzog: leftovers?

Omeramsc: indeed, fixed.

yftacherzog:
@gbenhaim can you have a look to see if you have any comments?

deploy-deps.sh
```diff
@@ -44,6 +46,29 @@ deploy() {
     deploy_registry
 }
 
+test_pvc_binding(){
+    PVC_RESOURCES="${script_path}/dependencies/pre-deployment-pvc-binding"
```
gbenhaim: please use `local` and lower case variables

deploy-deps.sh
```diff
@@ -44,6 +46,29 @@ deploy() {
...
+    PVC_NAME=test-pvc
+    PVC_NAMESPACE=test-pvc-ns
+
+    echo "Creating PVC from '$PVC_RESOURCES' using the cluster's default storage class"
+    kubectl apply -k "$PVC_RESOURCES"
+
+    timeout_start=$(date +%s)
+    while [[ $(kubectl get pvc "$PVC_NAME" -n "$PVC_NAMESPACE" -o jsonpath='{.status.phase}') != "Bound" ]]; do
+        if [[ "$(($(date +%s) - timeout_start))" -ge 60 ]]
```
gbenhaim: please use a single line if
```
if [[ ... ]]; then
...
fi
```

dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: test-pvc
+  namespace: test-pvc-ns
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: v1
```
gbenhaim, line 19: I think it's sufficient to just create the pvc and ensure it's bound. The pod is not needed.
I also don't see that the script is doing anything with the pod.

gbenhaim (changes requested):

dependencies/pre-deployment-pvc-binding/pre-deployment-pvc-binding.yaml
avi-biton: I am not sure it will be bound unless it will have a pod that uses it

yftacherzog: The default storage class in kind has volume binding mode of wait for first customer. It won't bind unless the pod is there. It binds before the pod starts, so the script doesn't have to wait for the image to be downloaded etc., but it needs the pod to be created.

avi-biton:
Instead of using this function we can use the `wait` in combine with the `retry`
Something like:

`retry kubectl wait --for=jsonpath='{status.phase}'=Bound  pvc/test-pvc -n test-pvc-ns --timeout=20s`
We need to update the readme file for the error message for the failure we will get so that the user will be able to understand where the problem is.

yftacherzog:
> Instead of using this function we can use the `wait` in combine with the `retry` Something like:
> 
> `retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s` We need to update the readme file for the error message for the failure we will get so that the user will be able to understand where the problem is.

@gbenhaim I think this approach makes sense and will look nicer. What do you think?

Omeramsc:
> Instead of using this function we can use the `wait` in combine with the `retry` Something like:
> 
> `retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s` We need to update the readme file for the error message for the failure we will get so that the user will be able to understand where the problem is.

We still need to create the resources, I personally think it is more readable to have it separated, with concise logs, rather than have the user reference the manual to understand why the script is failing on him before deploying anything...

avi-biton:
The script will fail for `condition on persistentvolumeclaims/test-pvc` so it will be clear, but we will also mention in the README file that this is an error caused by the inability to create a PVC

Omeramsc:
> condition on persistentvolumeclaims/test-pvc

Is `condition on persistentvolumeclaims/test-pvc` a clear error though? I integrated it currently, using conditional to include clear fail\success messages

deploy-deps.sh
Omeramsc: fixed

deploy-deps.sh
Omeramsc: fixed

yftacherzog:
> > condition on persistentvolumeclaims/test-pvc
> 
> Is `condition on persistentvolumeclaims/test-pvc` a clear error though? I integrated it currently, using conditional to include clear fail\success messages

It's a good question. It's a question of maintenance burden and code clarity against user experience.
We have many of those retry/wait calls throughout this script. If we change them all to conditionals, we will have a much longer script and a lot more code to maintain.

Maybe we can pass a failure message to the retry call and then display it if all retries fail. This way we keep the code simple and also allow an error message.

gbenhaim:
> > Instead of using this function we can use the `wait` in combine with the `retry` Something like:
> > `retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s` We need to update the readme file for the error message for the failure we will get so that the user will be able to understand where the problem is.
> 
> @gbenhaim I think this approach makes sense and will look nicer. What do you think?

@yftacherzog if we can do it using `kubectl wait` then we should.

yftacherzog:
> > > Instead of using this function we can use the `wait` in combine with the `retry` Something like:
> > > `retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s` We need to update the readme file for the error message for the failure we will get so that the user will be able to understand where the problem is.
> > 
> > 
> > @gbenhaim I think this approach makes sense and will look nicer. What do you think?
> 
> @yftacherzog if we can do it using `kubectl wait` then we should.

@Omeramsc let's do it then as @avi-biton suggests and document the error message on the troubleshooting section of the README. We can later on try to add an error message to the retry function.

README.md
```diff
@@ -27,7 +27,7 @@ The following applications are required on the host machine:
 
 ### Bootstrapping the Cluster
 
-From the root of this repository, run the following commands:
+From the root of this repository, run the following commands, using the setup scripts:
```
```suggestion
From the root of this repository, run the setup scripts:
```

README.md
```diff
@@ -622,6 +622,8 @@ kubectl delete secret pipelines-as-code-secret -n pipelines-as-code
 
 ### Setup Scripts or Pipeline Execution Fail
 
+#### Running out of resources
```
```suggestion
#### Running out of Resources
```

README.md
```diff
@@ -647,6 +649,14 @@ podman restart konflux-control-plane
 **Note:** It might take a few minutes for the UI to become available once the container
 is restarted.
 
+#### Cluster lacks deault storage
```
```suggestion
#### Unable to Bind PVCs
```

README.md
```diff
@@ -647,6 +649,14 @@ podman restart konflux-control-plane
...
+The `deploy-deps` script includes a check to verify whether the cluster has default storage configured.
```
```suggestion
The `deploy-deps` script includes a check to verify whether PVCs on the default storage class can be bind.
```

README.md
```diff
@@ -647,6 +649,14 @@ podman restart konflux-control-plane
...
+If default storage is absent, the script will fail, displaying:
```
```suggestion
If volume claims are unable to be fulfilled, the script will fail, displaying: 
```

README.md
```diff
@@ -647,6 +649,14 @@ podman restart konflux-control-plane
 **Note:** It might take a few minutes for the UI to become available once the container
 is restarted.
 
+#### Cluster lacks deault storage
+The `deploy-deps` script includes a check to verify whether the cluster has default storage configured.
+If default storage is absent, the script will fail, displaying: 
+`error: timed out waiting for the condition on persistentvolumeclaims/test-pvc`.
+
+If you are using Kind, try to re-create the cluster. 
+Otherwise, ensure that PVCs (Persistent Volume Claims) can be allocated for the cluster's default storage class.
```
yftacherzog: Let's introduce line breaks at 89 characters (same for other lines in this PR as well).

deploy-deps.sh
```diff
@@ -5,6 +5,8 @@ script_path="$(dirname -- "${BASH_SOURCE[0]}")"
 main() {
     echo "Checking requirements" >&2
     check_req
+    echo "Ensuring the cluster has default storage configured" >&2
```
```suggestion
    echo "Testing PVC creation for default storage class" >&2
```

deploy-deps.sh
```diff
@@ -44,6 +46,13 @@ deploy() {
     deploy_registry
 }
 
+test_pvc_binding(){
+    local pvc_resources="${script_path}/dependencies/pre-deployment-pvc-binding"
+    echo "Creating PVC from '$pvc_resources' using the cluster's default storage class"
+    kubectl apply -k "$pvc_resources"
+    set -e; kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s
```
yftacherzog: We should not have the `set -e` here and we should use `retry`.
```suggestion
    retry kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s
```

deploy-deps.sh
```diff
@@ -44,6 +46,13 @@ deploy() {
     deploy_registry
 }
 
+test_pvc_binding(){
+    local pvc_resources="${script_path}/dependencies/pre-deployment-pvc-binding"
+    echo "Creating PVC from '$pvc_resources' using the cluster's default storage class"
+    kubectl apply -k "$pvc_resources"
```
yftacherzog, line 52: Need to also delete this after the `wait` line

README.md
```diff
@@ -647,6 +649,14 @@ podman restart konflux-control-plane
 **Note:** It might take a few minutes for the UI to become available once the container
 is restarted.
 
+#### Cluster lacks deault storage
+The `deploy-deps` script includes a check to verify whether the cluster has default storage configured.
+If default storage is absent, the script will fail, displaying: 
+`error: timed out waiting for the condition on persistentvolumeclaims/test-pvc`.
+
+If you are using Kind, try to re-create the cluster.
```
yftacherzog: Thinking about it, we can actually do better with Kind. We can restart the container instead of restarting the cluster. You can point to the running out of resources section -- it's explained there:
```
To restart the container ( if using Docker, replace `podman` with `docker`)
...
```
(Can I ask you to delete the leading space after the opening bracket there? `( if` --> `(if` )

yftacherzog:
@Omeramsc can you rebase the PR on top of upstream/main?

deploy-deps.sh
Omeramsc: Why not fail the script if we can't bind PVC? We wanted it to `exit 1` before using retry

yftacherzog: We have multiple such statements which are failing the script without setting this option just before running them. Is this statement any different? What's the reason for it to need this while others don't?

Omeramsc: Simply the fact it does not fail the script:
```
error: timed out waiting for the condition on persistentvolumeclaims/test-pvc
namespace "test-pvc-ns" deleted
persistentvolumeclaim "test-pvc" deleted
pod "test-pvc-consumer" deleted
Deploying Konflux Dependencies
namespace/cert-manager unchanged
customresourcedefinition.apiextensions.k8s.io/certificaterequests.cert-manager.io unchanged
customresourcedefinition.apiextensions.k8s.io/certificates.cert-manager.io unchanged
....
```

deploy-deps.sh
Omeramsc: added

README.md
Omeramsc: fixed (also the last request)

deploy-deps.sh
Omeramsc: fixed

README.md
Omeramsc: fixed

README.md
Omeramsc: fixed

README.md
Omeramsc: fixed

README.md
Omeramsc: fixed

README.md
Omeramsc: fixed

README.md
Omeramsc: fixed

deploy-deps.sh
Omeramsc: AFAIK, The other statements shouldn't fail the script unless we explicitly act upon their return status

avi-biton: I think we should fail the script in case a PVC cannot be created.
There is no point in moving on further.

avi-biton:
I think its better to wait for https://github.com/konflux-ci/konflux-ci/pull/101 to be merged and then use the retry with the error message.

Omeramsc:
> I think its better to wait for #101 to be merged and then use the retry with the error message.

Either way, let's have them both merged by tomorrow. They do not depend on each other, so let's make sure we are not dragging them to the next sprint for nothing.

deploy-deps.sh
yftacherzog: Whatever is using `retry` does fail the script.

Omeramsc: Right. I'm not sure what caused it to continue, but I removed the pod for the sake of testing it and it indeed failed after the timeout reached.  
fixed

deploy-deps.sh
```diff
@@ -44,6 +46,15 @@ deploy() {
...
+    kubectl wait --for=jsonpath='{status.phase}'=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s
```
yftacherzog: Can we use retry here? If we merge the PR that allows adding message to retry then we can make use of it to make this failure clearer.

Omeramsc: Sorry, I left it out by mistake. yes, retry is intended to be used.
fixed

yftacherzog (approved):

yftacherzog:
@gbenhaim can you have another look and see if you're ok with merging this?

Issue #97, opened 2024-05-24
GIthub check results should direct users to the local UI


ralphbean:
PaC is deployed in deploy-deps.sh, and the UI is deployed in deploy-konflux.sh.

In order for PaC to present links to the user that direct them to the local UI for check results, the PaC config needs to be provided details about the custom console.

PR #98, opened 2024-05-24
feat: provide PaC links to the local console

ralphbean:
This way, a user who clicks on their github check result will be directed to their their local instance's UI.

Fixes https://github.com/konflux-ci/konflux-ci/issues/97

### DIFF
--- a/dependencies/pipelines-as-code/custom-console-patch.yaml
+++ b/dependencies/pipelines-as-code/custom-console-patch.yaml
@@ -1,0 +1,16 @@
+---
+- op: add
+  path: /data/application-name
+  value: Local Konflux
+- op: add
+  path: /data/custom-console-name
+  value: Local Konflux
+- op: add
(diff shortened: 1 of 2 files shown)


ralphbean:
I wasn't sure whether to approach this with a whole new kustomize directory, or just use kubectl patch like this. I'm open to changing it if you feel this is the wrong way.

We'll have more cases of customizing a dep's config down the road.

ralphbean:
Looking at this again now, the right way is to apply a kustomize patch in `dependencies/pipelines-as-code/`. I'll circle back and do that.

ralphbean:
(I poked at this for a few minutes the other day and got confused by the relationship between the TektonConfig and the PaC configmap. I tried setting details in the TektonConfig but they didn't seem to take effect.)

gbenhaim (approved):

PR #101, opened 2024-05-27
chore: print reason on retry failure

yftacherzog:
Changing the implementation of the retry function so it takes the command to retry and the error message to print if retry fails.

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -47,10 +47,12 @@ deploy() {
 deploy_tekton() {
     # Operator
     kubectl apply -k "${script_path}/dependencies/tekton-operator"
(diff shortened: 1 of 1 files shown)


yftacherzog:
@gbenhaim following the discussion on the other PR. Do you think an approach such as the one in this PR makes sense? It will allow us to keep using unified retry mechanism, while also supporting a customized error message.

yftacherzog:
cc @Omeramsc

Omeramsc (approved):
LGTM

avi-biton (approved):
LGTM

avi-biton:
I think there is an issue when we use double quotes to pass the parameters to the retry function.
The message I get when adding quotes to this command is (for example):
`kubectl wait --for=condition=Ready --timeout=120s -l app.kubernetes.io/instance=cert-manager -n cert-manager pod: No such file or directory
`

PR #102, opened 2024-05-28
test(RHTAPWATCH-1043): add kubelint to CI

yftacherzog:


yftacherzog:
Status:
- The linter is running - it might not be the most ideal way to make it run for all kustomization files (e.g. some nested files may run multiple times -- not sure if that's a problem), but it's working.

- I applied some fixes to some of the issues. Some of them are breaking the deployment (I think the main issue is low resources that need to be tweaked). This needs to be looked on one by one.

- Some fixes are still missing. Need to attend to those as well.

PR #103, opened 2024-05-29
Integrate the new import flow

gbenhaim:
Update hac-dev image
    
    The new image has the new simple import form and several fixes
    for the link to the documentation.

    Update build-service image
    
    The new image supports importing a component without a devfile and
    selecting a pipeline by adding an annotation on the component.
    See test/resources/demo-users/user/ns1/application-and-component.yaml
    for an example.

new-import

### DIFF
--- a/test/resources/demo-users/user/ns1/application-and-component.yaml
+++ b/test/resources/demo-users/user/ns1/application-and-component.yaml
@@ -17,12 +17,13 @@ metadata:
   annotations:
     build.appstudio.openshift.io/request: configure-pac
     image.redhat.com/generate: '{"visibility": "public"}'
+    build.appstudio.openshift.io/pipeline: '{"name":"docker-build", "bundle": "latest"}'
 spec:
   application: sample-component
   componentName: sample-component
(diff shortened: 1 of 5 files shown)


test/resources/demo-users/user/ns1/application-and-component.yaml
```diff
@@ -17,12 +17,13 @@ metadata:
   annotations:
     build.appstudio.openshift.io/request: configure-pac
     image.redhat.com/generate: '{"visibility": "public"}'
+    build.appstudio.openshift.io/pipeline: '{"name":"docker-build", "bundle": "latest"}'
```
yftacherzog, line 20: Should we have this for ns2 application-and-component as well?

gbenhaim: Until we don't sort out the issue with the private image in the build pipeline I don't want to change anything in ns2 (which the readme refers to).

yftacherzog (approved):
lgtm

PR #112, opened 2024-05-30
chore(RHTAPWATCH-1044): Run shellcheck on shell scripts

hmariset:
Using Github actions write shellcheck to run on
all the shell scripts.

Jira-Url: https://issues.redhat.com/browse/RHTAPWATCH-1044

### DIFF
--- a/.github/workflows/differential-shellcheck.yaml
+++ b/.github/workflows/differential-shellcheck.yaml
@@ -1,0 +1,34 @@
+name: Differential ShellCheck
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # only required for workflows in private repositories
+      actions: read
+      contents: read
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v4
+        with:
+          # Differential ShellCheck requires full git history
+          fetch-depth: 0
+
+      - id: DiffShellCheck
+        name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -19,8 +19,8 @@ check_req(){
     local uninstalled_requirements=()
 
     # check if requirements are installed
-    for i in ${requirements[@]}; do
-        if ! command -v $i &> /dev/null; then
+    for i in "${requirements[@]}"; do
+        if ! command -v "$i" &> /dev/null; then
                 uninstalled_requirements+=("$i")
         else
                 echo -e "$i is installed"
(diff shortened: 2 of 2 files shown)


yftacherzog (changes requested):
I think we can have the fix to deploy-deps.sh in this PR as well, rather than a different one.

Also, I think shellcheck is preinstalled on the `ubuntu-latest` image that we're using here, so I'd prefer not to rely on a personally-maintained action and instead see if it's going to be any issue running shellcheck directly. Can you give it a try?

hmariset:
> I think we can have the fix to deploy-deps.sh in this PR as well, rather than a different one.
> 
> Also, I think shellcheck is preinstalled on the `ubuntu-latest` image that we're using here, so I'd prefer not to rely on a personally-maintained action and instead see if it's going to be any issue running shellcheck directly. Can you give it a try?

@yftacherzog it worked without using the action and also updated the deploy-deps.sh here

yftacherzog:
> > I think we can have the fix to deploy-deps.sh in this PR as well, rather than a different one.
> > 
> > Also, I think shellcheck is preinstalled on the `ubuntu-latest` image that we're using here, so I'd prefer not to rely on a personally-maintained action and instead see if it's going to be any issue running shellcheck directly. Can you give it a try?
> 
> @yftacherzog it worked without using the action and also updated the deploy-deps.sh here

Looks neat! Did you confirm that the test can fail if you unfix the script that failed the action previously?

hmariset:
> > > I think we can have the fix to deploy-deps.sh in this PR as well, rather than a different one.
> > > Also, I think shellcheck is preinstalled on the `ubuntu-latest` image that we're using here, so I'd prefer not to rely on a personally-maintained action and instead see if it's going to be any issue running shellcheck directly. Can you give it a try?
> > 
> > 
> > @yftacherzog it worked without using the action and also updated the deploy-deps.sh here
> 
> Looks neat! Did you confirm that the test can fail if you unfix the script that failed the action previously?

Yes, I checked that and it's failing if we unfix the script.

yftacherzog:
@gbenhaim I asked @hmariset to give a try to [this action](https://github.com/redhat-plumbers-in-action/differential-shellcheck). I'm seeing now that it's developed by Red Hat people, but doesn't seem to be official Red Hat stuff.

@hmariset also implemented the check on her own, so we can do without that action. Do you think we should use the action or use our own check.

gbenhaim:
I think we should use [differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck)
Major projects already use it - https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#real-life-examples-of-usage

yftacherzog:
> I think we should use [differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck) Major projects already use it - https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#real-life-examples-of-usage

@hmariset could you remove the custom implementation and leave only differential-shellcheck?

yftacherzog (approved):
lgtm

hmariset:
@gbenhaim @avi-biton can you please review and merge the PR

yftacherzog (changes requested):
@hmariset what's the reason for removing the fix to the existing script?

hmariset:
> @hmariset what's the reason for removing the fix to the existing script?

Differentail shellcheck is only checking the files that are changed in this PR. So, got confused and removed that but I have added that now.

yftacherzog (approved):

PR #117, opened 2024-06-02
chore(RHTAPWATCH-1046): collect failed deployment information

### LINKED ISSUE #17: Collect deployment information if the deployment fails
If the deployment fails, in order to make it possible to debug, we should collect the deployment definitions and the logs of the failed pods from the kind cluster and make them available for the PR owner to review.

The script should be also runable by a user, that can run it when the deployment fails for generating logs that can be used for debugging.

Omeramsc:
Adds a script which collects failed pods information using k8s events. This will be used to debug failed deployments, either from the CI or locally.

fix #17

### DIFF
--- a/generate-err-logs.sh
+++ b/generate-err-logs.sh
@@ -1,0 +1,51 @@
+#!/bin/bash -e
+
+
+main() {
+    echo "Generating error logs" >&2
+    generate_logs
+    echo "Generated logs sucessfully" >&2
+}
+
+generate_logs() {
+    local logs_dir="logs"
+    local pod_definitions_file="$logs_dir/failed-pods-definitions.yaml"
+    local pod_logs_file="$logs_dir/failed-pods-logs.log"
+    local event_messages_file="$logs_dir/failed-deployment-event-log.log"
+
+    rm -rf "$logs_dir"
+    mkdir -p "$logs_dir"
+
+    local namespaces
+    namespaces=$(kubectl get namespaces -o name | xargs -n1 basename)
+
+    for namespace in $namespaces; do
+        # Get all 'Warning' type events that occured on the namespace and extract the relevant fields from it as variables.
+        local events
+        events=$(kubectl get events -n "$namespace" \
+                --field-selector type=Warning \
+                -o jsonpath='{range .items[*]}{.involvedObject.kind}{" "}{.involvedObject.name}{" "}{.message}{" ("}{.reason}{")\n"}{end}'
+        
+        )
+
+        echo "$events" | while read -r kind name message reason; do
+            if [ "$kind" == "Pod" ]; then
+                local pod_definition
+                local pod_logs
+                pod_definition=$(kubectl get pod -n "$namespace" "$name" -o yaml 2>&1)
+                pod_logs=$(kubectl logs -n "$namespace" "$name" --all-containers=true 2>&1)
+
+                printf "%s\n---\n" "$pod_definition" | tee -a "$pod_definitions_file"
+                echo "Pod '$name' under namespace '$namespace':" | tee -a "$pod_logs_file" | tee -a "$event_messages_file"
+                echo "$kind $name $message $reason" | tee -a "$event_messages_file"
+                echo "$pod_logs" | tee -a "$pod_logs_file"
+        fi
+    done
+done
+}
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    main "$@"
+fi
+
+
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -46,3 +46,7 @@ jobs:
       - name: Deploy test resources
         run: |
           ./deploy-test-resources.sh
+      
+      - name: Generate error logs
+        run: |
+          ./generate-err-logs.sh
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+logs/
+.vscode/


yftacherzog (commented):
https://kind.sigs.k8s.io/docs/user/quick-start#exporting-cluster-logs

Kind has some built-in tools for exporting logs. Have a look.

Omeramsc:
@yftacherzog I used that, but it's a bit more complicated and less user-friendly than what the issue suggests. From what I can see, one will need to thoroughly search the logs to find which container failed and why.

test/resources/demo-users/user/ns1/kustomization.yaml
```diff
@@ -6,4 +6,5 @@ resources:
   - application-and-component.yaml
   - ec-integration-test.yaml
   - release-plan.yaml
+  - afc.yaml
```
avi-biton: Why do we need it to be in `user/ns1` ?

Omeramsc: We don't, it'll be removed when out of draft. It is meant for seeing the results in the CI

generate-err-logs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
gbenhaim: should be `#!/bin/bash -e` so the script will stop if there is an error.

generate-err-logs.sh
```diff
@@ -0,0 +1,29 @@
...
+===== Definition for pod: $pod in namespace: $namespace =====
+$pod_definition
+
+===== Event message: $message =====
+
+
+EOF
+        fi
+    done
+done
```
gbenhaim, line 44: Can we add a main function like we have in the other scripts?

generate-err-logs.sh
```diff
@@ -0,0 +1,29 @@
...
+for namespace in $namespaces; do
+    events=$(kubectl get events -n $namespace --field-selector type=Warning -o jsonpath='{range .items[*]}{.involvedObject.kind}{" "}{.involvedObject.name}{" "}{.message}{" ("}{.reason}{")\n"}{end}')
+
+    echo "$events" | while read -r kind name message; do
+        if [ "$kind" == "Pod" ]; then
+            pod=$name
+            pod_logs=$(kubectl logs -n $namespace $pod --all-containers=true 2>&1)
+            pod_definition=$(kubectl get pod -n $namespace $pod -o yaml 2>&1)
+            cat <<EOF | tee -a $output_file
+===== Logs for pod: $pod in namespace: $namespace =====
```
gbenhaim: I think it would be better to generate 3 files instead of putting 3 different logs in the same file. For example, syntax highlighting will work for the yaml files.

generate-err-logs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+output_file="deployment-err.log"
+> $output_file
+
+namespaces=$(kubectl get namespaces -o=jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')
```
gbenhaim: `kubectl get namespace -o name` will be simpler

gbenhaim (changes requested):
In addition, lets merge https://github.com/konflux-ci/konflux-ci/pull/112 first.

generate-err-logs.sh
Omeramsc: Added

generate-err-logs.sh
Omeramsc: Added, now there are 3 files:
`failed-pods-logs.log` a log containing the containers' logs of each pod
`failed-pods-definitions.yaml`  a yaml file containing the definitions of the failed logs
`failed-deployment-event-log.log` a log containing the error\warning events of failed pods

as well as printing to the console

generate-err-logs.sh
Omeramsc: Right, added

generate-err-logs.sh
Omeramsc: changed, but it required getting rid of the `namespace/` prefix so I used `xargs -n1 basename`

Omeramsc:
I'm taking the PR out of draft, but keeping the `do-not-merge` label, so it won't be merged by mistake while having the extra resources.

generate-err-logs.sh
```diff
@@ -0,0 +1,42 @@
...
+generate_logs() {
+    pod_definitions_file="failed-pods-definitions.yaml"
+    pod_logs_file="failed-pods-logs.log"
+    event_messages_file="failed-deployment-event-log.log"
+    > $pod_logs_file > $pod_definitions_file > $event_messages_file
+
+    namespaces=$(kubectl get namespaces -o name | xargs -n1 basename)
+
+    for namespace in $namespaces; do
+        events=$(kubectl get events -n $namespace --field-selector type=Warning -o jsonpath='{range .items[*]}{.involvedObject.kind}{" "}{.involvedObject.name}{" "}{.message}{" ("}{.reason}{")\n"}{end}')
```
yftacherzog: Can you add a comment for what this line does?

generate-err-logs.sh
```diff
@@ -0,0 +1,42 @@
...
+
+        echo "$events" | while read -r kind name message; do
+            if [ "$kind" == "Pod" ]; then
+                pod_definition=$(kubectl get pod -n $namespace $name -o yaml 2>&1)
+                pod_logs=$(kubectl logs -n $namespace $name --all-containers=true 2>&1)
+
+                # printf "%s\n---\n" "$pod_definition" >> "$pod_definitions_file"
+                # echo "$pod_logs" >> "$pod_logs_file"
+                # echo "$kind $name $message" >> $event_messages_file
```
yftacherzog: Leftovers?

Omeramsc: yes, removed

generate-err-logs.sh
Omeramsc: Added. also broke the lines to add more clarity.

generate-err-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+generate_logs() {
+    pod_definitions_file="failed-pods-definitions.yaml"
+    pod_logs_file="failed-pods-logs.log"
+    event_messages_file="failed-deployment-event-log.log"
+    > $pod_logs_file > $pod_definitions_file > $event_messages_file
+
+    namespaces=$(kubectl get namespaces -o name | xargs -n1 basename)
+
+    for namespace in $namespaces; do
+        # Get all 'Warning' type events that occured on the namespace and extract the relevant fields from it as variables.
```
yftacherzog, line 23: What about higher severity events? Are we not interested in those?

avi-biton: The supported types are `Normal` and `Warning`:
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_events/#synopsis

yftacherzog (approved):

yftacherzog:
@Omeramsc can you rebase the PR? We have shellcheck now applied, and rebasing it should have it checked.

generate-err-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+
+
+main() {
+    echo "Generating error logs" >&2
+    generate_logs
+    echo "Generated logs sucessfully" >&2
+}
+
+generate_logs() {
+    pod_definitions_file="failed-pods-definitions.yaml"
```
gbenhaim: please use `local` for variables defined inside of the function.

generate-err-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+    pod_logs_file="failed-pods-logs.log"
+    event_messages_file="failed-deployment-event-log.log"
+    > $pod_logs_file > $pod_definitions_file > $event_messages_file
```
gbenhaim: what does it do?

Omeramsc: reset the files' contents; it is not much useful for the CI as it is for local use.

generate-err-logs.sh
Omeramsc: fixed for all

generate-err-logs.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+
+generate_logs() {
+    local pod_definitions_file="failed-pods-definitions.yaml"
+    local pod_logs_file="failed-pods-logs.log"
+    local event_messages_file="failed-deployment-event-log.log"
+
+    cat /dev/null > "$pod_logs_file"
+    cat /dev/null > "$pod_definitions_file"
+    cat /dev/null > "$event_messages_file"
```
yftacherzog: Maybe this will be more straightforward: 
```suggestion
    rm -f "$pod_logs_file" "$pod_definitions_file" "$event_messages_file"
```
Maybe can put them all inside a single directory and then we can delete the directory instead.
Also, I think we should add those files or the directory to .gitignore.

Omeramsc: yup, that actually makes much more sense. 
added

generate-err-logs.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo "Generated logs sucessfully" >&2
+}
+
+generate_logs() {
+    local logs_dir="logs"
+    local pod_definitions_file="$logs_dir/failed-pods-definitions.yaml"
+    local pod_logs_file="$logs_dir/failed-pods-logs.log"
+    local event_messages_file="$logs_dir/failed-deployment-event-log.log"
+
+    rm -f "$logs_dir"
```
yftacherzog:
```suggestion
    rm -rf "$logs_dir"
```

Omeramsc: just fixed. thanks

yftacherzog (commented):
@Omeramsc I think we can remove the always-failing pod and move this out of draft.

gbenhaim (approved):

PR #120, opened 2024-06-03
Increase wait timeout for keyclock and cert-manager

### LINKED ISSUE #75: increase retry count or sleep time
Some steps inside the deploy-deps scripts tend to fail randomly on retry. We might need to increase the retry count or increase the sleep time.

avi-biton:
Increase wait timeout for `keyclock` and `cert-manager` by 1 minute

fix #75

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -86,6 +86,7 @@ deploy_tekton() {
 
 deploy_cert_manager() {
     kubectl apply -k "${script_path}/dependencies/cert-manager"
+    sleep 5
     retry "kubectl wait --for=condition=Ready --timeout=120s -l app.kubernetes.io/instance=cert-manager -n cert-manager pod" \
           "Cert manager did not become available within the allocated time"
 }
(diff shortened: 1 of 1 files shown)


deploy-deps.sh
```diff
@@ -86,7 +86,7 @@ deploy_tekton() {
 
 deploy_cert_manager() {
     kubectl apply -k "${script_path}/dependencies/cert-manager"
-    retry "kubectl wait --for=condition=Ready --timeout=120s -l app.kubernetes.io/instance=cert-manager -n cert-manager pod" \
+    retry "kubectl wait --for=condition=Ready --timeout=180s -l app.kubernetes.io/instance=cert-manager -n cert-manager pod" \
```
yftacherzog: This won't address the issue with cert-manager. It was failing not because the timeout was reached, but because the pod was not created after 3 x sleep time allowed by the retry function.

yftacherzog: (I don't have a good solution to this issue)

avi-biton: In this case we can increase the number of retries, or the sleep between retries

yftacherzog: This will apply to any retry we do. Not sure that we want that. It might be that, instead, we can sleep a little before this retry to reduce the chance of this happening. This should not cost us anything if we assume that the retry will either iterate more than once or the pod will take some time to be ready.

avi-biton: We can use a parameter for how many retries we want, it involves changes to the retry and all the uses of it.
Or as you mentioned we can add sleep before the retry, something like:
```
retry() {

    for _ in {1..3}; do
        sleep 3
        local ret=0
        $1 || ret="$?"
        if [[ "$ret" -eq 0 ]]; then
            return 0
        fi
    done

    echo "$1": "$2."
    return "$ret"
}
```

yftacherzog: I meant adding a sleep before the retry is called specifically for the case of cert-manger. I don't think it will have any cost in terms of execution time for the entire script - it will just move the waiting for an earlier stage.

Yes. We can also introduce a new arg to the retry function as you mentioned.

avi-biton: Done (adding sleep before the retry)

yftacherzog (approved):
lgtm

PR #122, opened 2024-06-03
Add minimum and recommended requirements

avi-biton:
Add minimum and recommended requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <!-- toc -->
 
 - [Trying Out Konflux](#trying-out-konflux)
-  * [Machine Requirements](#machine-requirements)
+  * [Machine Minimum Requirements](#machine-minimum-requirements)
   * [Installing Software Dependencies](#installing-software-dependencies)
   * [Bootstrapping the Cluster](#bootstrapping-the-cluster)
   * [Enable Pipelines Triggering via Webhooks](#enable-pipelines-triggering-via-webhooks)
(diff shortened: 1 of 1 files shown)


yftacherzog (commented):
I think we should mention what the deployment will use rather than the system requirement.

E.g. If someone runs it on their own laptop, it will require different system resources comparing to running it on a dedicated VM.

I'd go with something like this:

```markdown
## Machine Minimum Requirements

The deployment requires the following **free** resources:
- RAM: 8GB
- CPU: 4 cores

**Note:** Additional load from running multiple pipelines in parallel will require
additional resources.
```

avi-biton:
Well, I can speak for what I have tested.
A machine with 4 cores and free 8GB was not tested, the minimum working tested environment is 4 CPU and 16 GB RAM.

yftacherzog:
> Well, I can speak for what I have tested. A machine with 4 cores and free 8GB was not tested, the minimum working tested environment is 4 CPU and 16 GB RAM.
> 
> We can make it our minimum requirement, and add the note in your comment (skip the recommended requirements)

I was referring to the resources that the cluster requires rather than the over system requirements (we don't know the load on the system - just what the cluster requires). We checked the overall resource requests that our pods have and it's around 4GB of RAM and 2 CPUs if I recall, which made me say that this is what we should ask the user to have available for the cluster.

yftacherzog (approved):

avi-biton:
The actual cluster resources were checked (`kubectl describe nodes konflux-control-plane`)
Requirements were set accordingly.

PR #127, opened 2024-06-04
doc: reference the troubleshooting section earlier

yftacherzog:
Move the reference to the troubleshooting section to an earlier point so that users will know it exists before trying to troubleshoot on their own.


yftacherzog:
@ShakedAviv50313 please take a look

avi-biton (approved):

PR #130, opened 2024-06-05
Add privileges to appstudio-pipeline service account in user-ns2 namespace

avi-biton:
Add privileges to appstudio-pipeline service account to be able to create and delete jobs and to get logs.


yftacherzog (approved):
lgtm

PR #136, opened 2024-06-06
Rename konflux-ci dir to konflux-core

gbenhaim:
This will help with the understating about which components are a must to have konflux, and which components can be installed additionally for enriching the experience.


yftacherzog:
@gbenhaim can you rebase and address the conflicts?

Issue #135, opened 2024-06-05
RPA documentation is missing setting the component name


yftacherzog:
In section `Create ReleasePlan and ReleasePlanAdmission Resources`, we need to also set the component name. This is not documented at the moment.

PR #137, opened 2024-06-06
doc: set component name in rpa manifest

yftacherzog:
The component name was missing from the instructions for editing the RPA manifest. This change adds it.

closes: #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,13 +659,16 @@ kubectl create -f ./test/resources/demo-users/user/ns2/release-plan.yaml
 Edit the `ReleasePlanAdmission`
 [manifest](./test/resources/demo-users/user/managed-ns2/rpa.yaml):
 
-1. Replace `<repository url>` with the URL of the repository on the registry to
-   which your released images are to be pushed. This is typically a different repository
-   comparing to the one builds are being pushed during tests.
+1. Under `applications`, verify that your application is the one listed.
 
-   For example, if using Quay.io and your username is `my-user` and you created a
-   repository called `my-konflux-component-release` under your own organization, then
-   the configs should look like this:
+2. Under the components mapping list, set the `name` field so it matches the name of
+   your component and replace `<repository url>` with the URL of the repository on the
+   registry to which your released images are to be pushed. This is typically a
+   different repository comparing to the one builds are being pushed during tests.
+
+   For example, if your component is called `test-component`, and you wish to release
+   your images to a Quay.io repository called `my-user/my-konflux-component-release`,
+   then the configs should look like this:
 
 ```yaml
     mapping:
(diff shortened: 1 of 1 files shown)


avi-biton:
I think that it would be easier to find you way according to the order in the RPA yaml file:
1 - set the application/s name/s
2 - set the component/s name/s and their respected  repositories

yftacherzog:
> I think that it would be easier to find your way according to the order in the RPA yaml file: 1 - set the application/s name/s 2 - set the component/s name/s and their respected repositories

I see what you mean.
We give the use a placeholder they can search for (`<repository url>`).
I imagine it's easiest to search for that rather than anything else.
Maybe we can change the order of the fields on the YAML file so the order of appearance is kept and they can search by the placeholder?

yftacherzog:
> > I think that it would be easier to find your way according to the order in the RPA yaml file: 1 - set the application/s name/s 2 - set the component/s name/s and their respected repositories
> 
> I see what you mean. We give the user a placeholder they can search by (`<repository url>`). I imagine it's easiest to search for that rather than anything else. Maybe we can change the order of the fields on the YAML file so the order of appearance is kept and they can search by the placeholder?

I can mention the app first, regardless.

avi-biton:
> > I think that it would be easier to find your way according to the order in the RPA yaml file: 1 - set the application/s name/s 2 - set the component/s name/s and their respected repositories
> 
> I see what you mean. We give the user a placeholder they can search by (`<repository url>`). I imagine it's easiest to search for that rather than anything else. Maybe we can change the order of the fields on the YAML file so the order of appearance is kept and they can search by the placeholder?

Since this yaml file cannot be applied as it is (one must update the `repository url` place holder before applying it), why not adding placeholders for the application and component as well?

yftacherzog:
> > > I think that it would be easier to find your way according to the order in the RPA yaml file: 1 - set the application/s name/s 2 - set the component/s name/s and their respected repositories
> > 
> > 
> > I see what you mean. We give the user a placeholder they can search by (`<repository url>`). I imagine it's easiest to search for that rather than anything else. Maybe we can change the order of the fields on the YAML file so the order of appearance is kept and they can search by the placeholder?
> 
> Since this yaml file cannot be applied as it is (one must update the `repository url` place holder before applying it), why not adding placeholders for the application and component as well?

For the manual approach, this would mean more changes. When using the manual flow, I'm not sure how much the user is aware of the name of the application and component (they see it on the UI, but I don't think they manually set it in any step).

I wanted not to introduce this requirement for those users. This way, they only need to make one change.

avi-biton:
OK, so lets change the order in the README file.
Application -> component -> repository

yftacherzog:
> OK, so lets change the order in the README file. Application -> component -> repository

I want to have app -> repo -> component, and change the order on the rpa.yaml so that repo appears before component.

yftacherzog:
> > OK, so lets change the order in the README file. Application -> component -> repository
> 
> I want to have app -> repo -> component, and change the order on the rpa.yaml so that repo appears before component.

I gave it another thought and I'll have it the way you suggest.

PR #148, opened 2024-06-12
Show the user's first and last name on the UI

gbenhaim:
There was bug where the user's first and last name were not visible in the UI since they weren't included in the JWT token. Include the "first_name" and "last_name" claims in the JWT token.

image


yftacherzog (approved):

ifireball (approved):
lgttm

PR #172, opened 2024-06-18
Add mise configuration for tool autmation

ifireball:
Using [mise][1] to setup kind and kubectl.

[1]: https://mise.jdx.dev/

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,3 @@
+[tools]
+kind = "0.22.0"
+kubectl = "1.30.0"


yftacherzog (commented):
@ifireball Can you elaborate on what this PR does? How does it setup `kind` and `kubectl`?

ifireball:
Mise is a dev environment setup tool,  kind of like pipenv or npm, but for general tools as opposed to a particular language. Its based on the asdf plugins, but also supports other installation backends.

In short, with this file in place, and with mise installed, going into the directory and typing `mise install` causes the tools listed to be installed and made available in $PATH. IF the needed versions of the tools are already available locally, just CD-ing into the repo dir causes the right versions to be in $PATH.

Mise also support env variable, hooks and scripts that may come handy in the future to make the experience of using this repo nicer.

yftacherzog:
Let's create an issue to add it.

ifireball:
> Let's create an issue to add it.

Added #175

yftacherzog (approved):

PR #180, opened 2024-06-19
chore(RHTAPWATCH-1043): run Kubelinter on Konflux-ci

ShakedAviv50313:
Running Kubelinter on Konflux-ci

### DIFF
--- a/.github/.kube-linter-config.yaml
+++ b/.github/.kube-linter-config.yaml
@@ -1,0 +1,16 @@
+
+checks:
+
+  include: [ ]
+
+  exclude: 
+    - "unset-cpu-requirements"
+    - "unset-memory-requirements"
+    - "no-read-only-root-fs"
+    - "run-as-non-root"
+    - "env-var-secret"
+    - "liveness-port"
+    - "readiness-port"
+    - "startup-port"
+    - "non-existent-service-account"
+    - "latest-tag"
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -1,0 +1,31 @@
+name: Static code checks
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+jobs:
+  kube-linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Create ./.kube-linter/ directory
+        shell: bash
+        run: mkdir -p ./.kube-linter/
+      - name: Generate default configuration
+        shell: bash
+        run: |
+          find . -name "kustomization.yaml" -o -name "kustomization.yml" | while read -r file; do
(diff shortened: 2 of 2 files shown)


ShakedAviv50313:
@yftacherzog This pr contains the kube-linter check with all relevant checks disabled.
each disabled check will be enabled one at a time to keep track of the progress of solving issues

yftacherzog:
> @yftacherzog @avi-biton This pr contains the kube-linter check with all relevant checks disabled.
> each disabled check will be enabled one at a time to keep track of the progress of solving issues
>  

Did you confirm that this check is able to fail the ci?

ShakedAviv50313:
I'm not sure what do you mean, each of these checks mentioned in `.kube-linter-config.yaml` is able to fail the ci, this is why they are excluded from the checks. (for now)

.github/.kube-linter-config.yaml
```diff
@@ -0,0 +1,9 @@
+
+checks:
+  # include explicitly adds checks, by name. You can reference any of the built-in checks.
+  # Note that customChecks defined above are included automatically.
+  include: [ ]
+  # exclude explicitly excludes checks, by name. exclude has the highest priority: if a check is
+  # in exclude, then it is not considered, even if it is in include as well.
+  exclude: [ "unset-cpu-requirements", "unset-memory-requirements", "no-read-only-root-fs", "run-as-non-root", "env-var-secret", "liveness-port",
```
yftacherzog: Can we make it like this (or something similar if I missed something on the syntax)?
```suggestion
  exclude:
    - "unset-cpu-requirements"
    - "unset-memory-requirements"
    - "no-read-only-root-fs"
    - "run-as-non-root"
    - "env-var-secret"
    - "liveness-port"
```

.github/workflows/kube-linter.yaml
```diff
@@ -0,0 +1,34 @@
...
+            dir=$(dirname "$file")
+            dir=${dir#./}
+            output_file=$(echo "out-$dir" | tr "/" "-")
+            kustomize build "$dir" > "./.kube-linter/$output_file.yaml"
+          done
+      - name: Scan yaml files with kube-linter
+        uses: stackrox/kube-linter-action@v1
+        id: kube-linter-action-scan
+        with:
+          # Adjust this directory to the location where your kubernetes resources and helm charts are located.
```
yftacherzog: Let's remove the comment

.github/workflows/kube-linter.yaml
```diff
@@ -0,0 +1,34 @@
...
+          directory: ./.kube-linter/
+          # Adjust this to the location of kube-linter config you're using, or remove the setting if you'd like to use
+          # the default config.
```
yftacherzog: Let's remove the comment

yftacherzog (approved):
lgtm

PR #181, opened 2024-06-20
chore(RHTAPWATCH-1043): fix kubelinter root-related issues

ShakedAviv50313:
Fix Kubelinter issues in Konflux ci.
This will be done by removing checks out of the excluded checks and solve them one by one

### DIFF
--- a/.github/.kube-linter-config.yaml
+++ b/.github/.kube-linter-config.yaml
@@ -1,4 +1,3 @@
-
 checks:
 
   include: [ ]
@@ -6,8 +5,6 @@ checks:
   exclude: 
     - "unset-cpu-requirements"
     - "unset-memory-requirements"
-    - "no-read-only-root-fs"
-    - "run-as-non-root"
     - "env-var-secret"
     - "liveness-port"
     - "readiness-port"
--- a/dependencies/keycloak/deployment/db.yaml
+++ b/dependencies/keycloak/deployment/db.yaml
@@ -2,6 +2,9 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: postgresql-db
+  annotations:
+    ignore-check.kube-linter.io/no-read-only-root-fs: "This component requires write permissions"
+    ignore-check.kube-linter.io/run-as-non-root: "This component requires root permissions"
 spec:
   serviceName: postgresql-db-service
   selector:
--- a/dependencies/keycloak/deployment/kustomization.yml
+++ b/dependencies/keycloak/deployment/kustomization.yml
@@ -7,3 +7,19 @@ resources:
   - db.yaml
   - configure-keycloak.yaml
 namespace: keycloak
+
+patches:
+  - patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1no-read-only-root-fs
+        value: "This deployment needs to have write permissions"
+    target:
+      kind: Deployment
+      name: keycloak-operator
+  - patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1run-as-non-root
+        value: "This deployment needs to have root permissions"
+    target:
+      kind: Deployment
+      name: keycloak-operator
--- a/dependencies/registry/registry.yaml
+++ b/dependencies/registry/registry.yaml
@@ -11,6 +11,9 @@ metadata:
     run: registry
   name: registry
   namespace: kind-registry
+  annotations:
+    ignore-check.kube-linter.io/no-read-only-root-fs: "This component requires write permissions"
+    ignore-check.kube-linter.io/run-as-non-root: "This component requires root permissions"
 spec:
   replicas: 1
   selector:
--- a/dependencies/tekton-results/kustomization.yml
+++ b/dependencies/tekton-results/kustomization.yml
@@ -4,4 +4,25 @@ kind: Kustomization
 resources:
   - https://storage.googleapis.com/tekton-releases/results/previous/v0.9.2/release.yaml
   - certificate.yaml
-  
+patches:
+  - patch: |
+      - op: add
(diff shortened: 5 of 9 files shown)


yftacherzog:
> Fix Kubelinter issues in Konflux ci. This will be done by removing checks out of the excluded checks and solve them one by one

I'd suggest reducing the scope of the PR for a single exclude, or maybe if you take `no-read-only-root-fs` you can have it together with `run-as-non-root`.

if needed (maybe for the resources errors), you can even have a single exclude line fixed by multiple PRs (i.e. we only remove the exclude after we fix the last offending manifest).

dependencies/registry/registry.yaml
```diff
@@ -24,6 +24,9 @@ spec:
       containers:
       - name: registry
         image: registry:2
+        securityContext:
+          readOnlyRootFilesystem: true
```
avi-biton: I suspect that this time the CI is failing for these configurations

dependencies/tekton-results/kustomization.yml
```diff
@@ -4,4 +4,26 @@ kind: Kustomization
...
+      kind: Deployment
+      name: tekton-results-watcher
+  - patch: |
+      - op: add
+        path: /spec/template/spec/containers/0/securityContext/readOnlyRootFilesystem
+        value: true
+    target:
+      kind: StatefulSet
+      name: tekton-results-postgres
+
```
avi-biton: Remove the spaces at the last line :)

.github/.kube-linter-config.yaml
```diff
@@ -1,13 +1,15 @@
 
 checks:
+  ignorePaths:
+    - "/home/runner/work/konflux-ci/konflux-ci/.kube-linter/out-dependencies-keycloak-deployment.yaml"
+    - "/home/runner/work/konflux-ci/konflux-ci/.kube-linter/out-dependencies-registry.yaml"
+    - "/home/runner/work/konflux-ci/konflux-ci/.kube-linter/out-dependencies-tekton-results.yaml"
```
yftacherzog: What's the reason for ignoring those?

generate-err-logs.sh
yftacherzog, line 1: Are the changes in this file related to this PR?

.github/.kube-linter-config.yaml
ShakedAviv50313: for all of them, Kubelinter fails because "readOnlyRootFilesystem" is not set to true. 
when setting it to be true, for each of them separately, the ci fails because apparently they need to be able to write and not to read only.
@avi-biton might be able to elaborate

yftacherzog: Need to check if there's a way to ignore a specific issue on a specific file rather than ignoring the file completely due to a specific issue.

yftacherzog: See if this helps: https://docs.kubelinter.io/#/configuring-kubelinter?id=ignoring-violations-for-specific-cases
As far as I understand, you'd need to patch those files with an annotation to ignore that specific check.

.github/.kube-linter-config.yaml
```diff
@@ -1,13 +1,15 @@
 
 checks:
+  ignorePaths:
+#    - "/home/runner/work/konflux-ci/konflux-ci/.kube-linter/out-dependencies-keycloak-deployment.yaml"
```
Omeramsc: Is this a leftover?

ShakedAviv50313: It was commented out because i wanted to test that the patches work, once I saw they did I deleted this line (and all `IgnorePaths` actually)

dependencies/registry/registry.yaml
```diff
@@ -11,6 +11,9 @@ metadata:
     run: registry
   name: registry
   namespace: kind-registry
+  annotations:
+    ignore-check.kube-linter.io/no-read-only-root-fs: ""
+    ignore-check.kube-linter.io/run-as-non-root: ""
```
yftacherzog: Let's add here a similar comment as added to the patches created for the same annotation.

dependencies/keycloak/deployment/db.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: postgresql-db
+  annotations:
+    ignore-check.kube-linter.io/no-read-only-root-fs: ""
+    ignore-check.kube-linter.io/run-as-non-root: ""
```
yftacherzog: Let's add here a similar comment as added to the patches created for the same annotation.

yftacherzog (commented):
@ShakedAviv50313 the configs looks good to me. Did you run through the readme flow once with those changes?

ShakedAviv50313:
@yftacherzog it worked and I have managed to release succesfully.

dependencies/keycloak/deployment/db.yaml
yftacherzog: @ShakedAviv50313 can you address this comment?

dependencies/keycloak/deployment/kustomization.yml
```diff
@@ -7,3 +7,19 @@ resources:
...
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1no-read-only-root-fs
+        value: "This deployment needs to have write permissions"
+    target:
+      kind: Deployment
+      name: keycloak-operator
+  - patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1run-as-non-root
+        value: ""
```
yftacherzog: add comment here as well.

dependencies/registry/registry.yaml
yftacherzog: @ShakedAviv50313  Can you address this?

ShakedAviv50313: on it

yftacherzog (approved):
lgtm

PR #182, opened 2024-06-23
konftool: Sacaffold Cobra and Echo

ifireball:
"konftool" is going to be a web-based installer that will make
setting up stuff easy.

For now we have some very basic Cobra command setup that
starts an Echo web server, and termintes it properly on ^C.


pkg/konftool/web/echo.go
```diff
@@ -0,0 +1,56 @@
...
+
+	return &Web{server: &srv}, nil
+}
+
+func (w *Web) Stop() error {
+	// Try to shut down up to 10 seconds
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	return w.server.Shutdown(ctx)
+}
```
avi-biton: Add a new line :)

PR #185, opened 2024-06-25
Use ClusterIP service for keycloak's DB

gbenhaim:
It shouldn't use a load balancer since it shouldn't be accessed from outside of the cluster.


avi-biton (approved):
lgtm

PR #187, opened 2024-06-30
konftool:gh_app: GH App creation flow (partial)

ifireball:
Submitting a context structure to GitHub now works
but we need much more work to make this usable.


.mise.toml
```diff
@@ -2,3 +2,4 @@
 kind = "0.22.0"
 kubectl = "1.30.0"
 go = "1.21.9"
+"go:golang.org/x/tools/cmd/stringer" = "0.22.0"
```
gbenhaim, line 5: what is this?

ifireball: `stringer` auto-generates String methods for enums (You can see the go generate comment invoking it [here][1]. This makes `mise` automatically install it for me via `Go get`.

[1]: https://github.com/konflux-ci/konflux-ci/pull/187/files#diff-a35322658f09adc8c669bdf5d631d724fdef6da36054c5a4035caf26a36de5ddR3

PR #188, opened 2024-06-30
chore: ensure logs are generated on task failure

Omeramsc:
ensures the `Generate error logs` step executes even, even if some of the tasks fail.


Omeramsc:
@gbenhaim

PR #194, opened 2024-07-08
feat(RHTAPWATCH-1107): Konftool-storage

ifireball:
Implement storing of user choices and configuration in Konftool.

To get there:
* Refactored various parts to keep function sizes small
* Removed some unneeded temp code
* Implemented top-of-page message display to show storage errors
* Implemented the storge itself (Its somewhat similar to
Viper, but not quite like it)
* Added Ginkgo and Gomega for some tests
* Used the storage from gh_app to ensure it works


yftacherzog:
@ifireball I lost track of what's going here and what the plan is, so I cannot effectively review it.

Issue #156, opened 2024-06-12
Readme: Mention that we plan to release to managed-ns2


yftacherzog:
https://redhat.enterprise.slack.com/archives/C077N3KH5K7/p1718204686299519

PR #201, opened 2024-07-09
doc: explicitly mention release's target namespace

yftacherzog:
We got user feedback saying this will be helpful, so adding.

closes: #156


ifireball (approved):
lgtm

PR #208, opened 2024-07-11
chore: new-ui scaffolding

yftacherzog:
Add configs for authenticating a dummy endpoint (httpbin) with dex + oath2-proxy as placeholder for the new UI.


dependencies/dex/dex.yaml
```diff
@@ -0,0 +1,124 @@
...
+        - name: https
+          containerPort: 9443
+        volumeMounts:
+        - name: dex
+          mountPath: /etc/dex/cfg
+        - name: tls
+          mountPath: /etc/dex/tls
+        readinessProbe:
+          httpGet:
+            path: /idp/healthz
```
yftacherzog: @gbenhaim should it be ok to have `idp` in the path here? It's defined in the nginx configs.

gbenhaim: you don't need to prefix the path with `/idp` since t he rediness probe doesn't go through the proxy, it contacts the container in the pod directly. `/healthz` should be enough.

yftacherzog: This is what I get without it:
```
Readiness probe failed: HTTP probe failed with statuscode: 404
```

dependencies/dex/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+storage:
+  type: kubernetes
+  config:
+    inCluster: true
+web:
+  https: 0.0.0.0:9443
+  tlsCert: /etc/dex/tls/tls.crt
+  tlsKey: /etc/dex/tls/tls.key
+connectors:
+- type: mockCallback
```
avi-biton: What is this for?

yftacherzog: It allows bypassing the authentication process. I'll remove it.

dependencies/dex/dex.yaml
gbenhaim: after looking at the dex helm chart and example config I think that you need to use another endpoint for health checks:

https://github.com/dexidp/dex/blob/master/config.yaml.dist#L68

https://github.com/dexidp/helm-charts/blob/master/charts/dex/templates/deployment.yaml#L110

yftacherzog: Thanks for that. I changed things accordingly.

konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -154,33 +131,51 @@ spec:
...
+          - oauth2-proxy
           - --http-address
           - "127.0.0.1:6000"
+          - --redirect-url
+          - https://localhost:9443/oauth2/callback
           - --oidc-issuer-url
-          - https://localhost:9443/idp/realms/redhat-external
+          - https://localhost:9443/idp/
+          - --cookie-secure
+          - "false"
```
gbenhaim: we need to use a secure cookie that can't get accessed via javascript. It's important for security.

When using a secure cookie, you would need to also set the `--cookie-name` flag. see https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview/

yftacherzog: Changed.

gbenhaim (approved):

PR #214, opened 2024-07-17
Add example pages for logging in and out

yftacherzog:


yftacherzog:
The user reaches this page:

image

If clicks on Login, they reach the login process:

image

After logging in, they reach this page:

image


Clicking Logout will get them to the start again, but with the cookie deleted, so login is required again

yftacherzog:
@gbenhaim see if this is what you had in mind
cc @avi-biton

gbenhaim (approved):

PR #215, opened 2024-07-17
Consume manifests from the services upstream repositories

gbenhaim:
Consume manifests from the services upstream repositories.
This will allow us to configure renovate so we can can automatic updates when the services change.
Each commit contains a change to one service.
The e2e flow was verified locally after making this change.


test/resources/demo-users/user/ns1/ec-integration-test.yaml
yftacherzog, line 1: We have a similar file under ns2 that should be updated as well.

yftacherzog (commented):
Other than the version for the integration test scenario. I was able to verify it works as before for both UI and non-UI flow.

test/resources/demo-users/user/ns1/ec-integration-test.yaml
gbenhaim: fixed

yftacherzog (approved):

PR #221, opened 2024-07-18
remove bearer token

yftacherzog:


yftacherzog:
@gbenhaim see if this is what you had in mind.

konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -29,37 +29,6 @@ spec:
         labelSelector:
           matchLabels:
             app: proxy
-      initContainers:
-      - name: add-sva-token-to-nginx-config
-        image: registry.access.redhat.com/ubi9/ubi@sha256:66233eebd72bb5baa25190d4f55e1dc3fff3a9b77186c1f91a0abdb274452072
-        command:
```
gbenhaim: this part we will need. it's the token used for impersonation when nginx forwards requests to the kube api.

konflux-ci/ui/core/proxy/nginx.conf
```diff
@@ -102,18 +102,6 @@ http {
             # nginx auth_request includes headers but not body
             proxy_set_header Content-Length   "";
             proxy_pass_request_body           off;
-
-            # See https://github.com/kubernetes/kubernetes/pull/47740 for information
```
gbenhaim, line 106: this is the only section that should be removed, we should keep all the rest.
The rest of the configuration are used for authenticating to the kube api, this is what gives nginx the permission to impersonate users when forwarding requests.

gbenhaim (changes requested):

PR #222, opened 2024-07-18
feat: e2e test

psturc:
### Description
Add e2e test for validating the build->integration-test->release scenario (similar to the one described in the [readme of this repo](https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#onboard-a-new-application))

The test is using https://github.com/konflux-ci/testrepo as a component repository and this GitHub app: https://github.com/apps/konflux-ci-test-app

### Notes:
There are currently [issues with running `clair-scan` task in github actions](https://github.com/konflux-ci/konflux-ci/issues/241) (see comments below for more info). The task just hangs in `Running` state until it times out (after 1h). Due to that I introduced a workaround to temporarily remove this task from the build pipeline.

### Current status
* ~after rebase the test is failing on `"couldn't get the secret public-key from openshift-pipelines namespace: secrets \"public-key\" not found"` (see the checkrun [here](https://github.com/psturc/konflux-ci/actions/runs/10071422234/job/27842067606?pr=6)) - probably due to issue https://github.com/konflux-ci/konflux-ci/issues/240~
* since this PR introduces `pull_request_target` type of workflow (due to fact that it requires to load repository secrets like github app ID, private key etc.), the workflow **can't be tested in this PR**, because `pull_request_target` expects this type of workflow to be present on the base (`main`) branch.

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -1,11 +1,19 @@
 name: Sanity Test
 
-on: pull_request
+on: pull_request_target
 
 jobs:
   sanity-test:
     runs-on: ubuntu-latest
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # this might remove tools that are actually needed,
+          # if set to "true" but frees about 6 GB
+          tool-cache: false
+          
+          docker-images: false
       - name: Checkout Repository
         uses: actions/checkout@v3
   
@@ -35,6 +43,15 @@ jobs:
         run: |
           ./wait-for-all.sh
 
+      - name: WORKAROUND - Set up tkn cli for the following task
+        uses: tektoncd/actions/setup-tektoncd-cli@main
+        with:
+          version: latest
+
+      - name: WORKAROUND - Remove clair-scan task from docker-pipeline
+        run: |
+          ./test/e2e/customize-docker-pipeline.sh
+
       - name: Deploying Konflux
         run: |
           ./deploy-konflux.sh
@@ -46,6 +63,25 @@ jobs:
       - name: Deploy test resources
         run: |
           ./deploy-test-resources.sh
+
+      - name: Prepare resources for E2E tests
+        env:
+          APP_ID: ${{ secrets.APP_ID }}
+          APP_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
+          APP_WEBHOOK_SECRET: ${{ secrets.APP_WEBHOOK_SECRET }}
+          QUAY_ORG: ${{ secrets.QUAY_ORG }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+          SMEE_CHANNEL: ${{ secrets.SMEE_CHANNEL }}
+        run: |
+          ./test/e2e/prepare-e2e.sh
+
+      - name: Run E2E tests
+        env:
+          GH_ORG: ${{ secrets.GH_ORG }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          QUAY_DOCKERCONFIGJSON: ${{ secrets.QUAY_DOCKERCONFIGJSON }}
+        run: |
+          ./test/e2e/run-e2e.sh
       
       - name: Generate error logs
         if: ${{ !cancelled() }}
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -1,6 +1,8 @@
 name: Sanity Test
 
-on: pull_request_target
+on:
+  pull_request_target:
+    types: [auto_merge_enabled]
 
 jobs:
   sanity-test:
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -16,6 +16,8 @@ jobs:
           docker-images: false
       - name: Checkout Repository
         uses: actions/checkout@v3
+        with:
+          ref: "${{ github.event.pull_request.head.sha }}"
   
       - name: Create k8s Kind Cluster
         uses: helm/kind-action@v1
--- a/test/e2e/run-e2e.sh
+++ b/test/e2e/run-e2e.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -e
+
+script_path="$(dirname -- "${BASH_SOURCE[0]}")"
+
+main() {
+    echo "Running E2E tests" >&2
+    # shellcheck disable=SC1091
+    source "${script_path}/vars.sh"
+    local github_org="${GH_ORG:?A GitHub org where the https://github.com/redhat-appstudio-qe/konflux-ci-sample repo is present/forked should be provided}"
+    local github_token="${GH_TOKEN:?A GitHub token should be provided}"
+    local quay_dockerconfig="${QUAY_DOCKERCONFIGJSON:?quay.io credentials in .dockerconfig format should be provided}"
+    docker run --network=host -v ~/.kube/config:/kube/config --env KUBECONFIG=/kube/config -e GITHUB_TOKEN="$github_token" -e QUAY_TOKEN="$(base64 <<< "$quay_dockerconfig")" -e MY_GITHUB_ORG="$github_org" -e E2E_APPLICATIONS_NAMESPACE=user-ns2 "$E2E_TEST_IMAGE" "--ginkgo.label-filter=upstream-konflux" "--ginkgo.focus=Test local" "--ginkgo.v"
+}
+
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    main "$@"
+fi
--- a/test/e2e/vars.sh
+++ b/test/e2e/vars.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -e
+export E2E_TEST_IMAGE CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_LOCALHOST CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_CLUSTER
+
+E2E_TEST_IMAGE=quay.io/konflux-ci/e2e-tests:5de23fe3dd41985e6ea1039f053f4165a60a414c
+CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_LOCALHOST="localhost:30001/test/test:customized-docker-pipeline"
+CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_CLUSTER="registry-service.kind-registry.svc.cluster.local/test/test:customized-docker-pipeline"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,6 +5,10 @@ Contributing Guidelines
 
 - [Editing Markdown Files](#editing-markdown-files)
 - [Using KubeLinter](#using-kubelinter)
+- [Running E2E test](#running-e2e-test)
+  * [Prerequisites](#prerequisites)
+  * [Setup](#setup)
+  * [Running the test](#running-the-test)
 
 <!-- tocstop -->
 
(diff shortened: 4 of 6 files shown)


psturc:
@gbenhaim I wasn't able to get the e2e test passing in github actions with "advanced" pipeline, because the "clair-scan" taskrun gets always stuck in `Running` state (and then times out after 1h). 

After a bit of digging I found out that it gets stuck specifically on [`clair-action` command](https://github.com/konflux-ci/build-definitions/blob/acd6b67b2b1dc9447b70e38ee4d696ad4d1c62ab/task/clair-scan/0.1/clair-scan.yaml#L82)
See [this link](https://github.com/psturc/konflux-ci/actions/runs/10007898726/job/27663526336) to the failed CI run (search for "before clair action", which is a debug msg I added to the customized clair-scan task I'm using in the pipelinerun)

However I was able to get e2e test passing with "simple" pipeline (which skips "clair-scan" and other security checks in build pipelinerun): https://github.com/psturc/konflux-ci/actions/runs/10025963019/job/27709723075. The full CI run took ~12 minutes

Do you think it'd be ok to proceed with adding e2e that uses simple pipeline and create issue for investigating the problem with "clair-scan" first?

gbenhaim:
@psturc great progress!
what does it mean simple pipeline? except for the clair-scan, what other tasks it misses?

psturc:
> @psturc great progress! what does it mean simple pipeline? except for the clair-scan, what other tasks it misses?

Basically it's a build pipeline with [`skip-checks` param set to true](https://github.com/konflux-ci/build-definitions/blob/acd6b67b2b1dc9447b70e38ee4d696ad4d1c62ab/pipelines/template-build/template-build.yaml#L36-L39), so the pipeline skips `deprecated-base-image-check`, `clair-scan`, `ecosystem-cert-preflight-checks`, `sast-snyk-check`, `clamav-scan` and `sbom-json-check`

Btw I just want to make clear that when I run the identical test with "advanced" pipeline locally, it works fine. This issue with "clair-scan" task being stuck happens only in github actions CI, but I'm not sure what exactly causes it yet.

> 
> how many tests do we run? and what those tests cover?

We run what's described in the test scenario [here](https://github.com/konflux-ci/e2e-tests/tree/main/tests/rhtap-demo#advanced-build-enabled-only-if-advancedbuildspec-is-set) except for the jvm-build-service part

I think it should more or less cover the build->integration-test->release scenario described in the [readme of this repo](https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#onboard-a-new-application)

gbenhaim:
@psturc sound good.
Can we use this component in the test https://github.com/konflux-ci/testrepo ?
This is the component we mention in the README for users to try out Konflux.

psturc:
@gbenhaim yep I'm running the test with https://github.com/psturc-org/konflux-ci-sample, which is a fork of the repo you mention. So before I mark this PR for review I will update the test to work with https://github.com/konflux-ci/testrepo. And in order to get the test working I will need a GitHub App (that will be used in the test) to install to the repo. Is it ok to create the app under https://github.com/organizations/konflux-ci/settings/apps?

gbenhaim:
@psturc when can we merge this one?

psturc:
@gbenhaim I just found out that konflux is [deprecating the simple build pipeline](https://github.com/konflux-ci/build-service/pull/329), which will cause our test to fail once that change is in 🙃 

So instead of relying on the simple build pipeline I'm currently putting together a workaround to customize the provided docker-build pipeline by removing the `clair-scan` task from it, just for the purpose of this test (until we figure out what's causing that taskrun to hand indefinitely in github actions).

gbenhaim:
@psturc how would the build-service submit your modified pipeline in a PR to the component repository?
the build service uses the following config map to choose which pipeline to submit to the component repository https://github.com/konflux-ci/konflux-ci/blob/main/konflux-ci/build-service/build-pipeline-config.yaml

and the following annotation on the component for selecting the pipeline - https://github.com/konflux-ci/konflux-ci/blob/main/test/resources/demo-users/user/ns1/application-and-component.yaml#L20

psturc:
@gbenhaim  the customization would be done as a part of the sanity-test github action script. I haven't finished that one yet, but just to give you an idea, [here is the change I have in mind](https://github.com/konflux-ci/konflux-ci/compare/main...psturc:konflux-ci:KONFLUX-3683-test-without-clair-scan?expand=1#diff-fd17c12f6a9db21ecea8c3d688e76cad475c81ac6b6ba0ef6eaa958c5bee288eR7-R15)

gbenhaim:
@psturc I'm ok with it as long as we are going to find a better solution asap. I do think you need to push the bundle you build with a unique tag since multiple prs can be tested at the same time.

.github/workflows/sanity-test.yml
```diff
@@ -1,11 +1,19 @@
 name: Sanity Test
 
-on: pull_request
+on: pull_request_target
```
psturc: I used `pull_request_target` event, because the workflow needs to access repository secrets required for the test. More info: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

Also due to that type of event this job wasn't triggered in this PR, but in order to validate that this job works I created a PR against the source branch of this PR (`psturc:KONFLUX-3683`) here: https://github.com/psturc/konflux-ci/pull/6

.github/workflows/sanity-test.yml
```diff
@@ -1,11 +1,19 @@
...
     runs-on: ubuntu-latest
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # this might remove tools that are actually needed,
+          # if set to "true" but frees about 6 GB
+          tool-cache: false
+          
+          docker-images: false
```
psturc, line 18: this is required, otherwise the job fails on pulling the build pipeline images due to `No space left on device` error

test/e2e/vars.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+export E2E_TEST_IMAGE CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_LOCALHOST CUSTOMIZED_DOCKER_PIPELINE_IMAGE_REF_CLUSTER
+
+E2E_TEST_IMAGE=quay.io/konflux-ci/e2e-tests:5de23fe3dd41985e6ea1039f053f4165a60a414c
```
psturc, line 4: In order to keep this image up-to-date, I would later add a task similar to [this one](https://github.com/konflux-ci/e2e-tests/blob/ecbc63890c6f756a3a2ca5b367a5d1eda8173e27/.tekton/konflux-e2e-tests-push.yaml#L417-L440) to create a PR against this repo each time e2e-tests `main` branch is updated.

psturc:
> @psturc I'm ok with it as long as we are going to find a better solution asap. I do think you need to push the bundle you build with a unique tag since multiple prs can be tested at the same time.

In the end in order to avoid logging in to registry via tkn cli I decided to use the [local registry instead](https://github.com/konflux-ci/konflux-ci/pull/222/files#diff-bd25b6440275685258f506f8e8a0cbb1782b31b2737647b615ba84e0147af52fR5-R6) - would that be ok?

yftacherzog:
> ### Current status
> * successful run: https://github.com/psturc/konflux-ci/actions/runs/10068385266/job/27840915402
> * after rebase the test is failing on `"couldn't get the secret public-key from openshift-pipelines namespace: secrets \"public-key\" not found"` (see the checkrun [here](https://github.com/psturc/konflux-ci/actions/runs/10071422234/job/27842067606?pr=6)) - probably due to issue [Failed to generate a key for chains #240](https://github.com/konflux-ci/konflux-ci/issues/240)

@psturc can you see if applying the changes [here](https://github.com/konflux-ci/konflux-ci/pull/242) addresses the issue you're seeing?

psturc:
> > ### Current status
> > 
> > * successful run: https://github.com/psturc/konflux-ci/actions/runs/10068385266/job/27840915402
> > * after rebase the test is failing on `"couldn't get the secret public-key from openshift-pipelines namespace: secrets \"public-key\" not found"` (see the checkrun [here](https://github.com/psturc/konflux-ci/actions/runs/10071422234/job/27842067606?pr=6)) - probably due to issue [Failed to generate a key for chains #240](https://github.com/konflux-ci/konflux-ci/issues/240)
> 
> @psturc can you see if applying the changes [here](https://github.com/konflux-ci/konflux-ci/pull/242) addresses the issue you're seeing?

thanks! I cherry-picked your changes [here](https://github.com/psturc/konflux-ci/pull/6), so let's see how it goes...

.github/workflows/sanity-test.yml
psturc: @gbenhaim @yftacherzog since `pull_request_target` [should be used with caution](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/), I think it would be a good idea to make running the test conditional, to avoid running something fishy in the CI.

One of the ways to do it could be using labels, i.e. something like [this](https://github.com/psturc/konflux-ci/commit/e54d35e93a662710227d5da10251863b0ad03216). Or another option I like would be to run tests when [auto-merge is enabled](https://medium.com/@all_the_things_data_ops/what-is-auto-merge-enabled-in-github-actions-4365f302b400). Wdyt?

yftacherzog: I'm ok with either.

psturc: Ok I went with the [auto-merge option](https://github.com/konflux-ci/konflux-ci/pull/222/commits/757a190b93fb9cc3bd2e690f25182c69944d61c0) and enabled it for the repo together with **showing "update branch" button** (to make user aware of newer changes in the source branch):

Screenshot 2024-07-25 at 8 40 21

We could also require the branch to be **always up-to-date before merge**, but I'm not too sure about enabling that:

Screenshot 2024-07-25 at 8 47 33

yftacherzog: > We could also require the branch to be always up-to-date before merge, but I'm not too sure about enabling that:

Let's try it, and if it becomes messy then we can remove it.

psturc: ok done

gbenhaim:
@psturc great progress! is it ready for review?

psturc:
> @psturc great progress! is it ready for review?

yes

gbenhaim:
@psturc why the `sanity-test ` job is pending?

psturc:
> @psturc why the `sanity-test ` job is pending?

that's because this PR introduces "pull_request_target" event, so it expects the workflow to be present on the main branch (and it's not there yet until we merge this PR). please see my in-line comments about it above.

I tested the exact workflow here: https://github.com/psturc/konflux-ci/pull/8

psturc:
@gbenhaim @yftacherzog in order to prove that this PR can be safe to merge without sanity-test check running, I created [this PR](https://github.com/konflux-ci/konflux-ci/pull/245), which is based on the upstream branch that has [identical content with the source branch of the current PR](https://github.com/konflux-ci/konflux-ci/compare/KONFLUX-3683...psturc:konflux-ci:KONFLUX-3683) (apart from the documentation part)

As you can see the `sanity check` job was triggered and finished successfully after I enabled the [automatic merge](https://github.com/konflux-ci/konflux-ci/pull/245#event-13651724257) (which is what I proposed [here](https://github.com/konflux-ci/konflux-ci/pull/222#discussion_r1690811980) we could try doing from now on)

test/e2e/vars.sh
gbenhaim: you can use renovate for that.

gbenhaim:
> > @psturc I'm ok with it as long as we are going to find a better solution asap. I do think you need to push the bundle you build with a unique tag since multiple prs can be tested at the same time.
> 
> In the end in order to avoid logging in to registry via tkn cli I decided to use the [local registry instead](https://github.com/konflux-ci/konflux-ci/pull/222/files#diff-bd25b6440275685258f506f8e8a0cbb1782b31b2737647b615ba84e0147af52fR5-R6) - would that be ok?

@psturc There are no certificate issues when the kubelet tries to pull the image?

test/e2e/run-e2e.sh
```diff
@@ -0,0 +1,18 @@
...
+script_path="$(dirname -- "${BASH_SOURCE[0]}")"
+
+main() {
+    echo "Running E2E tests" >&2
+    # shellcheck disable=SC1091
+    source "${script_path}/vars.sh"
+    local github_org="${GH_ORG:?A GitHub org where the https://github.com/redhat-appstudio-qe/konflux-ci-sample repo is present/forked should be provided}"
+    local github_token="${GH_TOKEN:?A GitHub token should be provided}"
+    local quay_dockerconfig="${QUAY_DOCKERCONFIGJSON:?quay.io credentials in .dockerconfig format should be provided}"
+    docker run --network=host -v ~/.kube/config:/kube/config --env KUBECONFIG=/kube/config -e GITHUB_TOKEN="$github_token" -e QUAY_TOKEN="$(base64 <<< "$quay_dockerconfig")" -e MY_GITHUB_ORG="$github_org" -e E2E_APPLICATIONS_NAMESPACE=user-ns2 "$E2E_TEST_IMAGE" "--ginkgo.label-filter=upstream-konflux" "--ginkgo.focus=Test local" "--ginkgo.v"
```
gbenhaim, line 12: I think that we should check if podman is installed, if so use it. Will help to support users that want to run e2e test locally and use podman.

gbenhaim:
@psturc great progress! I left some comments that can be addressed in other PRs.
I'll merge the e2e tests.
Thanks for working on this :)

gbenhaim (approved):

PR #234, opened 2024-07-21
Move internal services to its own kustomization

gbenhaim:
This change will allow renovate to properly update the core release service kustomization file.


yftacherzog:
Does this mean that we're currently limited by a single URL resource per kusotmization YAML?

gbenhaim:
@yftacherzog with the current renovate.json config yes. We can always add specific configuration for special cases for for this one I think the seperating the directories is the better approach. In addition, I hope that the internal-services manifests won't by include in core konflux in the future.

yftacherzog (approved):

PR #239, opened 2024-07-23
add self-signed certificate to local registry

yftacherzog:
This will eventually be used for allowing a flow that uses the internal registry instead of quay.io, which should be easier to follow.

Note: this PR is opened toward a feature branch (not main). It can potentially break the e2e flow. The e2e flow will be verified before this is merged to main.

### DIFF
--- a/dependencies/registry/certificate.yaml
+++ b/dependencies/registry/certificate.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: registry-cert
+  namespace: kind-registry
+spec:
+  isCA: true
+  subject:
+    organizations:
+      - konflux
+  dnsNames:
+  - localhost
+  - registry-service.kind-registry
+  issuerRef:
+    kind: ClusterIssuer
+    name: self-signed-cluster-issuer
+  secretName: local-registry-tls
+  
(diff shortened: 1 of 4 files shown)


avi-biton:
Don't we want to expose other port, like 443 (now we are exposing port 80)?

dependencies/registry/certificate.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: registry-cert
+  namespace: kind-registry
+spec:
+  isCA: true
```
gbenhaim, line 8: I don't think that this field is needed. It's needed when certificates can be generated out of this certificate, not when using it with an end service.

yftacherzog: I have a strong recollection that this field was needed last time we played with that. We were missing it and it took us a long time to figure out that this was the thing that was missing.
@avi-biton do you recall more details?

yftacherzog:
> Don't we want to expose other port, like 443 (now we are exposing port 80)?

I'll change it.

avi-biton (approved):
lgtm

dependencies/registry/certificate.yaml
avi-biton: I think it is needed for self signed certificates

gbenhaim (approved):

PR #243, opened 2024-07-25
distribute registry ca certificate

yftacherzog:
Copy the local registry certificate to the cert-manager namespace and use trust-manager to distribute to all namespaces a CA bundle containing a default CA together with the registry's custom CA as configmaps.

Tasks will later on mount those configmaps to allow them to use the self-signed certificate.

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -40,6 +40,7 @@ check_req(){
 
 deploy() {
     deploy_cert_manager
+    deploy_trust_manager
     kubectl apply -k "${script_path}/dependencies/cluster-issuer"
     deploy_tekton
     deploy_keycloak
@@ -91,6 +92,13 @@ deploy_cert_manager() {
           "Cert manager did not become available within the allocated time"
 }
 
+deploy_trust_manager() {
+    kubectl apply -k "${script_path}/dependencies/trust-manager"
+    sleep 5
+    retry "kubectl wait --for=condition=Ready --timeout=60s -l app.kubernetes.io/instance=trust-manager -n cert-manager pod" \
+          "Trust manager did not become available within the allocated time"
+}
+
 deploy_keycloak() {
     kubectl apply -k "${script_path}/dependencies/keycloak/crd"
     sleep 2 # Give the api time to realize it support new CRDs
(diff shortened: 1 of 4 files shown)


deploy-deps.sh
```diff
@@ -124,6 +132,11 @@ deploy_registry() {
     kubectl apply -k "${script_path}/dependencies/registry"
     retry "kubectl wait --for=condition=Ready --timeout=240s -n kind-registry -l run=registry pod" \
           "The local registry did not become available within the allocated time"
+    # Copy the registry secret to cert-manager ns so the ca cert can be distrubuted
+    kubectl delete secret -n cert-manager --ignore-not-found=true local-registry-tls
+    kubectl get secret local-registry-tls --namespace=kind-registry -o yaml \
+        | grep -v '^\s*namespace:\s' | kubectl apply --namespace=cert-manager -f -
```
avi-biton, line 138: Does it have to be created in cert-manager namespace for the distribution?

gbenhaim, line 138: why grep is needed here?

gbenhaim (approved):

deploy-deps.sh
yftacherzog: It's gerping everything but the namepsace line, and then we apply the same manifest to another namespace.

yftacherzog: When deploying trust-manager, we specify a trusted namespace (only one). We can make it a different one. I was thinking whether to make it the namespace of the registry in the first place (i.e. deploy trust-manager for that specific namespace), but decided to go with what we have here on a call with Gal earlier today.

avi-biton: Maybe we can create the certificate in `cert-manager` in the first place and it will be distributed to `kind-registry` as well?
This way we will not have to create a copy of it.
Do you think it might work?

yftacherzog: I was trying that, but it's not possible.
On the registry namespace, we need the private key for that certificate, and trust manager will not distribute private keys.
We can use trust manager to move the CA certificate to where we need it, but we cannot use it to move the tls.key  file to where the registry is created.

avi-biton: Thanks for the explanation
LGTM

avi-biton (approved):

PR #244, opened 2024-07-25
doc: curl registry over https

yftacherzog:
The registry now uses https and the documentation should reflect that.


avi-biton (approved):
LGTM

PR #247, opened 2024-07-28
Generic proxy deployment

gbenhaim:
1. Move the certmanager UI resources to their own kustomization. This is needed when deploying the UI overlay on a cluster that doesn't run certmanager, for example, when using openshift (since it can provide the certificates without certmanager).

2.  Move NodePort service configuration to a patch
The NodePort configurations were specific to the local
deployment on Kind, thus adding them in a patch.
Change the default type of the service to ClusterIP since in most production clusters services are exposed
using an ingress-controller, rather then directly with


yftacherzog:
@psturc how do we trigger the sanity test?

konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -228,22 +228,22 @@ kind: Service
 metadata:
   labels:
     app: proxy
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: serving-cert
```
yftacherzog: Does it make sense to have it in this repo?
We don't have anything to test it here.
Might make better sense to patch it when deployed on openshift.

psturc:
> @psturc how do we trigger the sanity test?

Currently by enabling the auto-merge (we could change it to any of [these activity types](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) it if it doesn't seem to be fit), for example we could configure it to trigger after a certain label is added to the PR

yftacherzog:
> > @psturc how do we trigger the sanity test?
> 
> Currently by enabling the auto-merge (we could change it to any of [these activity types](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) it if it doesn't seem to be fit), for example we could configure it to trigger after a certain label is added to the PR

I think the label option might make better sense, but let's see what @gbenhaim says as well (my apologies for not realizing this before the PR was merged).

gbenhaim:
> Currently by enabling the auto-merge (we could change it to any of [these activity types](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) it if it doesn't seem to be fit), for example we could configure it to trigger after a certain label is added to the PR

@psturc can we trigger it automatically if the owner of the PR is a member of the konflux-ci org?

konflux-ci/ui/core/proxy/proxy.yaml
gbenhaim: removed

yftacherzog (approved):
lgtm

psturc:
> > Currently by enabling the auto-merge (we could change it to any of [these activity types](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) it if it doesn't seem to be fit), for example we could configure it to trigger after a certain label is added to the PR
> 
> @psturc can we trigger it automatically if the owner of the PR is a member of the konflux-ci org?

Yes, we could definitely do that. So for the rest of the PRs (where the author is **not a member of konflux-ci org**), do we want to keep the existing `auto_merge_enabled` or use labels, e.g. look for "ok-for-test" label?

gbenhaim:
| want to keep the existing auto_merge_enabled, or use labels, e.g. look for "ok-for-test" label?
@psturc i think the label is enough

psturc:
@gbenhaim re: [the test failure](https://github.com/konflux-ci/konflux-ci/actions/runs/10141332908/job/28039224398?pr=247) - it's caused by the component [having the same name in all test instances](https://github.com/konflux-ci/testrepo/pull/73/files#diff-1833774c547352e8208448aa8959c6c8bfd8ce3ec83bed6adb550a8559d17d2dR15) which then causes a conflict when multiple tests are trying to create a PR, because the source branch has the same name (`appstudio-konflux-ci-upstream`). I will address it in a next PR

psturc:
@gbenhaim from the test result:
```
Pod 'proxy-5cb67596fb-67ddq' under namespace 'konflux-ui':
Pod proxy-5cb67596fb-67ddq MountVolume.SetUp failed for volume "serving-cert" : secret "serving-cert" not found (FailedMount)
Error from server (BadRequest): container "nginx-120" in pod "proxy-5cb67596fb-67ddq" is waiting to start: PodInitializing
Failed to get pod logs for proxy-5cb67596fb-67ddq in namespace konflux-ui
```

PR #248, opened 2024-07-29
Add merge queue event to the github actions

gbenhaim:


gbenhaim:
I made the configuration in Github. We only need to merge this PR.

yftacherzog (approved):
lgtm

.github/workflows/sanity-test.yml
```diff
@@ -3,7 +3,8 @@ name: Sanity Test
 on:
   pull_request_target:
     types: [auto_merge_enabled]
-
+  merge_group:
+    types: [checks_requested]
```
psturc, line 7: I cannot find anywhere if [`merge_group` type of workflow](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group) has an access to repository secrets (like `pull_request_target` does), so I guess we'll have to try and see 😬 

**update** [it worked](https://github.com/konflux-ci/konflux-ci/actions/runs/10141062232) 🎉

psturc (approved):

PR #242, opened 2024-07-24
fix: tekton-chains job fails to start

yftacherzog:
Configuring the job to run as non-root caused it to fail to start. Setting a specific user should address this.

closes: #240

PR #250, opened 2024-07-29
Remove runAsUser from UI deployments

gbenhaim:
In Openshift, additional Security context constraint should be applied on the service account to run with user outside the range [1001080000, 1001089999].

Remove the use of user 1001 from the UI deployments to be compatible with the deployment on Openshift.

When running on k8s, it's not required to use user 1001.

### DIFF
--- a/konflux-ci/ui/core/chrome/chrome-frontend.yaml
+++ b/konflux-ci/ui/core/chrome/chrome-frontend.yaml
@@ -82,7 +82,6 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
             runAsNonRoot: true
-            runAsUser: 1001
       dnsPolicy: ClusterFirst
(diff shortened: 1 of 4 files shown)


yftacherzog (commented):
Is it relevant only for the UI? We have runAsUser in other places as well. e.g. https://github.com/konflux-ci/konflux-ci/pull/242/files

gbenhaim:
> Is it relevant only for the UI? We have runAsUser in other places as well. e.g. https://github.com/konflux-ci/konflux-ci/pull/242/files

When running on Openshift, openshift-pipelines (chains is part of it) is deployed using an OLM operator, so we are not using the manifests you mentioned on Openshift.

yftacherzog (approved):
lgtm

Issue #252, opened 2024-07-30
[e2e test] test does not create unique component names which leads to conflicts


psturc:
@gbenhaim re: [the test failure](https://github.com/konflux-ci/konflux-ci/actions/runs/10141332908/job/28039224398?pr=247) - it's caused by the component [having the same name in all test instances](https://github.com/konflux-ci/testrepo/pull/73/files#diff-1833774c547352e8208448aa8959c6c8bfd8ce3ec83bed6adb550a8559d17d2dR15) which then causes a conflict when multiple tests are trying to create a PR, because the source branch has the same name (`appstudio-konflux-ci-upstream`). I will address it in a next PR

_Originally posted by @psturc in https://github.com/konflux-ci/konflux-ci/issues/247#issuecomment-2255479962_

PR #253, opened 2024-07-30
fix: e2e: create unique Component names

psturc:
fixes https://github.com/konflux-ci/konflux-ci/issues/252


gbenhaim:
@psturc can you please add to the pr the link to the change in the e2e repo?

yftacherzog (approved):

PR #261, opened 2024-08-01
feat(RHTAPWATCH-1191): Mount custom CA bundle to tekton chains

yftacherzog:
Tekton chains needs access to the custom CA. Otherwise, no attestations are pushed to the registry.

### DIFF
--- a/dependencies/tekton-config/tekton-config.yml
+++ b/dependencies/tekton-config/tekton-config.yml
@@ -19,5 +19,27 @@ spec:
     artifacts.taskrun.format: in-toto
     artifacts.taskrun.storage: ""
     transparency.enabled: "false"
+    options:
+      disabled: false
+      deployments:
+        tekton-chains-controller:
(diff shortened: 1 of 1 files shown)


psturc:
@yftacherzog should we update `sanity-test` in local-registry branch to include e2e test?

yftacherzog:
> @yftacherzog should we update `sanity-test` in local-registry branch to include e2e test?

Most of the tasks that we use from build definition do not support the use of a registry with self-signed certificate.
The epic this ticket is from is mainly for enabling those tasks to support it.

I did a proof of concept with a set of modified tasks to demonstrate how this can work. We can introduce such e2e test that uses the modified tasks (and modified EC and release pipelines) and gradually convert the use of those modified tasks and pipelines to official ones, as we progress with that epic.

I think that could be valuable.

gbenhaim (approved):

PR #264, opened 2024-08-04
Include build-pipeline-config in build-service core

gbenhaim:
The build-pipeline-config contains the reference to the build-pipelines distributed with Konflux.


yftacherzog (approved):
lgtm

PR #268, opened 2024-08-05
Improve nginx deployment

gbenhaim:
1. For 30 seconds, check if no containers in the pod crash loop and only then declare that the pod ready.

2. Add readiness probe. Traffic will be forwarded only to pods that pass the readiness check.

3. Adjust probes to start only after 30 seconds to match the change mention in 1.

4. Don't allow the deployment to have less than the desired replica count, even during upgrade.

### DIFF
--- a/konflux-ci/ui/core/proxy/proxy.yaml
+++ b/konflux-ci/ui/core/proxy/proxy.yaml
@@ -12,10 +12,11 @@ spec:
   selector:
     matchLabels:
       app: proxy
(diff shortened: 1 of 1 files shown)


yftacherzog (approved):
lgtm

yftacherzog:
@psturc any clues why the check membership check is failing here?

gbenhaim:
@yftacherzog I think because my membership was private. I changed it now.

PR #273, opened 2024-08-06
Split rbac into core and chains

ralphbean:
The purpose of this is to enable downstream consumers to consume the core configuration without consuming the chains configuration.


gbenhaim (approved):

yftacherzog (approved):
lgtm

PR #275, opened 2024-08-07
doc(RHTAPWATCH-1173): use local registry end to end

yftacherzog:
With all tasks and processes supporting working with registries with
self-signed CA certificates, we can now convert the "no UI" flow to
use an in-cluster registry end to end.


yftacherzog:
I'm going to modify this PR along the progress of enabling the use of the local registry end to end.

avi-biton (approved):
/lgtm

gbenhaim (approved):

PR #279, opened 2024-08-12
chore(RHTAPWATCH-1182): create secret for local registry

yftacherzog:
The dockerconfigjson file that is used by some of the pipelines to hold info for authentication with the registry is still needed even though the registry is not using any authentication.

With this change, we introduce a dummy secret that allows this file to be created without any authentication info.


avi-biton (approved):
\lgtm

psturc (approved):

gbenhaim (approved):

PR #294, opened 2024-08-15
Added script to restart kind cluster

raukadah:
While troubleshotting Konflux CI, We often need to restart the cluster, increase the PID and open file limit.

Users may miss some of the step. By using restart-kind-cluster.sh script, it will make the restarting the cluster step easier.

### DIFF
--- a/restart-kind-cluster.sh
+++ b/restart-kind-cluster.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -e
+
+script_path="$(dirname -- "${BASH_SOURCE[0]}")"
+
+main() {
+    echo "Restarting Kind cluster" >&2
+    podman restart konflux-control-plane
+    echo "Updating PID limit" >&2
+    podman update --pids-limit 4096 konflux-control-plane
+    echo "Increasing resource limit" >&2
+    sudo sysctl fs.inotify.max_user_watches=524288
+    sudo sysctl fs.inotify.max_user_instances=512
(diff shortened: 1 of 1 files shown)


restart-kind-cluster.sh
```diff
@@ -0,0 +1,18 @@
...
+script_path="$(dirname -- "${BASH_SOURCE[0]}")"
+
+main() {
+    echo "Restarting Kind cluster" >&2
+    podman restart konflux-control-plane
+    echo "Updating PID limit" >&2
+    podman update --pids-limit 4096 konflux-control-plane
+    echo "Increasing resource limit" >&2
+    sudo sysctl fs.inotify.max_user_watches=524288
+    sudo sysctl fs.inotify.max_user_instances=512
```
yftacherzog, line 12: This is only needed after the host is restarted. There's no need for rerunning this command if just restarting the container running the cluster.

yftacherzog (changes requested):
@raukadah Thank you for the PR!
While I understand the need for the mentioned steps, I do have some reservations for having such script inside the codebase.

The reasons being that it's specific to Podman while the user might be using Docker; It's not used anywhere in our processes so it has the potential of becoming dead code that we won't maintain and wouldn't know if it's needed or not. Lastly, as the `inotify` commands are not required in the same context as the `podman` commands, this script is left with 2 commands, and I don't think bundling those command together worth maintaining a script.

Ultimately, I wish we could figure out the underlying reason for this issue rather than institutionalizing the workaround.

Let me know your thoughts.

PR #296, opened 2024-08-15
Use konflux-ui

gbenhaim:
Copy static content from the konflux-ui image and serve it using nginx.

Since the static content contains .woff2 files, include the mime type mapping in the nginx config.

image


yftacherzog (approved):
lgtm

PR #303, opened 2024-08-16
Retry pre-deployment PVC Binding

### LINKED ISSUE #161: serviceaccount "default" not found
When running deploy-deps.sh just after creating the cluster, the script tends to fail. This is not an issue anymore after waiting a few seconds. Need to find a way to avoid the error message (e.g. separate creating of the ns and the pvc and wait between them or something like that).

```bash
→ ./deploy-deps.sh
Checking requirements
kubectl is installed
openssl is installed

All requirements are met
Continue
Testing PVC creation for default storage class
Creating PVC from './dependencies/pre-deployment-pvc-binding' using the cluster's default storage class
namespace/test-pvc-ns unchanged
persistentvolumeclaim/test-pvc unchanged
Error from server (Forbidden): error when creating "./dependencies/pre-deployment-pvc-binding": pods "test-pvc-consumer" is forbidden: error looking up service account test-pvc-ns/default: serviceaccount "default" not found
```

rh-hemartin:
Resolves #161

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -49,7 +49,8 @@ deploy() {
 test_pvc_binding(){
     local pvc_resources="${script_path}/dependencies/pre-deployment-pvc-binding"
     echo "Creating PVC from '$pvc_resources' using the cluster's default storage class"
-    kubectl apply -k "$pvc_resources"
+    sleep 10  # Retries are not enough to ensure the default SA is created, see https://github.com/konflux-ci/konflux-ci/issues/161
+    retry "kubectl apply -k ${pvc_resources}" "Error while creating pre-deployment-pvc-binding"
     retry "kubectl wait --for=jsonpath={status.phase}=Bound pvc/test-pvc -n test-pvc-ns --timeout=20s" \
           "Test PVC unable to bind on default storage class"
     kubectl delete -k "$pvc_resources"


deploy-deps.sh
```diff
@@ -49,7 +49,7 @@ deploy() {
 test_pvc_binding(){
     local pvc_resources="${script_path}/dependencies/pre-deployment-pvc-binding"
     echo "Creating PVC from '$pvc_resources' using the cluster's default storage class"
-    kubectl apply -k "$pvc_resources"
+    retry "kubectl apply -k ${pvc_resources}" "Error while creating pre-deployment-pvc-binding"
```
yftacherzog, line 53: The retries sleep intervals are not enough on my laptop if running the script as soon as the cluster starts.
It might make things better, but it won't work around the issue completely.

The thing is that even if we separate the creation of the namespace and wait on the default service account to be created, I think we'll still need to use retry for that service account creation which will potentially fail in some cases as it fails for me now.

We can maybe introduce a 5 seconds sleep before the retry. I don't like it, but I currently don't see a better alternative.

rh-hemartin: I don't like using sleep either. What about creating a service account other than default with a YAML? I saw you submitted a solution to create the default service account, did you try with another account? If the system that creates the default service account is slow using another service account could solve the problem.

yftacherzog: Yes. That might work.

rh-hemartin: Changing the service account did not work. I added a 10 seconds sleep, as 5 seconds was not enough.

yftacherzog: I'm good with that. Thank you for looking into it!

yftacherzog (approved):
lgtm

deploy-deps.sh
yftacherzog: I tried it now, and it failed the first retry out of 3 after waiting the 10 seconds, and then it succeeded on the second. **For me it means that 10 seconds is a good choice.**

rh-hemartin:
I can't merge, so feel free to merge when you want.

PR #309, opened 2024-08-19
chore(RHTAPWATCH-1076): use released workspace-manager images

yftacherzog:
Use proper released images for workspace-manager instead of the dev image that was used up until now.

### DIFF
--- a/konflux-ci/ui/core/proxy/proxy.yaml
+++ b/konflux-ci/ui/core/proxy/proxy.yaml
@@ -148,7 +148,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
           runAsUser: 1001
-      - image: quay.io/redhat-user-workloads/rhtap-o11y-tenant/workspace-manager/workspace-manager:06df38da81280ba2cf6b4a532d494f1ead8670fe
+      - image: quay.io/konflux-ci/workspace-manager@sha256:6b7bfcf197cbff2ae9b1d84abd73636ce5733921524fb550438d8a8143a536b5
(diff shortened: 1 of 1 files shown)


konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -148,7 +148,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
           runAsUser: 1001
-      - image: quay.io/redhat-user-workloads/rhtap-o11y-tenant/workspace-manager/workspace-manager:06df38da81280ba2cf6b4a532d494f1ead8670fe
+      - image: quay.io/konflux-ci/workspace-manager@sha256:6b7bfcf197cbff2ae9b1d84abd73636ce5733921524fb550438d8a8143a536b5
```
yftacherzog, line 151: @gbenhaim do we need to do anything in order for renovate to keep track of this image?

gbenhaim: you would need to also add an entry here https://github.com/konflux-ci/konflux-ci/blob/main/konflux-ci/ui/core/kustomization.yaml#L8

and yes, it would require enabling another manager in the renovate config. I think we can use the kustomize manager for it https://docs.renovatebot.com/modules/manager/kustomize/ (make sure to enable it only to this file https://github.com/konflux-ci/konflux-ci/blob/main/konflux-ci/ui/core/kustomization.yaml#L8)

gbenhaim (approved):

PR #346, opened 2024-09-03
chore(KFLUXBUGS-1564): use renovate to maintain ws-manager version

yftacherzog:
Also enabling the kustomize manager

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,7 +11,7 @@
       "automerge": true
     }
   ],
-  "enabledManagers": ["regex"],
+  "enabledManagers": ["regex", "kustomize"],
   "customManagers": [
     {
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -11,7 +11,6 @@
       "automerge": true
     }
   ],
-  "enabledManagers": ["regex"],
```
yftacherzog, line 14: @gbenhaim are we ok with having all managers enabled? I was not sure if we intentionally enabled just regex.

yftacherzog: Never mind, for now, I enabled just regex and kustomize, and we can figure out if we want others later.

gbenhaim (approved):

PR #604, opened 2024-10-27
chore: update tekton-operator to version 0.73.1

### LINKED ISSUE #567: Upgrade to the latest version of Tekton
I wanted to change the multi-arch build pipeline to use a matrix of size one by default in https://github.com/konflux-ci/build-definitions/pull/1518. The bug preventing us from being able to do that was resolved in https://github.com/tektoncd/pipeline/pull/8158. In order to consume this change, we need to upgrade to at least [v0.73.0 for the operator](https://github.com/tektoncd/operator/blob/v0.73.0/go.sum#L1764-L1765) as the change was included in `tektoncd/pipelines` [v0.62.1](https://github.com/tektoncd/pipeline/releases/tag/v0.62.1).

yftacherzog:
closes #567


gbenhaim (approved):

avi-biton (approved):

arewm (approved):
Approving the content of the change to solve the related issue. I do not have any knowledge about the impact of the change on functionality.

yftacherzog:
@psturc it seems this task is now blocked on our ability to update the docker-build pipeline (https://github.com/konflux-ci/konflux-ci/pull/509), as with the version that we're using we're seeing this error when running the latest docker-build pipeline:
```
Pipeline user-ns2/testrepo-on-pull-request-cvwl2 can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "resolver type bundles\nname = clamav-scan\n": invalid runtime object: strict decoding error: unknown field "spec.volumes[0].emptydir", unknown field "spec.volumes[1].emptydir"
```

I confirmed locally that this happens only when using the docker-build pipeline that we're using at the moment with the latest tekton-operator version (if I use the latest for both, then locally pipelines are running ok).

Issue #602, opened 2024-10-27
Release stuck in state pending


yftacherzog:
I followed the process for onboarding an application using Kubenetes manifests and a local registry. When I merge the PR, a release is created and the image is pushed to the registry, but the release stays in status pending in the UI.

This is the status of the Release resource:

```
Status:
  Artifacts:
    Images:
      Arches:
        amd64
      Name:  test-component
      Oses:
        linux
      Shasum:  sha256:d08224db735fd32902f01ac46f78a0dcd2fc0210883b46cf574835d28e50382e
      Urls:
        registry-service.kind-registry/test-component-release:latest
  Attribution:
    Author:                  kubernetes-admin
    Standing Authorization:  true
  Automated:                 true
  Completion Time:           2024-10-27T13:29:03Z
  Conditions:
    Last Transition Time:  2024-10-27T13:29:03Z
    Message:               
    Reason:                Succeeded
    Status:                True
    Type:                  Released
    Last Transition Time:  2024-10-27T13:26:25Z
    Message:               
    Reason:                Succeeded
    Status:                True
    Type:                  Validated
    Last Transition Time:  2024-10-27T13:26:25Z
    Message:               
    Reason:                Skipped
    Status:                True
    Type:                  TenantPipelineProcessed
    Last Transition Time:  2024-10-27T13:29:03Z
    Message:               
    Reason:                Succeeded
    Status:                True
    Type:                  ManagedPipelineProcessed
    Last Transition Time:  2024-10-27T13:29:03Z
    Message:               
    Reason:                Skipped
    Status:                True
    Type:                  FinalPipelineProcessed
  Expiration Time:         2024-11-03T13:26:24Z
  Managed Processing:
    Completion Time:  2024-10-27T13:29:03Z
    Pipeline Run:     managed-ns2/managed-xsljq
    Role Binding:     user-ns2/test-component-bjclj-rcm55-rolebinding-for-release-pipelingdxwr
    Start Time:       2024-10-27T13:26:25Z
  Start Time:         2024-10-27T13:26:24Z
  Target:             managed-ns2
  Validation:
    Time:  2024-10-27T13:26:25Z
```

The registry seems to include the released image:

```
→ curl -k https://localhost:30001/v2/test-component-release/tags/list
[...]

PR #750, opened 2024-11-20
fix(KFLUXBUGS-1793): release stuck in pending

yftacherzog:
Update hac-dev to address a regression in the UI that caused releases to be stuck with status pending instead of moving to status Succeeded.

Closes: #602


avi-biton (approved):
lgtm

PR #767, opened 2024-11-21
fix(KFLUXBUGS-1545): verify ec-policies being used

### LINKED ISSUE #298: Verify ec-policies being used
We have EC policies defined in multiple places inside the repo. We need to make sure that we use the correct ones and delete the rest to avoid confusion.

avi-biton:
- Remove ec policies from `managed-ns1` and `managed-ns2`
- Get the policies from EC and add them to the managed namespaces

Closes #298


yftacherzog (approved):
lgtm

PR #793, opened 2024-11-25
Better token handling for websockets

gbenhaim:
For websockets the token should be processed in a certain way before sending it to the k8s API. It should be based64 encoded without padding [1].

The current implementation assumed a padding of 1 char, but it was seen that it can be 2 chars. Adjust the processing to remove any padding from the token.

[1] - https://github.com/kubernetes/kubernetes/pull/47740


yftacherzog (approved):
lgtm

avi-biton (approved):
lgtm

PR #803, opened 2024-11-26
Make docker-build-oci-ta the default pipeline

ralphbean:
The UI already has it hardcoded as the default.

Before this change, user usage would fail since the UI tries to request this pipeline, but it doesn't exist in build-service's configmap.

### DIFF
--- a/konflux-ci/build-service/core/build-pipeline-config.yaml
+++ b/konflux-ci/build-service/core/build-pipeline-config.yaml
@@ -5,9 +5,11 @@ metadata:
   namespace: build-service
 data:
   config.yaml: |
-    default-pipeline-name: docker-build
+    default-pipeline-name: docker-build-oci-ta
     pipelines:
     - name: fbc-builder
       bundle: quay.io/konflux-ci/tekton-catalog/pipeline-fbc-builder@sha256:43be9a3396b62560110cdb883c1eb7966c76b0ef634563f1756b1c45ea001ee2
     - name: docker-build
       bundle: quay.io/konflux-ci/tekton-catalog/pipeline-docker-build@sha256:d51f5f0dd703fdcb7b4f9786ca9c204b0cdddd033c2516a80d55155527db89a6
+    - name: docker-build-oci-ta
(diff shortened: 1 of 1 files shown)


yftacherzog (approved):

psturc:
> My only concern now is that we should have e2e tests for that default pipeline. @psturc what would it take to add such e2e test?

do you mean **add** a new test or **modify** the existing one to use the oci-ta pipeline?

psturc:
> > > My only concern now is that we should have e2e tests for that default pipeline. @psturc what would it take to add such e2e test?
> > 
> > 
> > do you mean **add** a new test or **modify** the existing one to use the oci-ta pipeline?
> 
> If I understand correctly, currently we have a test for each of the pipelines in the file. Is that right? I was thinking about adding one. I suppose, we need to have a test for each of the pipelines the UI is currently offering (unless it's too many), but I'd say the minimum should be the default one and the one that we're asking to use in the README.

Our e2e test currently validates only docker-build pipelinerun.

Adding a new test would require a modification of an existing test in e2e-tests repo. But I think it's nothing complicated.
If you agree I can create a new jira ticket for this and add it to our team's (DevProd) backlog

yftacherzog:
> > > > My only concern now is that we should have e2e tests for that default pipeline. @psturc what would it take to add such e2e test?
> > > 
> > > 
> > > do you mean **add** a new test or **modify** the existing one to use the oci-ta pipeline?
> > 
> > 
> > If I understand correctly, currently we have a test for each of the pipelines in the file. Is that right? I was thinking about adding one. I suppose, we need to have a test for each of the pipelines the UI is currently offering (unless it's too many), but I'd say the minimum should be the default one and the one that we're asking to use in the README.
> 
> Our e2e test currently validates only docker-build pipelinerun.
> 
> Adding a new test would require a modification of an existing test in e2e-tests repo. But I think it's nothing complicated. If you agree I can create a new jira ticket for this and add it to our team's (DevProd) backlog

Thank you!

Let's create an issue here in GitHub, and the automation will create a Jira ticket for that that can be moved around as necessary.

psturc:
Filed https://github.com/konflux-ci/konflux-ci/pull/803#issuecomment-2503560205

ralphbean:
Rebased to address merge conflict.

PR #809, opened 2024-11-27
Authorization bearer token for websockets

gbenhaim:
https://github.com/kubernetes/kubernetes/pull/47740 was misunderstood. it was implemented to support authentication from a web browser. Since we have a proxy, it can include the Authorization bearer token, so no special header is needed for websocket authentication.

Since the UI already includes the "Sec-Websocket-Protocol" with a token, it should be redefined (in the new UI this won't be needed).


yftacherzog (approved):
lgtm

filariow (approved):

PR #810, opened 2024-11-27
Authorization bearer token for websockets - new UI

gbenhaim:
kubernetes/kubernetes#47740 was misunderstood.
    it was implemented to support authentication from a web browser.
    Since we have a proxy, it can include the Authorization bearer token,
    so no special header is needed for websocket authentication.


yftacherzog (approved):
lgtm

PR #824, opened 2024-11-29
Remove duplicate auth endpoint

gbenhaim:
We had two endpoints that are the same but one was used for regular requests and the other fo websockets.

We can use the same endpoint for both.


filariow (approved):
lgtm

PR #831, opened 2024-12-01
Make impersonation optional

gbenhaim:
It's possible to deploy Konflux without using impersonation for the proxy if the proxy and the cluster use the same IDP.

Allow this case by making impersonation optional.


yftacherzog (approved):

filariow (approved):

gbenhaim:
@filariow @yftacherzog I needed to make some additional changes. can you please review again?

PR #837, opened 2024-12-02
Make idp proxying optional

gbenhaim:
The IDP deployment is useful for a local setup. Not all Konflux deployments will need it. Make it optional.

### DIFF
--- a/konflux-ci/ui/core/proxy/nginx.conf
+++ b/konflux-ci/ui/core/proxy/nginx.conf
@@ -43,18 +43,6 @@ http {
         location = /404.html {
         }
 
-        location /idp/ {
-            # Identity Provider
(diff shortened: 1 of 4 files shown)


konflux-ci/ui/core/proxy/nginx.conf
```diff
@@ -157,5 +145,7 @@ http {
             # Used for liveness probes
             return 200;
         }
+
+        include /mnt/nginx-additional-location-configs/*.conf;
```
filariow, line 149: minor, we could use something like `/mnt/nginx.d/*.conf` here

gbenhaim: I wanted to make it obvious that this path should include configuration snippets for location blocks.

PR #949, opened 2024-12-16
fix: use email to login

yftacherzog:
With the new UI we need to use emails to login rather than usernames. This change fixes a couple of places in which it was not documented properly.


avi-biton (approved):
lgtm

PR #965, opened 2024-12-17
Update github-secrets.md

shairevivo:
The 3 dots in the command were wrong and if someone uses copy --> run it will fail.

### DIFF
--- a/docs/github-secrets.md
+++ b/docs/github-secrets.md
@@ -11,15 +11,12 @@ The same secret described there, should be deployed to the `build-service` and
 
 ```bash
 kubectl -n pipelines-as-code create secret generic pipelines-as-code-secret \
-...
 ```
 
 ```bash
 kubectl -n build-service create secret generic pipelines-as-code-secret \
-...
 ```
 
 ```bash
 kubectl -n integration-service create secret generic pipelines-as-code-secret \
-...
 ```


docs/github-secrets.md
```diff
@@ -11,15 +11,12 @@ The same secret described there, should be deployed to the `build-service` and
...
 ```
 
 ```bash
 kubectl -n build-service create secret generic pipelines-as-code-secret \
-...
 ```
 
 ```bash
 kubectl -n integration-service create secret generic pipelines-as-code-secret \
-...
```
yftacherzog, line 24: The commands here should not be used as-is (removing the `...` won't make those commands work).

The actual command is part of the pipelines-as-code documentation that is linked above, and the idea was that the user repeats the command that they just did for namespace `pipelines-as-code` also for the other two namespaces.

Any suggestions for what could have conveyed this intention better?

(We try to copy as little as possible from external docs so we don't have to maintain their changes as well in our docs)

shairevivo: Well, it actually works with removing the dots.  tried. 
Also, it has a copy command on the side hence the intention.

yftacherzog: The pipelines-as-code docs has this command:
```
kubectl -n pipelines-as-code create secret generic pipelines-as-code-secret \
        --from-literal github-private-key="$(cat $PATH_PRIVATE_KEY)" \
        --from-literal github-application-id="APP_ID" \
        --from-literal webhook.secret="WEBHOOK_SECRET"
```

The idea behind this section in our docs is that the user will run this command 3 times, each time for a different namespace.

In other words:

```
kubectl -n pipelines-as-code create secret generic pipelines-as-code-secret \
        --from-literal github-private-key="$(cat $PATH_PRIVATE_KEY)" \
        --from-literal github-application-id="APP_ID" \
        --from-literal webhook.secret="WEBHOOK_SECRET"

kubectl -n build-service create secret generic pipelines-as-code-secret \
        --from-literal github-private-key="$(cat $PATH_PRIVATE_KEY)" \
        --from-literal github-application-id="APP_ID" \
        --from-literal webhook.secret="WEBHOOK_SECRET"

kubectl -n integration-service create secret generic pipelines-as-code-secret \
        --from-literal github-private-key="$(cat $PATH_PRIVATE_KEY)" \
        --from-literal github-application-id="APP_ID" \
        --from-literal webhook.secret="WEBHOOK_SECRET"
```

I'm open to suggestions regarding ways to convey that to the user, preferably without actually copying the full command from the pipelines-as-code docs.

shairevivo: Ok, not sure what is the source of confusion. my suggestion is to keep everything as is and just delete the 3 points at the end of the command as suggested. I tested it and it is working.

yftacherzog: Do you mean you run this command?
```
kubectl -n build-service create secret generic pipelines-as-code-secret \
```
If so, this will keep the prompt open (as the lines ends with `\`), and if you press enter again, it will just create an empty secret, which will not return an error, but also will not allow the different controllers to talk with github.

shairevivo: oh, got it ... so maybe just remove the command from command box with copy so user will know there is no intent for him to run it?

yftacherzog: If we can make it look decent and it will make things clearer then I'm fine with that.

yftacherzog:
@shairevivo closing for now. Feel free to reopen in the future.

PR #966, opened 2024-12-17
Update README.md

shairevivo:
There is a missing step which took me a long time to understand. adding it in the (what i think) right order \ place in the installation flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,9 @@ ssh -L 9443:localhost:9443 $USER@$VM_IP
 
 `password:` `password`
 
+In order to build images you need to configure quay.io repository, This step is mandatory step for importing components using the UI.
+[https://github.com/konflux-ci/konflux-ci/blob/main/docs/quay.md#automatically-provision-quay-repositories-for-container-images]
+
 We now have Konflux up and running. Next, we shall configure Konflux to respond
 to Pull Request webhooks, build a user application and push it to a registry.
 


README.md
```diff
@@ -172,6 +172,9 @@ ssh -L 9443:localhost:9443 $USER@$VM_IP
 
 `password:` `password`
 
+In order to build images you need to configure quay.io repository, This step is mandatory step for importing components using the UI.
```
yftacherzog, line 175: I think we have it documented.

If onboarding using the UI, then image controller should take care of that. This is documented in [this section](https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#option-1-onboard-application-with-the-konflux-ui) under `Create an organization and an application in Quay.io...`.

If not using the UI then this is [explained here](https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#push-pull-request-builds-to-external-registry).

Can you have a look to see if this covered what we were missing?

shairevivo: I know it is documented just not in the right place IMHO. so either add a reference (i added the link) or something else. 
The place that I added it is exactly when you actually need to define it.

yftacherzog: The reason for it to be where it is right now is that we have more than one scenario for integrating with quay, so we provide the instructions according to the relevant scenario (we cannot provide it at this early stage as the scenarios diverge only later on).

We're trying to introduce the steps as they are needed. In this case, we first want to create the application and component and only integrate with quay once we're about to build an image.

shairevivo: That is where you are missing the point. You cannot! add component to the application without this quay integration. it will fail. happened to me :-(

yftacherzog: The user is not expected to add a component at that stage.

The intention is for them to follow the procedure step by step. If using the UI to onboard, their next steps are to enable pipeline triggering and to install the GitHub app, and only then they need to configure Quay integration and after that they are asked to create the application/component.

If following those steps in order, they should not face issues.

shairevivo: ok, got it.

Issue #171, opened 2024-06-18
Dump more helpful stuff to the screen at the end of deploy-konflux.sh


ifireball:
Currently `deploy-konflux.sh` includes the following two lines:
```
kubectl describe deployment proxy -n konflux-ui
kubectl logs deployment/proxy -n konflux-ui --all-containers=true --tail=10
```

This end up dumping some very messy stuff on the screen. I suspect that was used for debugging at some point. Perhaps we should remove those lines and instead tell the user the cluster is ready and how to access it, or suggest the next steps such as deploying the test data.

PR #1011, opened 2024-12-26
Dump information at the end of deploy-konflux.sh

avi-biton:
Add information for successful deployment
Dump error log in case of deployment failure

#171

### DIFF
--- a/deploy-konflux.sh
+++ b/deploy-konflux.sh
@@ -9,8 +9,16 @@ main() {
     echo "Waiting for Konflux to be ready" >&2
     local ret=0
     "${script_path}/wait-for-all.sh" || ret="$?"
-    kubectl describe deployment proxy -n konflux-ui
-    kubectl logs deployment/proxy -n konflux-ui --all-containers=true --tail=10
+    if [ $ret -ne 0 ]; then
+        echo "Deployment failed"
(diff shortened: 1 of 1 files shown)


deploy-konflux.sh
```diff
@@ -9,8 +9,18 @@ main() {
...
+        echo "Deployment failed"
+        ./generate-err-logs.sh
+    else
+        echo -e "\nKonflux is up and running:
+        To access the console, deploy the test resources ./deploy-test-resources.sh
+        Konflux Console is available at https://localhost:9443, you can login using the test user
+        username: user2@konflux.com
+        password: password
+            
+        Follow https://github.com/konflux-ci/konflux-ci?tab=readme-ov-file#enable-pipelines-triggering-via-webhooks to continue the onboarding process\n"
```
yftacherzog: I'm not sure we should have it here. Few reasons:

1. It duplicates documentation, which will translate to maintenance burden.
2. We provide 3 steps here, each should be done at a different point in time, and that could be confusing to the user.

I think we should just say that Konflux is up and running.

avi-biton: I agree, I removed the instructions

yftacherzog (approved):
lgtm

Issue #169, opened 2024-06-18
Automate the Smee setup flow


ifireball:
Manually setting up the smee.io channel and the client is cumbersome, especially when [tkn pac bootstrap][1] already automated it .

[1]: https://pipelinesascode.com/docs/guide/cli/

Suggestions:

1. Automate the process like `tkn pac` does (Perhaps simply call it directly)
2. Save the channel information to a file somehow so it can be reused if the cluster is recreated

yftacherzog:
I'm not sure I want to install PaC's code as it will mean another dependency for the user, which I think is more complicated than creating the channel.

Looking at their code, it seems as it creates the channel by generating a random string with 12 chars or more and appending it to this URL: https://hook.pipelinesascode.com/.
You don't need to actually create the channel afterwards.

With https://smee.io/, you don't have to create the channel either and it doesn't seem as there's any limitation to the string appended to this URL (it could be a single character).

All in all, we can automate this. Just need to figure out what's the best way to save the value once generated. Maybe can store a kustomize patch file with the channel name (add it to gitignore). `deploy-deps.sh` can generate the value and store it to the patch if it doesn't exist and deploy the smee manifest as part of `deploy-deps.sh`.

PR #1012, opened 2024-12-30
feat(KFLUXVNGD-3): automate smee channel creation

yftacherzog:
Randomize the smee channel ID and create the client providing that ID. The ID is stored in a kustomize patch file (ignored by git). The channel will not be created if the patch already exists, so users can store their pre-existing channels in the patch and keep using them.

closes: #169

### DIFF
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -19,6 +19,7 @@ jobs:
       - name: Generate default configuration
         shell: bash
         run: |
+          cp dependencies/smee/smee-channel-id.tpl dependencies/smee/smee-channel-id.yaml
           find . -name "kustomization.yaml" -o -name "kustomization.yml" | while read -r file; do
             dir=$(dirname "$file")
             dir=${dir#./}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 logs/
 .vscode/
 __debug_bin*
+dependencies/smee/smee-channel-id.yaml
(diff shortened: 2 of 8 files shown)


avi-biton:
dependencies/smee/smee-channel-id.tpl -> dependencies/smee/smee-channel-id.yaml

yftacherzog:
> dependencies/smee/smee-channel-id.tpl -> dependencies/smee/smee-channel-id.yaml

@avi-biton how do you mean?

avi-biton:
> > dependencies/smee/smee-channel-id.tpl -> dependencies/smee/smee-channel-id.yaml
> 
> @avi-biton how do you mean?

The file extension should be `yaml` 
You also refer to it in the kustomization file:
```
patches:
  - path: smee-channel-id.yaml
```

yftacherzog:
> > > dependencies/smee/smee-channel-id.tpl -> dependencies/smee/smee-channel-id.yaml
> > 
> > 
> > @avi-biton how do you mean?
> 
> The file extension should be `yaml` You also refer to it in the kustomization file:
> 
> ```
> patches:
>   - path: smee-channel-id.yaml
> ```

In git we hold the template (`.tpl`). The script modifies that file and stores the modified file to a `.yaml` file which was added to `.gitignore`.
So, once the script ran, the user will have a .yaml file that was used for deploying the client and from which they can read the value, but that yaml file will not be in git, so the user will not have to worry about it being changed.

avi-biton (approved):
lgtm

PR #1013, opened 2024-12-31
fix: kube-lint issues

yftacherzog:


dependencies/tekton-results/kustomization.yml
```diff
@@ -26,6 +26,27 @@ patches:
     target:
       kind: StatefulSet
       name: tekton-results-postgres
+  - patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1liveness-port
```
avi-biton, line 31: Shouldn't it be `/metadata/annotations/ignore-check.kube-linter.io/liveness-port` ?

yftacherzog: `~1` is the escape character for `/` (I'm not sure this is for YAML, or for something else but this is what works in this case), so this eventually translates to:
```yaml
metadata:
  annotations:
    ignore-check.kube-linter.io/liveness-port: <what's in value>
```

avi-biton: I checked it myself and it seems to work fine.
Thanks for the explanation.

avi-biton (approved):
lgtm

PR #1023, opened 2025-01-01
add go proxy test placeholder

avi-biton:
add a placeholder in actions for go proxy tests:

- Add a dummy test in go as a placeholder to proxy tests
- Update some parameters we need for the tests

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -95,6 +95,10 @@ jobs:
         run: |
           ./deploy-test-resources.sh
 
+      - name: Run Go Tests
+        run: |
+          go test ./test/go-tests/
+
       - name: Prepare resources for E2E tests
         env:
           APP_ID: ${{ secrets.APP_ID }}
@@ -113,7 +117,7 @@ jobs:
           QUAY_DOCKERCONFIGJSON: ${{ secrets.QUAY_DOCKERCONFIGJSON }}
         run: |
           ./test/e2e/run-e2e.sh
-      
+
       - name: Generate error logs
         if: ${{ !cancelled() }}
         run: |
--- a/dependencies/dex/config.yaml
+++ b/dependencies/dex/config.yaml
@@ -9,6 +9,7 @@ web:
   tlsKey: /etc/dex/tls/tls.key
 oauth2:
   skipApprovalScreen: true
+  passwordConnector: local
 staticClients:
 - id: oauth2-proxy
   redirectURIs:
--- a/konflux-ci/ui/core/proxy/proxy.yaml
+++ b/konflux-ci/ui/core/proxy/proxy.yaml
@@ -215,6 +215,7 @@ spec:
           - "true"
           - --whitelist-domain
           - localhost:9443
+          - --skip-jwt-bearer-tokens
         ports:
           - containerPort: 6000
             name: web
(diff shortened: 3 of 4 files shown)


test/test-proxy.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -e
```
yftacherzog: The original idea I recall was to have those tests in go rather than bash. @gbenhaim what do you think?

avi-biton: Is there a specific reason to do it with go?

yftacherzog: I mention go in this case because this is mostly what we're using in our repos nowadays. Could be something else, but I think it should not be bash.

Looking back at the team's experience (CPaaS days, etc.), there was much effort put into moving tests from bash to proper testing frameworks (mostly Python at that time).

The way I see it, Go has proper test frameworks that report errors in a better way and allow better checks/tests. Error capturing and handling is not that good with bash. Every now and then it happens that we think we assert one thing in bash and in reality we're not actually checking what we thought we're checking.

All in all, it all depends on what else we'd like to be tested here. If the tests in this PR is more or less what we want to have in the near future then it might not be worth the effort, but if @gbenhaim had something additional in mind then it might make sense to do it with a proper test framework.

gbenhaim: I agree with @yftacherzog. we should use a language that has a proper testing framework. I think that go is a good choice.

test/test-proxy.sh
```diff
@@ -0,0 +1,52 @@
...
+    -H "Authorization: Basic $auth_header" \
+    -H "Content-Type: application/x-www-form-urlencoded" \
+    --data-urlencode 'grant_type=password' \
+    --data-urlencode 'scope=openid profile email' \
+    --data-urlencode "username=$username" \
+    --data-urlencode "password=$password" | jq -r '.id_token' )
+
+    # Send a request for resources using the extracted id, request should succeed
+    echo "Get resource with authorization" >&2
+    secrets=$(curl -I -k -L -X GET "$secrets_url" -H "Authorization: Bearer $id_token")
```
gbenhaim: all the endpoints (location blocks) mentioned in the nginx config should be tested.
note that there are to location blocks that doesn't require auth (/ and /health)

gbenhaim (changes requested):

test/test-proxy.sh
avi-biton: I think that if we are only testing our ability to reach the endpoints with/without authorization, Bash can be sufficient, if we will add other tests then we might want to switch to Go

gbenhaim: We are going to have multiple tests. one for each endpoint. it fits perfectly to table testing. the test framework will ensure to run all the tests (even if some of them failed) and provide formatted test output and the correct return code.

avi-biton:
@yftacherzog @gbenhaim 
I fixed the description, I removed the bash test file and instead created a dummy go test as a placeholder in the workflow, and updated the needed parameters for our tests.

.github/workflows/sanity-test.yml
```diff
@@ -95,6 +95,10 @@ jobs:
         run: |
           ./deploy-test-resources.sh
 
+      - name: Test Proxy
+        run: |
+          go test ./test/proxy-tests/
```
yftacherzog: I wonder if we should make it more general.
It will allow us to add more tests in the future without having to go through this step again.

```suggestion
      - name: Run go tests
        run: |
          go test ./test/go-tests/
```

gbenhaim (approved):

yftacherzog (approved):
lgtm

Issue #299, opened 2024-08-15
Test the proxy endpoints require authentication


gbenhaim:
Create a test for the nginx proxy that verifies that endpoints that requires authentication are configured properly. This can be done by deploying Konflux, try to reach each endpoint that requires authentication (without authenticating) and check that the requests fails because of missing authentication, then repeat the process (with authentication) and verify that the calls succeed.

yftacherzog:
Do we need anything more implemented on the UI side to start this task?

gbenhaim:
@yftacherzog no.

yftacherzog:
The idea is to have a go module that runs tests, and have the [sanity tests](https://github.com/konflux-ci/konflux-ci/blob/main/.github/workflows/sanity-test.yml) run this module.

yftacherzog:
@avi-biton can we close this one? I don't recall if this was not closed intentionally or not.

avi-biton:
Closed by #1043

PR #1043, opened 2025-01-05
Test the proxy endpoints require authentication

avi-biton:
Add go tests to test proxy endpoints

#299


avi-biton:
@gbenhaim @yftacherzog I added some endpoint tests using `DescribeTable`  to parametrize the tests and reuse them.
The k8sclient is created in the test_suite and not in the test itself so it will be reusable for other tests, instead of redefining it in every test.
Do you think of other endpoints or other tests we should add?

yftacherzog (approved):
lgtm

PR #1130, opened 2025-01-21
ui-nslister: improve try_files for @namedlocation

filariow:
- **improve try_files to @namedlocation**
- **fix formatting**


yftacherzog:
@filariow for the tests to run without applying the ok-to-test label, you'd need to make you membership in the konflux-ci org public.

PR #1136, opened 2025-01-21
update client creation for tests

avi-biton:


test/go-tests/suite_test.go
```diff
@@ -17,12 +17,15 @@ func TestGoTests(t *testing.T) {
 
 // Create a new Kubernetes client using local config file
 func CreateK8sClient() (*kubernetes.Clientset, error) {
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return nil, err
+	kubeconfig := os.Getenv("KUBECONFIG")
```
yftacherzog, line 20: What's the benefit of the new approach over the existing one? Does this change the way this test suite run in envs other than CI?

yftacherzog (approved):

PR #1139, opened 2025-01-22
fix: flaky proxy test

yftacherzog:
Test suite will sometimes fail after fetching the token as it appears to be empty. This is caused by failed authentication, which seems to be related to a inconsistently-escaped characters when Dex is comparing the client secret.

To address this, we remove the characters that should cause such inconsistencies.


avi-biton (approved):
lgtm

PR #1148, opened 2025-01-23
chore(KONFLUX-6528): update dex with renovate

yftacherzog:
Track Dex updates using renovate.

Closes: #1142

### DIFF
--- a/dependencies/dex/dex.yaml
+++ b/dependencies/dex/dex.yaml
@@ -41,7 +41,7 @@ spec:
     spec:
       serviceAccountName: dex # This is created below
       containers:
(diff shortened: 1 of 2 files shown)


avi-biton:
Shouldn't we add `ghcr.io/dexidp/dex` to the renovate.json file?

yftacherzog:
> Shouldn't we add `ghcr.io/dexidp/dex` to the renovate.json file?

As far as I tested it locally, this is provided out-of-the-box with Renovate and does not require additional customizations.

avi-biton (approved):
/lgtm

PR #1229, opened 2025-02-04
add TLS support in namespace-lister

filariow:

### DIFF
--- a/konflux-ci/namespace-lister/certificate.yaml
+++ b/konflux-ci/namespace-lister/certificate.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: namespace-lister
+  namespace: namespace-lister
+spec:
+  isCA: true
+  subject:
+    organizations:
+      - konflux
+  dnsNames:
+  - localhost
+  - namespace-lister.namespace-lister
+  issuerRef:
+    kind: ClusterIssuer
+    name: ca-issuer
+  secretName: namespace-lister-tls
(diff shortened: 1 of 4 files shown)


konflux-ci/namespace-lister/certificate.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: namespace-lister
+  namespace: namespace-lister
+spec:
+  isCA: true
+  subject:
+    organizations:
+      - konflux
+  dnsNames:
+  - localhost
+  - namespace-lister.namespace-lister
```
sadlerap: I think these need to be changed.
```suggestion
  - namespace-lister.namespace-lister.svc
  - namespace-lister.namespace-lister.svc.cluster.local
```

konflux-ci/namespace-lister/certificate.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: namespace-lister
+  namespace: namespace-lister
+spec:
+  isCA: true
```
sadlerap: Why do we need to be a certificate authority?  We're not going to be issuing certificates of our own.

konflux-ci/namespace-lister/certificate.yaml
```diff
@@ -0,0 +1,18 @@
...
+  subject:
+    organizations:
+      - konflux
```
sadlerap: I don't think this field is necessary, especially since we shouldn't be issuing certificates.

sadlerap (changes requested):

sadlerap (approved):

PR #1257, opened 2025-02-06
bump namespace lister

filariow:


yftacherzog (approved):
@filariow are we ok with having this image covered by renovate, so that images are updated automatically?

filariow:
> @filariow are we ok with having this image covered by renovate, so that images are updated automatically?

yes, that's a great idea. let's add the namespace-lister to renovate!

PR #1285, opened 2025-02-11
add a default releaseserviceconfig

scoheb:
- with https://github.com/konflux-ci/release-service/pull/669 and https://github.com/redhat-appstudio/infra-deployments/pull/5508 we need to ensure that a default ReleaseServiceConfig is present when the release controller is running.

### DIFF
--- a/konflux-ci/release/core/kustomization.yaml
+++ b/konflux-ci/release/core/kustomization.yaml
@@ -4,6 +4,7 @@ resources:
   - https://github.com/konflux-ci/release-service/config/default?ref=5954f10a8c86134fca5f70fc1d9898254ee1f082
   - release-pipeline-resources-clusterrole.yaml
   - release-service-config-rbac.yaml
+  - release-service-config.yaml
 
 images:
   - name: quay.io/konflux-ci/release-service
--- a/konflux-ci/release/core/release-service-config.yaml
+++ b/konflux-ci/release/core/release-service-config.yaml
@@ -1,0 +1,6 @@
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ReleaseServiceConfig
+metadata:
+  name: release-service-config
+spec:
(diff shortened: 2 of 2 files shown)


yftacherzog:
@scoheb I think the configs pulled by kustomize from https://github.com/konflux-ci/release-service/config/default?ref=5954f10a8c86134fca5f70fc1d9898254ee1f082 already contains the same resource.

psturc:
@yftacherzog I can confirm that the missing `ReleaseServiceConfig` CR is causing a test failure in [this PR](https://github.com/konflux-ci/konflux-ci/pull/1313) for updating release-service

Adding that resource in this PR won't work though, because of the ID conflict with the [current release-service ref](https://github.com/konflux-ci/konflux-ci/pull/1285/files#diff-69d04ef851701dfdf5188683cd6da4679021278bb536c78564cc0b6fee329713R4)

So I cherry-picked Scott's commit in [that other PR](https://github.com/konflux-ci/konflux-ci/pull/1313/commits), which should resolve the test failure and the ID conflict as well

psturc:
https://github.com/konflux-ci/konflux-ci/pull/1313 was successfully merged, I think we can close this PR

yftacherzog:
> @yftacherzog I can confirm that the missing `ReleaseServiceConfig` CR is causing a test failure in [the PR for updating release-service](https://github.com/konflux-ci/konflux-ci/pull/1313)
> 
> Adding that resource in this PR won't work though, because of the ID conflict with the [current release-service ref](https://github.com/konflux-ci/konflux-ci/pull/1285/files#diff-69d04ef851701dfdf5188683cd6da4679021278bb536c78564cc0b6fee329713R4)
> 
> So I cherry-picked Scott's commit in #1313, which should resolve the test failure and the ID conflict as well

Thank you @psturc!

PR #1402, opened 2025-02-26
update namespace-lister image name to be renovate friendly

filariow:

### DIFF
--- a/konflux-ci/namespace-lister/kustomization.yaml
+++ b/konflux-ci/namespace-lister/kustomization.yaml
@@ -9,9 +9,9 @@ resources:
 - certificate.yaml
 namespace: namespace-lister
 images:
-- name: namespace-lister
+- name: quay.io/konflux-ci/namespace-lister
   newName: quay.io/konflux-ci/namespace-lister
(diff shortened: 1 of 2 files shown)


konflux-ci/namespace-lister/kustomization.yaml
```diff
@@ -9,7 +9,7 @@ resources:
 - certificate.yaml
 namespace: namespace-lister
 images:
-- name: namespace-lister
+- name: quay.io/konflux-ci/namespace-lister
   newName: quay.io/konflux-ci/namespace-lister
   newTag: 3f758b0f004adb5c7b94d079ddc485573d9c6823
```
yftacherzog: I tried it with `newTag` and it doesn't seem to work.
It might be because renovate expects that tags can be compared (semver-ish).
Digest seems to work for renovate here.
```suggestion
  digest: sha256:4f33da08444567b757e140e7cc532cbe347b9d796c487c4526a2aabc28135eed
```

filariow: great, thanks! I committed the suggestion

yftacherzog:
lgtm. Can you squash the commits?

PR #1530, opened 2025-03-12
feat: add PAC_WEBHOOK_URL to allow UI builds

rh-hemartin:
This PR adds a `kubectl` command to set the `PAC_WEBHOOK_URL` for the `build-service` so users can trigger builds directly from the UI. Currently if you try to trigger a build from the UI the `build-service` tries to reach for an OpenShift Route that does not exist, so this change is required.

PD: I'm not sure about the PR format, if any, let me know.

### DIFF
--- a/deploy-konflux.sh
+++ b/deploy-konflux.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-script_path="$(dirname -- "${BASH_SOURCE[0]}")" 
+script_path="$(dirname -- "${BASH_SOURCE[0]}")"
 
 main() {
     echo "Deploying Konflux" >&2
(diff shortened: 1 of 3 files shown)


deploy-konflux.sh
```diff
@@ -36,6 +36,9 @@ deploy() {
 
     # The build-service depends on CRDs from the release-service
     retry kubectl apply -k "${script_path}/konflux-ci/build-service"
+    kubectl set --namespace build-service env \
+        deployment/build-service-controller-manager \
+        PAC_WEBHOOK_URL=http://pipelines-as-code-controller.pipelines-as-code.svc.cluster.local:8080
```
yftacherzog: Can we patch the deployment manifest instead?
I think it will be easier to track.

rh-hemartin: Done, I submitted the change.

yftacherzog (approved):

ralphbean:
FYI, this one broke the Fedora cluster downstream, fixed here -> https://gitlab.com/fedora/infrastructure/konflux/infra-deployments/-/commit/fb541a696a72bde6c3e20ae5deff4b10f1c9b4e3

PR #1560, opened 2025-03-14
Fix revision for Release Service repo

rh-hemartin:
This PR changes the revision specified for the pipelines on the example ReleasePlanAdmissions because the SHA specified is not on the repository. See https://github.com/konflux-ci/release-service-catalog/commit/b0586bdee6e5721d773f4b5be6ca8cd8f70ddd32 and the warning at the top. The current SHA references a fix for a PR that is already merged, so I searched for the corresponding PR and used the SHA created by that PR when merged on their default branch.

cc @johnbieren @scoheb

### DIFF
--- a/test/resources/demo-users/user/managed-ns1/rpa.yaml
+++ b/test/resources/demo-users/user/managed-ns1/rpa.yaml
@@ -25,7 +25,7 @@ spec:
         - name: url
           value: "https://github.com/konflux-ci/release-service-catalog.git"
         - name: revision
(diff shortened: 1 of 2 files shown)


scoheb (approved):

johnbieren (commented):
I don't really know the point of these files, but why not just use one of our branch names instead? There could be breaking changes in the operator down the line that breaks an old commit in the catalog. This would get fixed for free with a branch like `production`

PR #1569, opened 2025-03-14
Create the managed namespaces by default

ralphbean:
One of the ReleasePlans expects that they are present.

I ran into this in a practical way while working on https://github.com/konflux-ci/konflux-ui/pull/153

### DIFF
--- a/test/resources/demo-users/user/kustomization.yaml
+++ b/test/resources/demo-users/user/kustomization.yaml
@@ -3,4 +3,6 @@ kind: Kustomization
 resources:
   - ns1
   - ns2
+  - managed-ns1
+  - managed-ns2
   - cluster-rbac.yaml


test/resources/demo-users/user/kustomization.yaml
```diff
@@ -3,4 +3,6 @@ kind: Kustomization
 resources:
   - ns1
   - ns2
+  - managed-ns1
+  - managed-ns2
```
yftacherzog, line 7: At least for `managed-ns2`, that resource contains the RPA which needs to be edited by the user before it's deployed to the cluster.
the way things are deployed needs to change a bit to keep things coherent.

ralphbean: Ah, I see. Ok, then we can drop this.

Issue #812, opened 2024-11-27
Sanity check does not validate all available build PipelineRuns


psturc:
More info in [this PR](https://github.com/konflux-ci/konflux-ci/pull/803#issuecomment-2503560205)

The goal is to update e2e tests in a way that the sanity check can validate all build PipelineRuns defined [here](https://github.com/konflux-ci/konflux-ci/blob/main/konflux-ci/build-service/core/build-pipeline-config.yaml)

PR #1626, opened 2025-03-20
test: validate docker-build-oci-ta pipelines

psturc:
closes https://github.com/konflux-ci/konflux-ci/issues/812

depends on https://github.com/konflux-ci/e2e-tests/pull/1550 (should be merged after that one is in)

### DIFF
--- a/test/e2e/customize-docker-pipeline.sh
+++ b/test/e2e/customize-docker-pipeline.sh
@@ -1,7 +1,43 @@
-#!/bin/bash -e
-set -x
+#!/bin/bash
+set -euo pipefail
 
 script_path="$(dirname -- "${BASH_SOURCE[0]}")"
+
+update_pipeline() {
+    local pipeline_name=$1
+    local tmp_pipeline_path="/tmp/customized-${pipeline_name}.yaml"
+    local original_pipeline_bundle_ref
+
+    original_pipeline_bundle_ref=$(yq ".data[\"config.yaml\"]" konflux-ci/build-service/core/build-pipeline-config.yaml | yq ".pipelines[] | select(.name == \"${pipeline_name}\").bundle")
+    tkn bundle list --remote-skip-tls "$original_pipeline_bundle_ref" -o yaml > "${tmp_pipeline_path}"
+
+    for task_name in $(yq "(.spec.tasks[] | .name)" "${tmp_pipeline_path}");
+    do
+        local original_task_bundle_ref
+        local tmp_task_path="/tmp/customized-task.yaml"
+        local customized_task_image_ref_localhost="localhost:30001/test/test:customized-task-${task_name}-${pipeline_name}"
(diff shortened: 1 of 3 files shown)


yftacherzog:
> Hi @ralphbean @yftacherzog
> 
> following up on [this comment](https://github.com/konflux-ci/konflux-ci/pull/803#issuecomment-2503560205) - in this PR I've been trying to update the test to validate 2 pipelines: `docker-build` and `docker-build-oci-ta` (instead of only the `docker-build`, which is currently being tested in PR checks of this repo).
> 
> All the time I'm hitting an issue with insufficient memory in github actions, even after [customizing all build pipelines' Tasks to request a minimum cpu/memory](https://github.com/konflux-ci/konflux-ci/pull/1626/files#diff-fd17c12f6a9db21ecea8c3d688e76cad475c81ac6b6ba0ef6eaa958c5bee288eR24-R27) and [increasing the system-reserved memory to 12Gi](https://github.com/konflux-ci/konflux-ci/pull/1626/files#diff-b085642d962662414dd873573d4e087a3c45e50a036271863583244f93fa04a6R12).
> 
> I'm at the stage where I think that it's not worth to test both of those pipelines (even if we'd overcome the memory limitation, the test would take ~10 minutes more).
> 
> So if we're not going to increase the compute resources on github actions' node size (it'd be no longer free), I'd suggest to update our test to validate only `docker-build-oci-ta` pipeline (the default one) for now in PR checks of this repo (I think it's not a bad situation, because other build pipelines are still being tested in build-definitions PR checks).
> 
> Wdyt?

@psturc thank you for picking this up!

I think it would be fine to only have the oci-ta variant for now (if we decide we need both eventually, then we can create another workflow that will use additional runner. No?).

In parallel, we're (@amisstea is) also working on making this oci-ta pipeline compatible with the in-cluster registry, so it should be what we will use for all our documented use cases.

All in all, having the oci-ta variant tested here instead of the non-oci-ta variant will be a great improvement!

cc @gbenhaim @avi-biton

amisstea:
@psturc the latest bundle of the `docker-build-oci-ta` pipeline now (within the last ~24hours) fully supports the internal registry in case you want to start testing with that.

yftacherzog (approved):
lgtm. Thank you!

yftacherzog:
@psturc is it ready to be merged?

psturc:
@yftacherzog yes I had to update the e2e image after https://github.com/konflux-ci/e2e-tests/pull/1550 was merged, so let's wait for the e2e to finish and then merge 🚀

yftacherzog:
> @yftacherzog yes I had to update the e2e image after [konflux-ci/e2e-tests#1550](https://github.com/konflux-ci/e2e-tests/pull/1550) was merged, so let's wait for the e2e to finish and then merge 🚀

Alright. Feel free to merge it once you're ready.

PR #1726, opened 2025-03-28
remove workspaces manager

filariow:
superseded by konflux-ci/namespace-lister


sadlerap (approved):

yftacherzog:
@filariow Does this affect the way users should be added? Do we need to change the test user data or the documentation?

filariow:
@yftacherzog this doesn't affect how users are added. Test user data is good. Documentation could be enhanced, but I'll leave that for another PR.

yftacherzog (dismissed):

yftacherzog:
@filariow I ran the deployment scripts locally a few times.
First 3 times, it failed 2 out of 3, but since then I was not able to see it fails.
When it did fail, this was the output:
```
Waiting for Konflux to be ready
tektonconfig.operator.tekton.dev/config condition met
deployment.apps/build-service-controller-manager condition met
deployment.apps/cert-manager condition met
deployment.apps/cert-manager-cainjector condition met
deployment.apps/cert-manager-webhook condition met
deployment.apps/trust-manager condition met
deployment.apps/dex condition met
deployment.apps/integration-service-controller-manager condition met
deployment.apps/registry condition met
timed out waiting for the condition on deployments/proxy
timed out waiting for the condition on deployments/coredns
...
timed out waiting for the condition on deployments/tekton-triggers-core-interceptors
timed out waiting for the condition on deployments/tekton-triggers-webhook
Deployment failed
```

The first time I saw it failing, I ran deploy-konflux.sh again and it passed, which made me think it just took it longer to start than the timeout allowed.

The `kubectl wait` command is waiting on all deployments on the cluster together, it lists them and then reports on every item on the list once it's available and waits on the next one on the list (which might be available already and was not checked yet).

I started watching how much time it takes the proxy deployment to become available, and it seems to take around 2:30 minutes.

I tend to think that sometimes it takes longer than the timeout (3 min), so I think we should increase that timeout.

I suggest we increase [this one](https://github.com/konflux-ci/konflux-ci/blob/main/wait-for-all.sh#L5) to 240 seconds.

yftacherzog (approved):

PR #1751, opened 2025-03-31
feat: add sealights integration to get coverage numbers about konflux and all core products

flacatus:
Creates script that will get coverage metrics about Konflux CI product and in the future we want to get it at PR level when sealights is ready to do it.

### DIFF
--- a/test/e2e/coverage-metrics/create-test-session.sh
+++ b/test/e2e/coverage-metrics/create-test-session.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -euo pipefail
+
+# Ensure necessary environment variables are set
+export SEALIGHTS_TOKEN="${SEALIGHTS_TOKEN:-""}"
+export SEALIGHTS_LAB_ID="${SEALIGHTS_LAB_ID:-""}"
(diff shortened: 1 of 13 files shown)


flacatus:
cc @gbenhaim @psturc @yftacherzog

test/e2e/coverage-metrics/create-test-session.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
psturc, line 1: these new scripts are currently not used in the workflow right? is the workflow going to be updated in another PR?

flacatus: This script will be used in a workflow that will run tests periodically. Once we decide are stable and agree to run them in prs then we can add them to prs checks

psturc: could you link the workflow run where these scripts were used? (I guess in your fork of this repo?)

flacatus: Unfortunetly I deleted the workflows in my main fork

PR #1837, opened 2025-04-06
fix: kyverno policy for steps with undefined resources

yftacherzog:
The policy will fail to be deployed if the pod it captures had some containers with resource requests undefined. This change should address that.


avi-biton (approved):
lgtm

PR #1840, opened 2025-04-06
chore: use kyverno for ci workaround

yftacherzog:
Replace the script used for reducing task resources in CI with a kyverno policy.

### DIFF
--- a/dependencies/kyverno/policy/e2e-reduce-resources.yaml
+++ b/dependencies/kyverno/policy/e2e-reduce-resources.yaml
@@ -1,0 +1,42 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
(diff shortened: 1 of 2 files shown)


yftacherzog:
@psturc I get the feeling that this affects performance again. I wonder if there's a way to make the policy more efficient (logic, compute resources,  etc.). I'll give it some more  thought, and then we'll decide whether to use it or not.

yftacherzog:
> @psturc I get the feeling that this affects performance again. I wonder if there's a way to make the policy more efficient (logic, compute resources, etc.). I'll give it some more thought, and then we'll decide whether to use it or not.

I created [another PR](https://github.com/konflux-ci/konflux-ci/pull/1844) that seems to provide better performance consistently (seems similar to the original workaround). I'll be closing this one.

PR #1975, opened 2025-04-17
fix: smee client liveliness probe

yftacherzog:
Probe was failing, as it turns out the command field should be provided as a list of words.

Error message: OCI runtime exec failed: exec failed: unable to start container process: exec: "echo \"Service up\"": executable file not found in $PATH

### DIFF
--- a/dependencies/smee/smee-client.yaml
+++ b/dependencies/smee/smee-client.yaml
@@ -30,7 +30,8 @@ spec:
           livenessProbe:
             exec:
               command:
-                - echo "Service up"
+                - echo
+                - "Hello smee client"
             initialDelaySeconds: 30
             periodSeconds: 5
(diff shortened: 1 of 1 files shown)


yftacherzog:
@rubenmgx fyi

avi-biton (approved):

yftacherzog:
@rubenmgx the probe now works, but I don't think it's solving the issue you wanted to address. As far as I understand, for times in which the pods is not responding to events (e.g. after laptop wakes from sleep), the pod is alive enough to allow the kubelet to run the probe (don't see failures or restarts), but it still won't track the channel events. We probably need a different approach.

rubenmgx:
@yftacherzog thanks for letting me know! I will check that out

Issue #2031, opened 2025-04-23
Use the new label format for namespaces


gbenhaim:
We recently changed the label for marking Konflux user namespaces (`konflux-ci.dev/type=tenant`), namespace-lister and the test resources (https://github.com/konflux-ci/konflux-ci/tree/main/test/resources/demo-users/user) should be configured accordingly.

gbenhaim:
@sadlerap can you please take care of it?

PR #2064, opened 2025-04-25
use konflux-ci.dev/type=tenant as user ns label

sadlerap:
The label for user namespaces has moved from `konflux.ci/type=user` to `konflux-ci.dev/type=tenant`.  Update namespace-lister, documentation, and tests to use this new label.

### DIFF
--- a/konflux-ci/namespace-lister/kustomization.yaml
+++ b/konflux-ci/namespace-lister/kustomization.yaml
@@ -14,6 +14,12 @@ images:
   name: quay.io/konflux-ci/namespace-lister
   newName: quay.io/konflux-ci/namespace-lister
 patches:
+- path: ./patches/with_cachenamespacelabelselector.yaml
+  target:
+    group: apps
+    kind: Deployment
+    name: namespace-lister
+    namespace: namespace-lister
 - path: ./patches/with-header-auth-impersonate-user.yaml
   target:
     group: apps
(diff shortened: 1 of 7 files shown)


sadlerap:
Fixes #2031.

@filariow would appreciate review on the changes to namespace-lister.

konflux-ci/namespace-lister/kustomization.yaml
```diff
@@ -14,6 +14,12 @@ images:
   name: quay.io/konflux-ci/namespace-lister
   newName: quay.io/konflux-ci/namespace-lister
 patches:
+- path: ./patches/with_cachenamespacelabelselector.yaml
+  target:
+    group: apps
+    kind: Deployment
+    name: namespace-lister
+    namespace: namespace-lister
```
yftacherzog: Can we put it directly in `deployment.yaml` instead?

sadlerap: Fixed in be718af68395f0a0802b44791c3a8ce67d03b61b.

yftacherzog (approved):
lgtm. @filariow can you take a look as well?

filariow:
@yftacherzog @gbenhaim will this change impact anyone? As an example I'm thinking about the fedora cluster

filariow (approved):
/lgtm

Issue #2174, opened 2025-05-08
Update for e2e tests is failing CI


yftacherzog:
https://github.com/konflux-ci/konflux-ci/pull/2153

I think a recent change in the e2e tests might not be compatible with our CI environment.

flacatus:
@tisutisu Can you take a look why this changes https://github.com/konflux-ci/e2e-tests/commit/fb417d9208db377c8281ad2f7484846d67ca3873 affect the runs? cc @tnevrlka

tisutisu:
> [@tisutisu](https://github.com/tisutisu) Can you take a look why this changes [konflux-ci/e2e-tests@fb417d9](https://github.com/konflux-ci/e2e-tests/commit/fb417d9208db377c8281ad2f7484846d67ca3873) affect the runs? cc [@tnevrlka](https://github.com/tnevrlka)

For #2153 to work, we have to update the build-service component [here](https://github.com/konflux-ci/konflux-ci/blob/457d4a22517bd6fc7ccd9e74958d1a635f9dced6/konflux-ci/build-service/core/kustomization.yaml#L4) and [here](https://github.com/konflux-ci/konflux-ci/blob/457d4a22517bd6fc7ccd9e74958d1a635f9dced6/konflux-ci/build-service/core/kustomization.yaml#L13) to the [this commit](https://github.com/konflux-ci/build-service/commit/51a629aed4bd36b6321b23e4ca54d302079fd8f8) atleast or latest one

PR #2187, opened 2025-05-09
fix: update build-service and e2e-tests component together

tisutisu:
closes: https://github.com/konflux-ci/konflux-ci/issues/2174

### DIFF
--- a/konflux-ci/build-service/core/kustomization.yaml
+++ b/konflux-ci/build-service/core/kustomization.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - https://github.com/konflux-ci/build-service/config/default?ref=3087d89e119fed73130275081b4d58e6235ae0db
+  - https://github.com/konflux-ci/build-service/config/default?ref=9bf819f025fdddd2fef7abd020979af294b425b2
   - build-pipeline-config.yaml
   - build-config-rolebinding.yaml
 
@@ -10,7 +10,7 @@ namespace: build-service
 images:
   - name: quay.io/konflux-ci/build-service
     newName: quay.io/konflux-ci/build-service
-    newTag: 3087d89e119fed73130275081b4d58e6235ae0db
+    newTag: 9bf819f025fdddd2fef7abd020979af294b425b2
 
 patches:
   - target:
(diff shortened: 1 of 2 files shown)


yftacherzog:
@gbenhaim I see that this PR is failing with this on the build-service logs:
```
Waiting for appstudio-pipelines-runner Cluster Role
```
This is coming from a recent change: https://github.com/konflux-ci/build-service/pull/425/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R403

Looking at the way this cluster role is defined, it seems to be related to OpenShift-specific resources.

Do you know if we really need it? If so, I'm trying to figure out whether that recent change is upstream-compatible.

yftacherzog:
@mmorhun Hi!
The build service is failing to start when upgrading it to the changes made in [this PR](https://github.com/konflux-ci/build-service/pull/425/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R403).

We're able to work around that issue by introducing an empty `appstudio-pipelines-runner` cluster role to our cluster (see [this other PR](https://github.com/konflux-ci/konflux-ci/pull/2195/files)). This allows build-service to start, but it also allows e2e tests to pass.

It makes me wonder whether the check for that cluster role is relevant regardless of the environment build-service is deployed to.

What do you say?

mmorhun (approved):

mmorhun:
Hello @yftacherzog 
I've added that check because now Build Service manages build pipeline Service Accounts and it cannot properly function without appropriate Cluster Role for builds. However, I haven't changed anything in the regard of Konflux deployment. To have a successful build Konflux admin would still need to create the build pipeline `appstudio-pipelines-runner` Cluster Role as before.

yftacherzog:
> Hello @yftacherzog I've added that check because now Build Service manages build pipeline Service Accounts and it cannot properly function without appropriate Cluster Role for builds. However, I haven't changed anything in the regard of Konflux deployment. To have a successful build Konflux admin would still need to create the build pipeline `appstudio-pipelines-runner` Cluster Role as before.

Interesting. We never had it in konflux-ci/konflux-ci. Looking at the role, it seems related to OpenShift stuff. We're running on Kind. Might that be the reason we did not encounter any issues with it being missing?

yftacherzog:
@tisutisu I merged another PR that includes your changes and also the cluster role that its absence is making this PR fail CI.
I think this PR and the issue it was opened for can be closed.
Thank you!

PR #2532, opened 2025-06-19
fix: add ttlSecondsAfterFinished

yftacherzog:
For some reason, kube-linter started complaining about ttlSecondsAfterFinished not being set for a job.
Doesn't seem as anything related changed.


avi-biton (approved):
/lgtm

PR #2616, opened 2025-07-01
chore(KFLUXVNGD-363): cert-manager integration for Internl-Changes

avi-biton:
- Added squid-proxy namespace
- Certificate for squid-proxy was created.

The certificate uses ca-issuer as the ClusterIssuer


dependencies/squid-proxy/certificate.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - localhost
+  isCA: true  
+  subject:
+    organizations:
+      - konflux
+    organizationalUnits:
+      - Squid Proxy
+  issuerRef:
+    kind: ClusterIssuer
+    name: ca-issuer
```
ifireball: I think this effectively make this CA a subordinate CA to the existing cluster-level "ca-issuer" I'm not sure we want that

avi-biton: self-signed-cluster-issuer -> creates root-secret (cluster CA)
ca-issuer -> uses root-secret to sign other certificates
root-secret -> is already in the trust bundle

Do you think we need to update the issuerRef to [self-signed-cluster-issuer](https://github.com/konflux-ci/konflux-ci/blob/main/dependencies/cluster-issuer/self-signed-cluster-issuer.yaml)?
Note that if we will, we will have to update the [bundle](https://github.com/konflux-ci/konflux-ci/blob/main/dependencies/registry/trust-bundle.yaml) to add the secret that will be created by squid-proxy.

ifireball: I want to have separate CAs for different things - it lowers the dangers posed of leaking the private keys - this does mean we will need to work harder to update the trust bundles.

Lets try not to use cluster issuers unless we find reasons why we have to. In the case of Squid I expect all the certificate signings will happen in the "proxy" namespace, so it seems to me like we may not need cluster issuers at all.

deploy-deps.sh
```diff
@@ -151,6 +153,12 @@ deploy_kyverno() {
     kubectl apply -k "${script_path}/dependencies/kyverno/policy"
 }
 
+deploy_proxy() {
+    kubectl apply -k "${script_path}/dependencies/proxy"
+    retry "kubectl wait --for=condition=Synced --timeout=120s bundle/proxy-ca-bundle -n proxy" \
```
yftacherzog, line 158: I think we should wait here on a pod or at least make sure this wait statement is working correctly. It might be that it returns too early and this is why we sometime fail on a later wait statement when we wait for all deployments.

yftacherzog:
@avi-biton should we keep this PR open?

PR #2661, opened 2025-07-07
Give user admin the ability to manage serviceaccounts

ralphbean:


ralphbean:
In the Fedora cluster, @jhutar was trying to create a token for a service acccount and couldn't.

I assume this is something we would expect a namespace admin to be able to do in general.

konflux-ci/rbac/core/rbac.yaml
```diff
@@ -178,15 +178,20 @@ rules:
   - verbs:
       - get
       - list
-      - watch
+      - create
```
yftacherzog: Does create give you watch? I.e. is create enough or we need both?

konflux-ci/rbac/core/rbac.yaml
```diff
@@ -178,15 +178,20 @@ rules:
...
     resources:
       - serviceaccounts
-    resourceNames:
-      - appstudio-pipeline
+  - verbs:
+      - create
+    apiGroups:
+      - ""
+    resources:
+      - serviceaccounts/token
```
yftacherzog: What's the reason for this change?

ralphbean: It's the thing that brought me here. In the Fedora cluster, @jhutar was trying to create a token for an SA but failed. I found this in the admin rbac from the RH clusters and assumed it was related.

yftacherzog: alright

konflux-ci/rbac/core/rbac.yaml
ralphbean: Hm, added `watch` back.

PR #2738, opened 2025-07-16
feat: prepare for accessing konflux from public IP

psturc:
https://issues.redhat.com/browse/KFLUXDP-379

we're in the middle of the process of migrating to a "konflux on kind" to be used as a primary testing environment on konflux components' PR tests and in order to get all tests to work, we need to
1. expose pipelines-as-code API endpoint publicly (to be used tests that are using Gitlab)
2. update oauth2-proxy deployment so the UI login works when the service is exposed publicly (via public IP)


konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -186,6 +186,13 @@ spec:
           - https://localhost:9443/oauth2/callback
           - --oidc-issuer-url
           - https://localhost:9443/idp/
+          - --skip-oidc-discovery
```
psturc, line 189: this is needed so oauth2-proxy won't try to connect to the external IP for discovery during deployment (when the service UI is exposed via public IP, i.e. `https://<public-ip>:9443`), otherwise the deployment fails

konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -186,6 +186,13 @@ spec:
...
+          - --login-url
+          - https://localhost:9443/idp/auth
+          - --redeem-url
+          - https://dex.dex.svc.cluster.local:9443/idp/token
+          - --oidc-jwks-url
+          - https://dex.dex.svc.cluster.local:9443/idp/keys
```
psturc, line 195: these need to be explicitly defined when we're deploying UI for public access, otherwise it's trying to communicate with dex on `https://<public-ip>:9443` which obviously does not work

dependencies/pipelines-as-code/kustomization.yml
```diff
@@ -89,3 +89,16 @@ patches:
...
+  - patch: |
+      - op: add
+        path: /spec/ports/0/nodePort
+        value: 30012
+      - op: add
+        path: /spec/ports/0/port
+        value: 8080
+      - op: replace
+        path: /spec/type
+        value: NodePort
```
psturc, line 101: When deployed, PaC endpoint becomes available at `http://<public-ip>:8180`

kind-config.yaml
```diff
@@ -14,6 +14,11 @@ nodes:
...
     hostPort: 8888
     protocol: TCP
+    # UI
   - containerPort: 30011
     hostPort: 9443
     protocol: TCP
+    # PaC
+  - containerPort: 30012
+    hostPort: 8080
+    protocol: TCP
```
yftacherzog: add newline

kind-config.yaml
```diff
@@ -14,6 +14,11 @@ nodes:
   - containerPort: 30010
     hostPort: 8888
     protocol: TCP
+    # UI
   - containerPort: 30011
     hostPort: 9443
     protocol: TCP
+    # PaC
+  - containerPort: 30012
+    hostPort: 8080
```
yftacherzog: I have some reservations about taking port 8080 for every local user that runs the system. Can we consider some alternatives? I'd prefer not to duplicate the configs (i.e. maintaining another configs file). Maybe we can just have it on a less broadly used port?

psturc: > Can we consider some alternatives?

sure thing, the port number doesn't matter, it can be any valid port number

> I'd prefer not to duplicate the configs (i.e. maintaining another configs file).

not sure what do you mean by that? what's your suggestion?

yftacherzog: > not sure what do you mean by that? what's your suggestion?

Was trying to avoid having another config file just for a setup with and without the port. Never mind. If we can use a less commonly-used port then that should be enough.

psturc: thanks, done, see the latest commit

Issue #3095, opened 2025-09-04
move integration-service-related kyverno policy under integration-service


yftacherzog:
The policy here: [dependencies/kyverno/policy/bootstrap-namespace.yaml](https://github.com/konflux-ci/konflux-ci/blob/main/dependencies/kyverno/policy/bootstrap-namespace.yaml) should be define in the same place in which we define integration-service.

Once this is done, we might be able to have it pulled to the Fedora cluster automatically, making defining it explicitly there redundant.

PR #3096, opened 2025-09-04
chore: move integration kyverno policy

yftacherzog:
The policy is related to integration-service so can be created there. This way, it will not have to be created explicitly for downstream consumers (e.g. Fedora cluster).

Closes: #3095


psturc (approved):

dirgim (approved):
LGTM, good idea

Issue #3101, opened 2025-09-04
New init-ns-integration ClusterPolicy doesn't get reconciled until the target namespace is edited


ralphbean:
We noticed this in Fedora when https://github.com/konflux-ci/konflux-ci/pull/3034 and then https://github.com/konflux-ci/konflux-ci/pull/3096 rolled out to the Fedora cluster.

The `init-ns-integration` ClusterPolicy is created, but it doesn't get reconciled to create the `konflux-integration-runner` SA in any tenant namespaces until those namespaces are edited. I had to `oc edit` my namespace to add a random annotation before kyverno would create the SA.

This need to be fixed. If the change is rolled out to a cluster, all tenants break while integration-service expects the SA to exist.

PR #3104, opened 2025-09-04
fix: bump kyverno version, RBAC, policy

psturc:
* bump kyverno to version 1.13.4
* update policy to generate SA in existing namespaces
* update kyverno cluster role to include secrets in the resources list
closes https://github.com/konflux-ci/konflux-ci/issues/3101

### DIFF
--- a/dependencies/kyverno/kustomization.yaml
+++ b/dependencies/kyverno/kustomization.yaml
@@ -2,44 +2,10 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 # renovate: datasource=git-refs depName=https://github.com/kyverno/kyverno versioning=semver
-- https://github.com/kyverno/kyverno/releases/download/v1.11.1/install.yaml
+- https://github.com/kyverno/kyverno/releases/download/v1.13.4/install.yaml
+- kyverno-rbac.yaml
 
 patches:
-- patch: |
-    - op: add
-      path: /rules/-
-      value:
-        apiGroups:
-        - ""
-        resources:
-        - pods
-        - serviceaccounts
-        verbs:
-        - create
-        - update
-        - patch
-        - delete
-        - get
-        - list
-        - watch
-    - op: add
-      path: /rules/-
-      value:
-        apiGroups:
-        - rbac.authorization.k8s.io
-        resources:
-        - rolebindings
-        verbs:
-        - create
-        - update
-        - patch
-        - delete
-        - get
-        - list
-        - watch
-  target:
-    kind: ClusterRole
-    name: kyverno:background-controller:core
 - patch: |-
     - op: replace
       path: /spec/replicas
--- a/konflux-ci/integration/core/bootstrap-namespace.yaml
+++ b/konflux-ci/integration/core/bootstrap-namespace.yaml
@@ -1,4 +1,4 @@
-# Copied from https://github.com/redhat-appstudio/infra-deployments/tree/da73b82a53bf491f178c06d87c1071d712f64105/components/policies/development/integration/bootstrap-namespace
+# Copied from https://github.com/redhat-appstudio/infra-deployments/blob/53c2620f55a9ec321e8ef539f4ce8d71c6273dee/components/policies/production/base/integration/bootstrap-namespace/bootstrap-namespace.yaml
 ---
 apiVersion: kyverno.io/v1
 kind: ClusterPolicy
@@ -9,9 +9,9 @@ metadata:
       This policy creates a service account named konflux-integration-runner and
       binds it to the konflux-integration-runner cluster role in tenant namespaces.
 spec:
-  background: false
   rules:
   - name: create-konflux-integration-runner-serviceaccount
+    skipBackgroundRequests: true
     match:
       any:
       - resources:
@@ -20,16 +20,15 @@ spec:
           selector:
             matchLabels:
               konflux-ci.dev/type: tenant
-    celPreconditions:
-    - name: "on update, oldObject had no konflux-ci.dev/type=tenant label"
-      expression: "request.operation != UPDATE || ! (has(oldObject.metadata.labels) && 'konflux-ci.dev/type' in oldObject.metadata.labels && oldObject.metadata.labels['konflux-ci.dev/type] == 'tenant')"
     generate:
+      generateExisting: true
       synchronize: false
       apiVersion: v1
       kind: ServiceAccount
       name: konflux-integration-runner
       namespace: '{{request.object.metadata.name}}'
   - name: create-konflux-integration-runner-rolebinding
+    skipBackgroundRequests: true
     match:
       any:
       - resources:
@@ -38,10 +37,8 @@ spec:
           selector:
             matchLabels:
               konflux-ci.dev/type: tenant
-    celPreconditions:
-    - name: "on update, oldObject had no konflux-ci.dev/type=tenant label"
-      expression: "request.operation != UPDATE || ! (has(oldObject.metadata.labels) && 'konflux-ci.dev/type' in oldObject.metadata.labels && oldObject.metadata.labels['konflux-ci.dev/type] == 'tenant')"
     generate:
+      generateExisting: true
       synchronize: false
       apiVersion: rbac.authorization.k8s.io/v1
       kind: RoleBinding
--- a/dependencies/kyverno/kyverno-rbac.yaml
+++ b/dependencies/kyverno/kyverno-rbac.yaml
@@ -1,4 +1,5 @@
 # Copied from https://github.com/redhat-appstudio/infra-deployments/tree/da73b82a53bf491f178c06d87c1071d712f64105/components/policies/development/integration/bootstrap-namespace
+# and added rule for managing secrets.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/dependencies/kyverno/policy/kustomization.yaml
+++ b/dependencies/kyverno/policy/kustomization.yaml
@@ -2,5 +2,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - kyverno-rbac.yaml
   - reduce-tekton-pr-taskrun-resource-requests.yaml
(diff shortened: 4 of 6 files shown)


dependencies/kyverno/kustomization.yaml
```diff
@@ -1,44 +1,9 @@
...
-        - create
-        - update
-        - patch
-        - delete
-        - get
-        - list
-        - watch
-  target:
-    kind: ClusterRole
-    name: kyverno:background-controller:core
```
psturc, line 42: I moved the rules to `dependencies/kyverno/policy/kyverno-rbac.yaml`

dependencies/kyverno/policy/kyverno-rbac.yaml
```diff
@@ -10,9 +10,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods
+  - secrets
```
psturc: I had to add secrets to the list of resources due to this error
```
2025-09-04T15:50:51Z ERR github.com/kyverno/kyverno/pkg/background/generate/controller.go:366 > failed to apply generate rule error="rolebindings.rbac.authorization.k8s.io \"konflux-integration-runner-rolebinding\" is forbidden: user \"system:serviceaccount:kyverno:kyverno-background-controller\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:kyverno\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"\"], Resources:[\"secrets\"], Verbs:[\"get\" \"list\"]}" logger=background name=ur-7v2tn policy=init-ns-integration trigger=/test
```

dependencies/kyverno/kustomization.yaml
yftacherzog: My idea when creating the kyverno/policy dur was to have policies there. Ideally, we should try putting other kyverno stuff someplace else under the kyverno dir.

konflux-ci/integration/core/bootstrap-namespace.yaml
```diff
@@ -20,16 +20,15 @@ spec:
           selector:
             matchLabels:
               konflux-ci.dev/type: tenant
-    celPreconditions:
```
yftacherzog, line 23: @filariow mentioned on slack that the idea behind this was to only trigger the policy when the ns first gets the tenant label. 

With that in mind, I think the condition could be the same as before but replacing the !update with create. In other words, either when the label first applied or when the ns is created.

konflux-ci/integration/core/bootstrap-namespace.yaml
```diff
@@ -9,9 +9,9 @@ metadata:
       This policy creates a service account named konflux-integration-runner and
       binds it to the konflux-integration-runner cluster role in tenant namespaces.
 spec:
-  background: false
   rules:
   - name: create-konflux-integration-runner-serviceaccount
+    skipBackgroundRequests: true
```
yftacherzog, line 14: What does this mean?

konflux-ci/integration/core/bootstrap-namespace.yaml
```diff
@@ -20,16 +20,15 @@ spec:
           selector:
             matchLabels:
               konflux-ci.dev/type: tenant
-    celPreconditions:
-    - name: "on update, oldObject had no konflux-ci.dev/type=tenant label"
-      expression: "request.operation != UPDATE || ! (has(oldObject.metadata.labels) && 'konflux-ci.dev/type' in oldObject.metadata.labels && oldObject.metadata.labels['konflux-ci.dev/type] == 'tenant')"
     generate:
+      generateExisting: true
```
yftacherzog, line 24: What does this mean?

psturc: https://release-1-12-0.kyverno.io/docs/writing-policies/generate/#generate-existing-examples

dependencies/kyverno/kustomization.yaml
psturc: ok so adding it to `dependencies/kyverno/kyverno-rbac.yaml` sound ok?

yftacherzog: Sounds good

konflux-ci/integration/core/bootstrap-namespace.yaml
psturc: @amisstea recommended to use the policy from the production env of infra-deployments, so I replaced the existing policy with the production one.

but when looking at the [docs](https://main.kyverno.io/docs/policy-types/cluster-policy/policy-settings/), `skipBackgroundRequests`'s default value is `true`, so it probably doesn't even have to be here

konflux-ci/integration/core/bootstrap-namespace.yaml
psturc: I removed it because [the production version of the policy](https://github.com/redhat-appstudio/infra-deployments/blob/21bf7fb57dd14d632b29dc0e54176f2d3cef2e7f/components/policies/production/base/integration/bootstrap-namespace/bootstrap-namespace.yaml) was recommended to use, which does not contain these preconditions

konflux-ci/integration/core/bootstrap-namespace.yaml
psturc: it basically replaces what I had in the previous commit for an older version of kyverno, which was `generateExistingOnPolicyUpdate: true`

konflux-ci/integration/core/bootstrap-namespace.yaml
filariow: that's related to ArgoCD: https://kyverno.io/docs/installation/platform-notes/#notes-for-argocd-users

If not present, Kubernetes sets this field by default and ArgoCD loops. On the short term we started adding the field to the policies, long term we're applying [Kyverno's Configuration Best Practices](https://github.com/redhat-appstudio/infra-deployments/pull/7348)

dependencies/kyverno/policy/kyverno-rbac.yaml
filariow: to make this more robust, should we add a ClusterRoleBinding in `konflux-ci/integration/core` and deploy it together with the ClusterPolicy to bind the `system:serviceaccount:kyverno:kyverno-background-controller` and the ClusterRole `konflux-integration-runner` ? This way if the ClusterRole is updated, we won't face permission problems.

psturc: yep, I agree. I suggested in the chat to "add `rbac.kyverno.io/aggregate-to-background-controller: "true"` label to the [konflux-integration-runner clusterrole](https://github.com/redhat-appstudio/infra-deployments/blob/be393d878e33b77da36c48907cad0906fed5cb73/components/konflux-rbac/base/konflux-integration-runner.yaml#L4)" which I think would solve that too... not sure if I'm missing something? 🤔

dependencies/kyverno/kustomization.yaml
```diff
@@ -1,44 +1,10 @@
...
 
 patches:
-- patch: |
-    - op: add
-      path: /rules/-
-      value:
-        apiGroups:
-        - ""
-        resources:
-        - pods
```
yftacherzog, line 15: Do we have a role with pods somewhere now?

dependencies/kyverno/kustomization.yaml
yftacherzog: It seems like the rules in `kyverno-rbac.yaml` have fewer verbs than here. Is that ok?

yftacherzog (changes requested):
Can you squash the commits?

dependencies/kyverno/policy/kyverno-rbac.yaml
psturc: in the end we decided to go with a clusterrolebinding

dependencies/kyverno/kustomization.yaml
psturc: yes, this is taken from [the production env from infra-deployments](https://github.com/redhat-appstudio/infra-deployments/blob/d2271913a895f5b7e1fb15cf871a80383c84d1ac/components/policies/production/base/integration/bootstrap-namespace/kyverno-rbac.yaml)

more verbs should not be needed

dependencies/kyverno/kustomization.yaml
psturc: we don't and as I learned recently Kyverno does not need that for standard mutation policies

yftacherzog: The pods were needed for the policy that reduced the resources from the tekton tasks.

psturc: that's required only for modifying **existing** resources. but since we're using that policy only for CREATE operations, it is not needed.

I validated it on a fresh cluster with just kyverno installed and `reduce-tekton-pr-taskrun-resource-requests` applied:

```
➜  konflux-ci git:(fix-generate-sa-in-existing-ns) cat <<EOF | oc create -f -
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  labels:
    pipelines.appstudio.openshift.io/type: build
spec:
  containers:
  - name: test
    image: busybox
    resources:
      requests:
        memory: "500Mi"
        cpu: "500m"
    command: ["sh", "-c", "echo 'hello world'"]
EOF
pod/test-pod created
➜  konflux-ci git:(fix-generate-sa-in-existing-ns) oc get pod test-pod -o yaml | grep -A2 requests
      requests:
        cpu: 1m
        memory: 1Mi
```

psturc:
@yftacherzog the sanity check is failing on
```
cannot retrieve the oci image: GET https://quay.io/v2/***/konflux-vanguard/task-rpms-signature-scan/manifests/sha256:49ff6d117c3e9dc3966d1244e118e168b3501742ec14c3a4161a276ff48d04d5: MANIFEST_UNKNOWN: manifest unknown; map[]"***]
```

any idea what's causing this?

dependencies/kyverno/kustomization.yaml
yftacherzog: I see. I wonder why I added it in the first place. Might be that I tried a different approach at some point and the role stayed.

PR #3152, opened 2025-09-09
fix: renovate configs for Conforma

yftacherzog:
The previous configs caused the comment lines that were added for renovate to be duplicated. It seems as we don't need those comments in this case. So this change is simplifying the implementation and dropping the comments.


avi-biton (approved):
lgtm

PR #3226, opened 2025-09-23
chore(deps): update enterprise contract CRD source to conforma/crds

robnester-rh:
Switch from enterprise-contract-controller to conforma/crds repository for CRD definitions with updated git reference

Ref: EC-1422

### DIFF
--- a/konflux-ci/enterprise-contract/core/kustomization.yaml
+++ b/konflux-ci/enterprise-contract/core/kustomization.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - https://github.com/enterprise-contract/enterprise-contract-controller/config/crd?ref=6336da883e561bfa307fef42bfd6ebb23541a41d
+  - https://github.com/conforma/crds/config/crd?ref=526b80826ab9b2a2d5d28086aa4165d4f3218c1f
   - ns.yaml
   - public-ecp-rbac.yaml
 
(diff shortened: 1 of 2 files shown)


yftacherzog (approved):

konflux-ci/enterprise-contract/core/kustomization.yaml
yftacherzog, line 1: Any changes at this point to the `configMapGenerator` at the bottom of this file?

robnester-rh: This is currently automatically redirected. I can update it explicitly in a subsequent PR so that we're not muddying the water.

robnester-rh:
> I think we're also going to need another line here for the new location: https://github.com/konflux-ci/konflux-ci/blob/main/renovate.json#L7

I've added this in a subsequent commit.

renovate.json
```diff
@@ -4,7 +4,7 @@
     {
       "matchPackageNames": [
         "ghcr.io/chmouel/gosmee",
-        "https://github.com/enterprise-contract/*",
+        "https://github.com/conforma/*",
```
yftacherzog, line 8: I get the feeling that we still have stuff under `https://github.com/enterprise-contract/*`. Maybe add another line rather than replace it.

robnester-rh: Ack. I've updated and pushed updated commit.

yftacherzog: I don't see this.

yftacherzog:
@robnester-rh I was out on public holidays. Back today. I see conflicts. Let's resolve them and then I'll try to figure out what's failing it.

yftacherzog:
@robnester-rh I'd prefer not to merge it without CI passing, but I agree it does not seem to be related to your changes.

Is this blocking you from proceeding to something else?

PR #3231, opened 2025-09-24
Konflux readme update

cdeyoung:
This update fixes a lot of outdated steps, adds clarity to some of the complex steps that were vague, and fixes grammar and spelling errors.


README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+**Note:** You can configure your server to operate the same way using either the
+Konflux UI method or the terminal and Kubernetes manifests procedure. For example,
+you can also configure Konflux to support an image registry using the terminal process.
 
-In both cases, the recommended way to try out Konflux is using
-[Kind](https://kind.sigs.k8s.io/).
-The process below creates a Kind cluster using the provided config in this repository.
-The config tells Kind to forward port `9443` from the host to the Kind cluster. The port
-forwarding is needed for accessing Konflux.
+In both cases, the recommended way to try out Konflux is using [Kind](https://kind.sigs.k8s.io).
```
yftacherzog, line 67: We're keeping line length at 89 characters. That's the reason for the end of this line to be pushed to the next line.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+This guide will walk you through creating a Kind cluster using the `kind-config.yaml`
+manifest, located in the root directory of this repository. This manifest tells
+[Kind](https://kind.sigs.k8s.io) to forward port `9443` from the host to the
+[Kind](https://kind.sigs.k8s.io) cluster. Kind must configure this port forwarding
```
yftacherzog, line 71: I think including this link only one time in this paragraph should be enough.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+to access Konflux.
 
-**Note:** If using a remote machine for setup, you'd need to port-forward port `9443` on
-the remote machine to port `9443` on your local machine to be able to access the UI from
-your local machine.
+**Note**: For security reasons, Konflux limits browser access to the local server.
```
yftacherzog, line 74: > Konflux limits browser access to the local server.

I don't think this statement is true. I'm not sure it's related to in this context either.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+If you are using a remote machine to set up Konflux, you'll need to use SSH to set
+up port forwarding to access Konflux through your local computer's browser. To do this,
+open an additional terminal and type the following command:
+
+```bash
+ssh -L 9443:localhost:9443 <remote_host_user_name>@<remote_host_ip_address>
```
yftacherzog, line 80: I believe this won't work until Konflux is up an running. For that reason, it was mentioned later in the document.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+```
 
 ## Machine Minimum Requirements
 
-The deployment is currently only supported on **x86_64 Linux** platforms.
+Konflux currently only supports the **x86_64 Linux** platform.
```
yftacherzog, line 85: I'm not sure if this statement is true for any deployment of Konflux. The way Konflux is deployed in this context was only tested on x86_64 Linux platforms.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
 The deployment requires the following **free** resources:
 
-**CPU**: 4 cores\
-**RAM**: 8 GB
+- **CPU**: 4 cores
+- **RAM**: 8 GB
 
-**Note:** Additional load from running multiple pipelines in parallel will require
-additional resources.
+**Note**: You will need to add more resources if you plan to run multiple pipelines in parallel.
```
yftacherzog, line 92: Line too long
```suggestion
**Note**: You will need to add more resources if you plan to run multiple
pipelines in parallel.
```

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
 
-* [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) (`v0.26.0` or
-  newer) along with `podman` (`v5.3.1` or newer) or `docker` (`v27.0.1` or newer)
-* `kubectl` (`v1.31.1` or newer)
-* `git` (`v2.46` or newer)
-* `openssl` (`v3.2.2` or newer)
+- **git** (v2.46 or newer)
+- **golang** (v1.24.6 or newer)
+- **[Kind](https://kind.sigs.k8s.io)** (v0.26.0 or newer) along with **podman** (v5.3.1 or newer)
+or **docker** (v27.0.1 or newer)
```
```suggestion
- **[Kind](https://kind.sigs.k8s.io)** (v0.26.0 or newer) along with **podman** (v5.3.1
or newer) or **docker** (v27.0.1 or newer)
```

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+- **kubectl** (v1.31.1 or newer)
+- **make** (v4.4.1 or newer)
+- **openssl** (v3.2.2 or newer)
 
 ## Bootstrapping the Cluster
-:gear: Clone this repository:
 
- ```bash
+:gear: Before you can bootstrap and configure your cluster, you will need to clone the Konflux
```
yftacherzog, line 108:
```suggestion
:gear: Before you can bootstrap and configure your cluster, you will need to clone the
Konflux
```

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+```bash
 git clone https://github.com/konflux-ci/konflux-ci.git
 cd konflux-ci
 ```
 
-**Note:** It is recommended that you increase the `inotify` resource limits in order to
-avoid issues related to
-[too many open files](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files). To increase the limits temporarily, run the
-following commands:
+**Note**: We recommend you increase the `inotify` resource limits to avoid issues with having
```
yftacherzog, line 120:
```suggestion
**Note**: We recommend you increase the `inotify` resource limits to avoid issues with
having
```

README.md
yftacherzog: I commented the same for some other lines, but there are more.

README.md
```diff
@@ -48,100 +48,127 @@ the reader to fulfill an action.
...
+    hostPort: 8888
+    protocol: TCP
+    # UI
+  - containerPort: 30011
+    hostPort: 9443
+    protocol: TCP
+    # PaC
+  - containerPort: 30012
+    hostPort: 8180
+    protocol: TCP
```
yftacherzog: Adding those fields here is introducing a potential maintenance burden, as while they hold no meaning in this context, if they're changed in their original context, we will also have to know that they should be updated here (which we probably won't remember).

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
-5. :gear: If Konflux was installed on a cluster hosted in a remote machine, SSH port-forwarding can 
-be used to access. Open an additional terminal and run the following command
-(make sure to add the details of your remote machine and user):
+5. :gear: As a reminder, if you've installed Konflux on a remote host, you will need to configure SSH port
+forwarding to access your cluster. To do this, open an additional terminal and run the following command
+(make sure to add the details for your remote user and IP address):
 
 ```bash
-ssh -L 9443:localhost:9443 $USER@$VM_IP
+ssh -L 9443:localhost:9443 <remote_host_user_name>@<remote_host_ip_address>
```
yftacherzog: I think the changes here are fine. I don't think we should include the command at the top of the document, as it may cause confusion there and also become a maintenance burden.

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
 ```
 
-6. The UI will be available at https://localhost:9443. You can login using a test user.
+6. Once you've set up port forwarding, your cluster's UI will be available at https://localhost:9443.
```
yftacherzog, line 300: The common user (deploying Konflux locally), will not be setting up port forwarding, so this statement might be confusing for most users.

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
+You can log in using the test user's account.
 
-`username:` `user2@konflux.dev`
+**username**: **user@konflux.dev**
```
```suggestion
**username**: **user2@konflux.dev**
```

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
-We now have Konflux up and running. Next, we shall configure Konflux to respond
-to Pull Request webhooks, build a user application and push it to a registry.
+You now have Konflux up and running. Next, we'll configure Konflux to respond to pull request webhooks,
+build a user application, and push it to a registry.
 
 ## Enable Pipelines Triggering via Webhooks
 
-Pipelines Can be triggered by Pull Request activities, and their outcomes will be
-reported back to the PR page in GitHub.
+You can configure Konflux to trigger pipelines anytime someone submits a pull request.
```
yftacherzog, line 311: This suggests that this sort of configuration is optional, while in fact it's not. It might be that the original wording was not perfect either in that sense...

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
 
-A GitHub app is required for creating webhooks that Tekton will listen on. When deployed
-in a local environment like Kind, GitHub will not be able to reach a service within the
-cluster. For that reason, we need to use a proxy that will listen on such events
-from within the cluster and will relay those events internally.
+You can also configure Konflux to send the outcomes of these triggers back to your
```
yftacherzog, line 313: Here as well, we need to make sure that this does not come out as an optional step.

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
-The second option demonstrates using Kubernetes manifests to onboard, and releases
-the builds to a container registry deployed to the cluster. The idea behind this
-scenario is to simplify onboarding in order to demonstrate Konflux with greater ease.
+    1. Go to your [Applications page on GitHub](https://github.com/settings/apps).
 
-Both options will use an example repository containing a Dockerfile to be built by
-Konflux:
+    2. Click on your GitHub App (if you haven't done so yet, follow the instructions on the
+    [Create a Pipelines-as-Code GitHub App](https://pipelinesascode.com/docs/install/github_apps/#manual-setup)
+    page.
```
yftacherzog, line 385:
```suggestion
    page).
```

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
-triggered by push events (e.g. when PRs are merged).
-
-Your application is now onboarded, and you can continue to the
+
+1. Manifests in GitHub for the application pipelines it will run.
+
+2. The [Quay.io](https://quay.io) repositories into which it will push container images.
+
+You can configure the first item by creating
+[GitHub Application Secrets](https://github.com/konflux-ci/konflux-ci/blob/main/docs/github-secrets.md)
```
yftacherzog, line 409: Is there a benefit of having the complete URL here comparing to having only the path inside the repository? I think that having the full path is reducing the portability of this document (i.e. if the entire repo is moved).

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
+for the `pipelines-as-code`, `build-service`, and `integration-service` namespaces,
+and then installing your newly created GitHub App on your repository, as explained above.
+
+To configure these three namespaces, run the following commands from the terminal of your Konflux server:
+
+1. Configure the `pipelines-as-code` namespace.
+    ```
+    kubectl -n pipelines-as-code create secret generic pipelines-as-code-secret \
```
yftacherzog, line 417: People will try to copy-paste those commands to the terminal, which won't work as they are missing the values for the literals. I also don't like it very much that we repeat PAC documentation, as in case those change, we will also need to change our docs.

If we do decide that we want to make it something that the users can copy-paste, then we can parameterize those lines and add a section in which we set the parameters, but this needs to be tested before we put it in this doc.

README.md
yftacherzog: Also, we had this content in the main readme file and decided to move it to another document in order to keep this document cleaner. I'm not sure we should reverse this.

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
+            --from-literal github-application-id=<your_app_id> \
+            --from-literal webhook.secret="<your_webhook_secret>"
+    ```
+
+To achieve the second item, create an organization and an application
+in [Quay.io](https://quay.io) that will allow Konflux to create
+repositories for your applications. To do that, follow the steps below.
+
+**Note**: You can also reference the
+[Quay.io Configurations](https://github.com/konflux-ci/konflux-ci/blob/main/docs/quay.md#automatically-provision-quay-repositories-for-container-images)
```
yftacherzog, line 444: This procedure is very long, and that's the reason it was moved to another document. I don't think we should move it back here, but in any case, I don't think we should repeat this same lengthy documentation in multiple places as it will be a maintenance burden.

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
+
+    **Note**: If you make a mistake and need to redo this command, you must first
+    delete the `quay-robot-secret` namespace before you can recreate it. You can
+    do this with the following command:
+
+    ```
+    kubectl delete secret quay-robot-secret -n image-controller-system
+    ```
+
+3. :gear: Download, configure, and deploy the `image-controller` project.
```
yftacherzog, line 525: You don't need to clone image-controller manually to have it deployed. What's the reason for changing the procedure comparing to what we have in the designated document?

README.md
```diff
@@ -162,225 +189,414 @@ and modify the `system-reserved` line under `kubeletExtraArgs`:
...
+the Application and Component resources, and we manually create the pull
+request to configure and run the pipelines using Konflux.
 
-To do that:
+Let's step through the process of deploying a manifest via the terminal.
 
-1. :gear: Use a text editor to edit your local copy of the
-   [example application manifests](./test/resources/demo-users/user/ns2/application-and-component.yaml):
+1. :gear: Use your text editor of choice to edit your local copy of the
+[example application manifests](https://github.com/konflux-ci/konflux-ci/blob/main/test/resources/demo-users/user/ns2/application-and-component.yaml).
```
yftacherzog: I think that in this case, contrary to the current and suggested text, it will make sense to have the path to the file visible directly, as this is what the user will need in order to locate the file on their local repo.

```suggestion
1. :gear: Use your text editor of choice to edit your local copy of the
example application manifests at
`test/resources/demo-users/user/ns2/application-and-component.yaml`.
```

README.md
cdeyoung: Thanks! Yes, I'm planning on using a linter to fix the lengths before I submit this for approval.

README.md
cdeyoung: My understanding is that this is the default setting, and you'll receive a security warning if you attempt to enable remote access. Is this not the case?

README.md
cdeyoung: You can set up port forwarding at any time, and it will work once Konflux is set up. However, that's not why I mentioned it here. In the original document, this section talked about the need to port-forward port 9443 for Kind and also to access Konflux from a remote browser.

I added this paragraph to add additional clarity. I will remove the command and the sentence before it, though.

README.md
cdeyoung: You're right. I only heard of Konflux three weeks ago, so I merely adjusted the existing sentence for clarity. I'll add that it also supports arm64.

README.md
cdeyoung: Thanks! I'll adjust all the line lengths before converting this to a pull request.

README.md
yftacherzog: The UI is a service inside the cluster. For the service to be accessible, the Kubernetes cluster needs to somehow allow access to the service. We do it using port-forwarding (we make calls to a local port reach a service inside the cluster).

When the cluster runs on a remote machine, this is slightly different and requires port forwarding on the SSH level as well.

In any case, this is not a Konflux security feature, but the way Kubernetes networking works.

README.md
yftacherzog: We only test upstream konflux on x86_64 Linux, so I'm not sure it's safe to mention ARM here, as it will make users expect it to work, while we don't have any tests done on that platform, and we don't know what it takes to get it to run there.

PR #3331, opened 2025-10-14
add Maintainer and Contributor personas

filariow:
Requires:
- [x] https://github.com/konflux-ci/konflux-ci/pull/3416

This change introduces the ClusterRoles for Maintainer and Contributor.
It also removes permissions on non existing resources.

### DIFF
--- a/konflux-ci/rbac/core/konflux-contributor-user-actions-core.yaml
+++ b/konflux-ci/rbac/core/konflux-contributor-user-actions-core.yaml
@@ -1,0 +1,101 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: konflux-contributor-user-actions-core
+  labels:
+    rbac.konflux-ci.dev/aggregate-to-contributor: "true"
+rules:
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - applications
+      - components
+      - imagerepositories
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - snapshots
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - tekton.dev
+    resources:
+      - pipelineruns
+      - taskruns
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - results.tekton.dev
+    resources:
+      - results
+      - records
+      - logs
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - integrationtestscenarios
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - enterprisecontractpolicies
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - releases
+      - releaseplans
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - releaseplanadmissions
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - ''
+    resources:
+      - configmaps
+      - pods
+      - pods/log
+  - verbs:
+      - get
+    apiGroups:
+      - ''
+    resources:
+      - namespaces
+  - verbs:
+      - get
+      - list
+      - watch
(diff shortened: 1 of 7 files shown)


konflux-ci/rbac/core/konflux-admin-user-actions.yaml
```diff
@@ -190,7 +130,6 @@ rules:
   - verbs:
       - get
       - list
-      - watch
```
yftacherzog: What's the reason for dropping `watch`?

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
```diff
@@ -206,60 +145,10 @@ rules:
     resources:
       - serviceaccounts/token
   - verbs:
-      - create
-    apiGroups:
-      - ''
-    resources:
-      - pods/exec
```
yftacherzog: What's the reason for dropping this?

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
```diff
@@ -206,60 +145,10 @@ rules:
...
-      - create
-      - update
-      - patch
-      - delete
-    apiGroups:
-      - projctl.konflux.dev
-    resources:
-      - projects
-      - projectdevelopmentstreams
-      - projectdevelopmentstreamtemplates
```
yftacherzog: What's the reason for removing those (also in other roles)?

filariow: they are not installed as part konflux-ci

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
filariow: we are not allowing this downstream. not sure if we need it here. cc: @gbenhaim

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
filariow: we don't allow this downstream. However it makes sense to me to have it. I'd propose to revert this change here and bring it downstream too

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
yftacherzog: I'm not sure why they were added here originally, but I think they are used in the Fedora cluster, so this might affect users.

filariow: should we patch the roles in the fedora cluster instead of adding these permissions here?

yftacherzog: Makes sense, but we need some sort of plan just not to take users by surprise (we can also merge without this part and figure it out later).

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
gbenhaim: Letting users exec into pods is a security violation since, for example, files related to the bulld can be changed at runtime.

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
gbenhaim: This is a valid point. @filariow I think thar we should add permissions for managing those resources to the downstream deployment.

gbenhaim: Or maybe we should start using role aggregation that will allow add-ons to add their permission to the default cluster roles konflux provides.

filariow: nice, I was looking into this. I filed a PR for introducing Aggregated ClusterRoles for the `rbac` component: https://github.com/konflux-ci/konflux-ci/pull/3389

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
filariow: https://github.com/redhat-appstudio/infra-deployments/pull/8740 and 089a1f2b9a43e8ffbed82e52c5318b9856fce8f4

yftacherzog:
@filariow I think we're good. Squash the commits and we can merge it.

konflux-ci/rbac/core/konflux-admin-user-actions.yaml
yftacherzog: Should https://github.com/konflux-ci/konflux-ci/pull/3389 be merged first?

filariow: Yes, let me move this one in draft until the other is merged

gbenhaim (approved):

PR #3416, opened 2025-10-22
cleanup RBAC permissions

filariow:
This change removes permissions on outdated resources and deprecates permissions on resources not installed as part of `konflux-ci`.

### DIFF
--- a/konflux-ci/rbac/core/konflux-admin-user-actions-core.yaml
+++ b/konflux-ci/rbac/core/konflux-admin-user-actions-core.yaml
@@ -20,7 +20,6 @@ rules:
     resources:
       - applications
       - components
-      - componentdetectionqueries
       - imagerepositories
   - verbs:
       - get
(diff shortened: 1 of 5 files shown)


konflux-ci/rbac/core/konflux-admin-user-actions-core.yaml
```diff
@@ -33,34 +32,15 @@ rules:
...
       - watch
     apiGroups:
-      - managed-gitops.redhat.com
+      - tekton.dev
     resources:
-      - gitopsdeployments
-      - gitopsdeploymentmanagedenvironments
-      - gitopsdeploymentrepositorycredentials
-      - gitopsdeploymentsyncruns
+      - taskruns
```
hugares, line 43: this change is granting the right to create taskruns, it used to be get/list only before. Is this intentional?

filariow: this change is actually [removing create/edit/update](
https://github.com/konflux-ci/konflux-ci/blob/ee41445e382a2017590ac23b8d4ccb038a5aebf7/konflux-ci/rbac/core/konflux-admin-user-actions-core.yaml#L36C1-L55C21) on `taskruns`. That's something we are not allowing [upstream](https://github.com/redhat-appstudio/infra-deployments/blob/ac780cb26e1f45585d2382327b99c7d3ec0b652e/components/konflux-rbac/production/base/konflux-admin-user-actions.yaml#L36-L43).

hugares: oh, sorry about that, I checked the permission block from after, got lost in the indentation...

filariow: @yftacherzog @gbenhaim are we fine with this or it could be a problem for the fedora cluster?

yftacherzog: I would not assume anyone is directly using this permission there. @ralphbean, is that correct?

ralphbean: Yeah, I agree. No user-admin should be directly creating taskruns

filariow: thanks all for the feedback. I'll close this thread and merge this PR

PR #3495, opened 2025-10-29
chore: ignore aggregated rules in clusterroles

yftacherzog:
This should prevent sync issues in argoCD downstream.


avi-biton (approved):
/lgtm

yftacherzog:
> could it be related to https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/#ignoring-rbac-changes-made-by-aggregateroles?

Looks like it. I might want to have that done as well, but will be harder for me to test.

PR #3510, opened 2025-10-30
chore: allow user admin to manage jobs and cronjobs

yftacherzog:
This allows users to trigger periodic pipelines.

Assisted-by: Cursor


yftacherzog:
This PR ports upstream a change already merged to the Fedora cluster.

The role name is different to avoid conflicts downstream until the role there is deleted.
Other than the names, the roles upstream and downstream are identical.

PR #3515, opened 2025-10-30
chore(KONFLUX-9778): Add version checks to deploy-deps.sh

### LINKED ISSUE #2936: Add versions checks to the deploy-deps.sh script
In the `deploy-deps.sh' script, we are checking for the requirements of Konflux-ci before we deploy it.
The checks are done in the check_req function in the script.
Currently we only check that the needed dependencies are installed, however we do not check the versions of the dependencies (for example, we check that `kubectl` exists, but not that it has the needed version).

We need to add minimum versions required check to our dependencies.

avi-biton:
This PR adds version checks to the deploy-deps.sh script.
It checks if the kubectl and openssl commands are installed and
that kubectl has server-side apply support.

Close #2936 

Assisted-by: Cursor

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,9 @@ additional resources.
 
 * [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) (`v0.26.0` or
   newer) along with `podman` (`v5.3.1` or newer) or `docker` (`v27.0.1` or newer)
-* `kubectl` (`v1.31.1` or newer)
+* `kubectl` (`v1.31.1` or newer -- server-side apply support is required)
 * `git` (`v2.46` or newer)
-* `openssl` (`v3.2.2` or newer)
+* `openssl` (`v3.0.13` or newer)
 
 ## Bootstrapping the Cluster
 :gear: Clone this repository:
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -21,11 +21,25 @@ check_req(){
     # check if requirements are installed
     for i in "${requirements[@]}"; do
         if ! command -v "$i" &> /dev/null; then
-                uninstalled_requirements+=("$i")
+                if [ "$i" == "kubectl" ]; then
+                    uninstalled_requirements+=("kubectl (server-side support required - v1.31.1 or newer)")
+                else
+                    uninstalled_requirements+=("$i")
+                fi
         else
                 echo -e "$i is installed"
         fi
     done
+
+    # check if kubectl has server-side support
+    if command -v kubectl &> /dev/null; then
+        echo -e "Checking if kubectl has server-side support"
+        if ! kubectl apply --help | grep -q -- --server-side; then
+            uninstalled_requirements+=("kubectl (server-side support required - v1.31.1 or newer)")
+        else
+            echo -e "kubectl supports server-side apply"
+        fi
+    fi
 
     if (( ${#uninstalled_requirements[@]} == 0 )); then
         echo -e "\nAll requirements are met\nContinue"


deploy-deps.sh
```diff
@@ -14,26 +14,63 @@ main() {
...
 check_req(){
-    # declare the requirements
-    local requirements=(kubectl openssl)
+    # Helper function to compare versions
+    # Returns 0 if installed_version >= min_required_version, 1 otherwise
+    version_ge() {
+        local installed_version="$1"
+        local min_required_version="$2"
+        # Use sort -V (version sort) to compare versions
+        printf '%s\n%s\n' "$min_required_version" "$installed_version" | sort -V -C
```
yftacherzog: I'm not sure how robust this is. It depends on the local implementation of `sort` the user has, and it might not support those flags (AI says sort -V is a non-portable GNU extension and will fail on systems like macOS). It also depends on the structure of the version strings, which I'm not sure we want to depend on.

I wonder if we, at the very least, would want to provide a mechanism to skip this check based on some input.

avi-biton: What other way do you suggest to get the versions?
Each dependency has its own way to get the version, these are the documented ways to get the version for kubectl and openssl.

avi-biton: We can use other method than `sort -V`, maybe a strict bash flow

yftacherzog: > What other way do you suggest to get the versions?

It's more about parsing than getting. Suppose that there's some sort of special build the user is using which is new enough but contains a weird suffix.

> We can use other method than sort -V, maybe a strict bash flow

Need to make sure this does become a maintenance burden. AI suggests `awk`, but personally I think this becomes "write-only code" very easily.

Maybe we can print the failure (either for wrong version or command failures) and prompt the user to either fix the installed version or run with a flag that will skip the check.

avi-biton: BTW, the CI is failing because the `ubuntu-latest` which we are using does not contain the minimum required openssl version.
In the requirements it is mentioned that the openSSL version should be v3.2.2 or higher, while we have v3.0.13.
So we probably need either to lower the requirement (if possible) or to somehow install openssl with the appropriate version.

yftacherzog: We probably just documented the version that the one to document had on their system. We can lower it.

avi-biton:
I don't mind updating it to check if OpenSSL exist and if kubectl exists and has server-side support

yftacherzog:
> I don't mind updating it to check if OpenSSL exist and if kubectl exists and has server-side support

I think we should do that. It will address the original user issue and will be much cleaner comparing to the alternatives.

README.md
```diff
@@ -92,9 +92,9 @@ additional resources.
 
 * [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) (`v0.26.0` or
   newer) along with `podman` (`v5.3.1` or newer) or `docker` (`v27.0.1` or newer)
-* `kubectl` (`v1.31.1` or newer)
+* `kubectl` with server-side apply support
```
```suggestion
* `kubectl` (`v1.31.1` or newer -- server-side apply support is required)
```
I'd still keep the version number, otherwise, users will have to go to the release notes to figure out what's the minimal version.

deploy-deps.sh
```diff
@@ -27,6 +27,16 @@ check_req(){
         fi
     done
 
+    # check if kubectl has server-side support
+    if command -v kubectl &> /dev/null; then
+        echo -e "Checking if kubectl has server-side support"
+        if ! kubectl apply --help | grep -q -- --server-side; then
+            echo -e "\nkubectl does not have server-side support"
+            echo -e "Please install a kubectl version that supports server-side apply"
```
```suggestion
            echo -e "Please install a kubectl version that supports server-side apply (v1.31.1 or newer)"
```

deploy-deps.sh
```diff
@@ -27,6 +27,16 @@ check_req(){
         fi
     done
 
+    # check if kubectl has server-side support
+    if command -v kubectl &> /dev/null; then
```
yftacherzog, line 35: What error message will you get if you don't have kubectl installed?
I think it will make sense to add this block below the one that check if the command exists.

avi-biton: The outputs we will get are:

No openSSL:
```
kubectl is installed
Checking if kubectl has server-side support
kubectl supports server-side apply

Some requirements are missing, please install the following requirements first:
openssl
```
No kubectl:
```
openssl is installed

Some requirements are missing, please install the following requirements first:
kubectl (server-side support required - v1.31.1 or newer)
```
kubectl with no server-side support:
```
kubectl is installed
openssl is installed
Checking if kubectl has server-side support

Some requirements are missing, please install the following requirements first:
kubectl (server-side support required - v1.31.1 or newer)
```
All requirements met:
```
kubectl is installed
openssl is installed
Checking if kubectl has server-side support
kubectl supports server-side apply

All requirements are met
Continue
```

yftacherzog: If you get to the `kubectl apply --help | grep -q -- --server-side` line and kubectl is missing, will the script exit immediately or get to print the proper error message?

avi-biton: You will not hit this line in case kubectl is missing.
It first checks that kubectl is installed:
`if command -v kubectl &> /dev/null;`
And only if it is, it will check for server-side support.

In this case, we already have the kubectl in the uninstalled_requirements with the proper message, so the output will be:
```
openssl is installed

Some requirements are missing, please install the following requirements first:
kubectl (server-side support required - v1.31.1 or newer)
```

We can of course use other method than the `uninstalled_requirements` array, we can simply run 2 checks, one for `openssl` and one for `kubectl`, but I wanted it to be robust and to fit other requirements that we might have in the future.

yftacherzog: ok. Got it now. I think I'm good with the current approach.

yftacherzog (approved):

PR #3540, opened 2025-11-02
fix(KONFLUX-9697): konflux-info namespace is missing

avi-biton:
Adding Konflux Info namespace and resources.

- Define only rbac rules in the `info.json` file
- Not defining `sbom-server` and `github` (for github application)
- Add  `banner-content.yaml` with proper message

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -69,6 +69,8 @@ deploy() {
     deploy_smee
     echo "🛡️  Deploying Kyverno..." >&2
     deploy_kyverno
+    echo "📋 Deploying Konflux Info..." >&2
+    deploy_konflux_info
 }
 
 test_pvc_binding(){
@@ -167,6 +169,10 @@ deploy_kyverno() {
     # Wait for policy CRD to be installed. Don't need to wait for everything to be up
     sleep 5
     kubectl apply -k "${script_path}/dependencies/kyverno/policy"
+}
+
+deploy_konflux_info() {
+    kubectl apply -k "${script_path}/dependencies/konflux-info"
 }
 
 retry() {
--- a/dependencies/konflux-info/banner-content.yaml
+++ b/dependencies/konflux-info/banner-content.yaml
@@ -1,0 +1,3 @@
+# Only the first banner will be displayed. Put the one to display at the top and remove any which are no longer relevant
+- summary: "Welcome to Konflux-CI! This is a development environment for testing and development purposes."
+  type: "info"
--- a/dependencies/konflux-info/base/kustomization.yaml
+++ b/dependencies/konflux-info/base/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - namespace.yaml
+  - rbac.yaml
+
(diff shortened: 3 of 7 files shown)


konflux-ci/konflux-info/banner-content.yaml
```diff
@@ -0,0 +1,2 @@
+# Only the first banner will be displayed. Put the one to display at the top and remove any which are no longer relevant
+[]
```
yftacherzog: Can we put some sort of placeholder here (e.g. Welcome to Konflux-CI)?

konflux-ci/konflux-info/banner-schema.json
```diff
@@ -0,0 +1,84 @@
+{
```
yftacherzog: Is this schema defined in a manifest on another repo? Can we take it from there instead of maintaining it here?

avi-biton:
@yftacherzog I removed the banners completely, as this was not the issue.
I updated it to only create the needed rbac.
We can add the banners later if we want to

konflux-ci/konflux-info/banner-schema.json
yftacherzog: (I mean upstream repo - not infra-deployments)

deploy-deps.sh
```diff
@@ -169,6 +171,10 @@ deploy_kyverno() {
     kubectl apply -k "${script_path}/dependencies/kyverno/policy"
 }
 
+deploy_konflux_info() {
+    kubectl apply -k "${script_path}/konflux-ci/konflux-info"
```
yftacherzog: I think we should avoid deploying stuff from the konflux-ci dir inside deploy-deps.sh. We should either move it to the other script or move the manifests to the dependencies dir.

avi-biton: right, good catch

avi-biton: Done

konflux-ci/konflux-info/banner-schema.json
avi-biton: Schema was removed

konflux-ci/konflux-info/banner-content.yaml
avi-biton: banner resources were completely removed

avi-biton: Got the banner back, without schema.
The banner that will appear:
`Welcome to Konflux-CI! This is a development environment for testing and development purposes.`
image

yftacherzog: Cool!

konflux-ci/konflux-info/banner-schema.json
avi-biton: BTW, the schema is defined in 
https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-info/banner-schema.json

I did not see it anywhere else (upstream) and it is not mandatory

yftacherzog: ok. We should not take it from downstream. Maybe we can do without it for now.

avi-biton: Currently it is working well without it :)

Issue #3408, opened 2025-10-22
regcred-empty was removed but might be still needed


yftacherzog:
https://github.com/konflux-ci/konflux-ci/pull/3263/files#diff-7cbf893fd133a761df12dd47ed7ce5035726759c5b558ca81904fe0a8ba64438

This PR removed this secret, but I think it should still be needed for working with the local registry. Need to test this scenario to verify.

PR #3263, opened 2025-09-30
Do not use deprecated 'appstudio-pipeline' service account

mmorhun:
Since we are gonna remove deprecated `appstudio-pipeline` Service Account, it should not be used in Konflux deployment.

PR #3573, opened 2025-11-04
fix(KONFLUX-10618): restore regcred-empty secret in user-ns2

hmariset:
The regcred-empty secret was incorrectly removed from user-ns2 namespace in PR #3263. This secret is required by the ecosystem-cert-preflight-checks task which expects a Docker config file to exist even when working with registries that don't require authentication.

Without this secret, pipelines fail with:
  "could not find authfile: /root/.docker/config.json"

Closes: #3408

Assisted-by: Claude-3.5-Sonnet (via Claude Code)

### DIFF
--- a/test/resources/demo-users/user/ns2/kustomization.yaml
+++ b/test/resources/demo-users/user/ns2/kustomization.yaml
@@ -4,4 +4,5 @@ resources:
   - ns.yaml
   - rbac.yaml
   - release-pipeline-sa.yaml
+  - regcred-empty.yaml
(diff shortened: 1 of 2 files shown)


yftacherzog:
@hmariset did you test the scenario of onboarding an application using k8s manifests and the local registry, including verifying on-pr, on-push and release pipelines working?

hmariset:
> @hmariset did you test the scenario of onboarding an application using k8s manifests and the local registry, including verifying on-pr, on-push and release pipelines working?

@yftacherzog Yes, verified all the above mentioned and it's working fine with this change.

PR #3980, opened 2025-12-14
Require kubectl v1.34.2+ and use version check

gbenhaim:
Require kubectl v1.31.4+ and use version check instead of feature detection

- Update minimum kubectl version from v1.31.1 to v1.31.4 (which is
used in the e2e test job).
- Replace server-side apply feature check with version comparison

Server-side apply support is guaranteed in kubectl v1.31.4+, so explicit
feature detection is no longer needed.

Ref: https://konflux-ci.slack.com/archives/C09LE7E20FP/p1765560818181239?thread_ts=1765552331.343709&cid=C09LE7E20FP
Assisted-By: Cursor
Signed-off-by: Gal Ben Haim <gbenhaim@redhat.com>

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -17,12 +17,13 @@ check_req(){
     # declare the requirements
     local requirements=(kubectl openssl)
     local uninstalled_requirements=()
+    local min_kubectl_version="v1.31.4"
 
     # check if requirements are installed
     for i in "${requirements[@]}"; do
         if ! command -v "$i" &> /dev/null; then
                 if [ "$i" == "kubectl" ]; then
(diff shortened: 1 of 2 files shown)


deploy-deps.sh
```diff
@@ -17,12 +17,13 @@ check_req(){
     # declare the requirements
     local requirements=(kubectl openssl)
     local uninstalled_requirements=()
+    local min_kubectl_version="v1.31.4"
```
yftacherzog, line 20: Need to change the readme to reflect this.

gbenhaim: done

yftacherzog (approved):

PR #4247, opened 2025-12-25
test(KFLUXVNGD-586): run kube-linter for operator manifests

yftacherzog:
The kube-linter script was not running kustomize build for anything under the operator directory. Instead we now exclude the upstream kustomizations which are pre-built by automation and include the rest (e.g. operator's own manifests).

### DIFF
--- a/.github/scripts/run-kube-linter.sh
+++ b/.github/scripts/run-kube-linter.sh
@@ -31,8 +31,26 @@ echo "Generating Kubernetes manifests..."
 echo "Creating temporary smee-channel-id.yaml for manifest generation..."
 cp dependencies/smee/smee-channel-id.tpl dependencies/smee/smee-channel-id.yaml
 
-# Exclude operator directory from kustomize builds
-find . \( -name "kustomization.yaml" -o -name "kustomization.yml" \) ! -path "*/operator/*" | while read -r file; do
+# Generate operator CSV file (required for config/manifests)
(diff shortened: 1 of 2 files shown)


.github/scripts/run-kube-linter.sh
```diff
@@ -31,8 +31,26 @@ echo "Generating Kubernetes manifests..."
...
+# Install operator-sdk using Makefile (requires Go to be set up)
+# The Makefile will use system operator-sdk if available, otherwise installs to bin/
+make operator-sdk
+# Determine operator-sdk path: use bin/operator-sdk if it exists, otherwise use system one
+if [ -f "bin/operator-sdk" ]; then
+  OPERATOR_SDK="./bin/operator-sdk"
+else
+  OPERATOR_SDK="operator-sdk"
+fi
+# Generate the CSV file needed by config/manifests/kustomization.yaml
```
gbenhaim, line 46: basically we need to generate the CSV file and commit it to this repository. In the ci we need to check if it's up to date or someone forget to generate it locally.

yftacherzog: I considered that, but I'd prefer not to go this way until we use the CSV somehow.

yftacherzog: In other words, I'm worried I'll start figuring things out for the CSV, when all I wanted was to run kube-linter properly.

PR #4250, opened 2025-12-25
chore(KFLUXVNGD-596): install image-controller with operator

yftacherzog:
Assisted-by: Cursor


operator/internal/controller/konflux_controller.go
```diff
@@ -255,6 +267,24 @@ func (r *KonfluxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 	subCRStatuses = append(subCRStatuses, CopySubCRStatus(konflux, konfluxNamespaceLister, "namespace-lister"))
 
+	// Get and copy status from the KonfluxImageController CR (if enabled and exists)
+	if konflux.Spec.IsImageControllerEnabled() {
+		imageController := &konfluxv1alpha1.KonfluxImageController{}
+		if err := r.Get(ctx, client.ObjectKey{Name: KonfluxImageControllerCRName}, imageController); err != nil {
+			if client.IgnoreNotFound(err) != nil {
```
gbenhaim: Why do you ignore not found error?  We are not doing it in for the other crs

operator/internal/controller/konflux_controller.go
```diff
@@ -496,14 +526,58 @@ func (r *KonfluxReconciler) applyKonfluxNamespaceLister(ctx context.Context, own
...
+// applyKonfluxImageController creates or updates the KonfluxImageController CR if enabled,
+// or deletes it if disabled.
+func (r *KonfluxReconciler) applyKonfluxImageController(ctx context.Context, owner *konfluxv1alpha1.Konflux) error {
+	log := logf.FromContext(ctx)
+
+	// Check if image-controller is enabled
+	if !owner.Spec.IsImageControllerEnabled() {
+		// Delete the CR if it exists
+		imageController := &konfluxv1alpha1.KonfluxImageController{}
+		if err := r.Get(ctx, client.ObjectKey{Name: KonfluxImageControllerCRName}, imageController); err != nil {
```
gbenhaim: Is it possible to call delete and ignore not found? Without calling get first?

test/e2e/prepare-e2e.sh
```diff
@@ -11,8 +11,16 @@ main() {
     local app_webhook_secret=${APP_WEBHOOK_SECRET:?GitHub App webhook secret should be provided}
     local smee_channel=${SMEE_CHANNEL:?Link to the smee.io channel should be provided}
 
+    # Enable image-controller in the Konflux CR before deploying (if operator is present)
+    if kubectl get konflux konflux &>/dev/null; then
```
gbenhaim, line 15: I suggest to just enable it in the example cr for konflux

yftacherzog: Did it originally,  but the cr won't be ready until we have the secrets in place, which this script introduces.

yftacherzog: Did it originally,  but the cr won't be ready until we have the secrets in place, which this script introduces.

operator/internal/controller/konflux_controller.go
yftacherzog: It was something that was supposed to account for a short time window in which the CR is created, but not fully present here. I think we can avoid this logic for now. I'll make it consistent.

operator/internal/controller/konflux_controller.go
yftacherzog: You're right. We should avoid this pattern.

gbenhaim (approved):

PR #4267, opened 2025-12-28
chore(KFLUXVNGD-607): operator top-level cleanup

yftacherzog:
Cleanup top-level operator now that all deployments are managed by their respective controllers.

Assisted-by: Cursor

### DIFF
--- a/operator/internal/controller/conditions.go
+++ b/operator/internal/controller/conditions.go
@@ -179,14 +179,10 @@ type SubCRStatus struct {
 	Ready bool
 }
 
-// AggregateReadiness computes the overall readiness from deployment summary and sub-CR statuses.
+// AggregateReadiness computes the overall readiness from sub-CR statuses.
 // Returns whether all components are ready and a slice of reasons for any not-ready components.
-func AggregateReadiness(deploymentSummary DeploymentStatusSummary, subCRStatuses []SubCRStatus) (allReady bool, notReadyReasons []string) {
-	allReady = deploymentSummary.AllReady
-
-	if !deploymentSummary.AllReady {
-		notReadyReasons = append(notReadyReasons, fmt.Sprintf("deployments not ready: %v", deploymentSummary.NotReadyNames))
-	}
+func AggregateReadiness(subCRStatuses []SubCRStatus) (allReady bool, notReadyReasons []string) {
+	allReady = true
 
 	for _, status := range subCRStatuses {
 		if !status.Ready {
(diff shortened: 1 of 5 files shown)


operator/internal/controller/conditions.go
```diff
@@ -309,3 +308,88 @@ func UpdateComponentStatuses(
 
 	return nil
 }
+
+// generationChangedPredicate filters out events where the generation hasn't changed
+// (i.e., status-only updates that shouldn't trigger reconciliation)
+var generationChangedPredicate = predicate.Funcs{
```
gbenhaim: not sure if the predicates should be in the conditions file

gbenhaim:
this method can be removed as well https://github.com/konflux-ci/konflux-ci/pull/4267/files#diff-747bce4ca188bad95738507fc4c701791ddcd8b8621f8e768cdccb3d0e87239aR612

yftacherzog:
> this method can be removed as well https://github.com/konflux-ci/konflux-ci/pull/4267/files#diff-747bce4ca188bad95738507fc4c701791ddcd8b8621f8e768cdccb3d0e87239aR612

@gbenhaim what method were you referring to? (Not clear to me from the diff)

operator/internal/controller/conditions.go
yftacherzog: Would you prefer them back in their original location or in something like a `predicates.go` file?

operator/internal/controller/konflux_controller.go
```diff
@@ -105,16 +98,6 @@ func (r *KonfluxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		log.Error(err, "Failed to apply manifests")
-		SetFailedCondition(konflux, ConditionTypeReady, "ApplyFailed", err)
-		if updateErr := r.Status().Update(ctx, konflux); updateErr != nil {
-			log.Error(updateErr, "Failed to update status")
-		}
-		return ctrl.Result{}, err
-	}
-
 	// Apply the KonfluxApplicationAPI CR
 	if err := r.applyKonfluxApplicationAPI(ctx, konflux); err != nil {
```
avi-biton, line 102: Don't we want to refactor this function to fit all the components in this PR as well?

yftacherzog: How do you mean? It follows the same pattern as the rest of the components did.

avi-biton: It is, but we mentioned that when all the components will be onboarded, we will refactor this part.
It can be in another PR

yftacherzog: Refactor how?

avi-biton: NM, I now see that we have a different implementation for different components (ImageController for example).
So you can ignore it :)

operator/internal/controller/conditions.go
gbenhaim: predicates.go sounds good

yftacherzog: done

gbenhaim (approved):

PR #4272, opened 2025-12-28
chore(KFLUXVNGD-604): operator customize for nlister

Omeramsc:
Allow users to configure namespace-lister via CRD spec fields: env vars, replicas, resources, and metrics settings.
    
Controller applies overrides to the deployment before reconciling.
All fields optional with sensible defaults.

Assisted-by: Cursor


operator/api/v1alpha1/konfluxnamespacelister_types.go
yftacherzog, line 1: We need the change here to also be configurable via `operator/api/v1alpha1/konflux_types.go` (see UI and integration PRs). The user provides only the top level CR, and the content needs to be translated (not necessarily 1:1) to the underlying CRs.

operator/api/v1alpha1/konflux_types.go
```diff
@@ -57,6 +59,14 @@ type KonfluxUIConfig struct {
 	Spec *KonfluxUISpec `json:"spec,omitempty"`
 }
 
+// NamespaceListerConfig defines the configuration for the namespace-lister component.
+// The Spec field is the runtime configuration passed to the component.
+type NamespaceListerConfig struct {
+	// Spec configures the namespace-lister component.
+	// +optional
+	Spec *KonfluxNamespaceListerSpec `json:"spec,omitempty"`
```
yftacherzog: I suggest dropping the konflux prefix in this case.
```suggestion
	Spec *NamespaceListerSpec `json:"spec,omitempty"`
```

Omeramsc, line 106: I kept it like this to be consistent with the UI component having both CR name and spec with the Konflux prefix. any reason it should be treated differently here?

yftacherzog: I meant to comment on the Config struct. We should not change the existing type.

avi-biton:
Although its unrelated to your PR, we should update `konfluxnamespacelister_types.go`.
The type is defined:
```
type KonfluxNamespaceLister struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   KonfluxNamespaceListerSpec   `json:"spec,omitempty"`
	Status KonfluxNamespaceListerStatus `json:"status,omitempty"`
}
```

We should update it to:
```
type KonfluxNamespaceLister struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   KonfluxNamespaceListerSpec   `json:"spec"`
	Status KonfluxNamespaceListerStatus `json:"status,omitempty"`
}
```
Remove the `omitempty` from `Spec   KonfluxNamespaceListerSpec`

avi-biton (approved):

PR #4277, opened 2025-12-28
feat(operator): Add Dex IdP configuration generation

gbenhaim:
- Add pkg/dex package with Config structs and NewDexConfig builder
- Add DexConfig to KonfluxUI CRD for configurable Dex IdP settings
- Implement Dex ConfigMap reconciliation with content-based hash naming
- Add WithConfigMapVolumeUpdate to customization package for updating existing ConfigMap volume references in deployments
- Support OpenShift connector auto-configuration via ConfigureLoginWithOpenShift
- Delete the dex config from the manifests (since it's now generated) by the operator, and configure static users through the konflux CR.

Assisted-By: Cursor


operator/internal/controller/konfluxui_controller.go
```diff
@@ -115,9 +145,33 @@ func (r *KonfluxUIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	return ctrl.Result{}, nil
 }
 
+func (r *KonfluxUIReconciler) ensureNamespaceExists(ctx context.Context, owner *konfluxv1alpha1.KonfluxUI) error {
```
yftacherzog, line 148: What's the reason we now need this function?

operator/internal/controller/konfluxui_controller.go
```diff
@@ -268,6 +325,46 @@ func generateRandomBytes(length int, urlSafe bool) ([]byte, error) {
...
+func (r *KonfluxUIReconciler) reconcileDexConfigMap(ctx context.Context, ui *konfluxv1alpha1.KonfluxUI) (string, error) {
+	var dexConfig *dex.Config
+	// Check if DexConfig is configured
+	if ui.Spec.Dex != nil && ui.Spec.Dex.Config != nil {
+		dexConfig = dex.NewDexConfig(ui.Spec.Dex.Config)
+	} else {
+		dexConfig = dex.NewDexConfig(
+			&dex.DexParams{
+				Hostname: "localhost",
+				Port:     "9443",
```
yftacherzog, line 340: Do we need to set the defaults explicitly here? The definition of `DexParams` will not have them set anyway?

operator/pkg/dex/builder.go
```diff
@@ -0,0 +1,120 @@
...
+	return &Config{
+		Issuer: fmt.Sprintf("%s/idp/", baseURL),
+		Storage: &Storage{
+			Type: "kubernetes",
+			Config: &StorageConfig{
+				InCluster: true,
+			},
+		},
+		Web: &Web{
+			HTTPS:   "0.0.0.0:9443",
```
yftacherzog, line 95: Should `9443`  be hard-coded here or taken from `Port`?

operator/internal/controller/konfluxui_controller.go
gbenhaim: because I need to create resource in this order:

1. namespace
2. render the dex config into a configmap in the namespace created in 1
3. inject the name of the configmap to the dex deployment config

operator/internal/controller/konfluxui_controller.go
gbenhaim: I will add defaults using kubebuilder in the next PRs.

operator/pkg/dex/builder.go
gbenhaim: ATM I don't see a reason for making the port configurable since we create the service for dex.

PR #4283, opened 2025-12-29
chore(KFLUXVNGD-600): ReleaseService - add resource and replica customization 

avi-biton:
- Add the ability to customize resource requests/limits and replica counts for the release service controller-manager owned by KonfluxReleaseService CR.
- Add unit tests for the customization logic.
- Move `ControllerManagerDeploymentSpec` to `common_types.go` to reuse it across KonfluxIntegrationService and KonfluxReleaseService CRs. Example KonfluxReleaseService spec:
```
spec:
  releaseControllerManager:
    replicas: 1
    manager:
      resources:
        requests:
          cpu: 30m
          memory: 128Mi
        limits:
          cpu: 30m
          memory: 128Mi
```

Assisted-By: Cursor


operator/internal/controller/konfluxreleaseservice_controller.go
```diff
@@ -126,6 +140,35 @@ func (r *KonfluxReleaseServiceReconciler) applyManifests(ctx context.Context, ow
...
+// buildReleaseControllerManagerOverlay builds the pod overlay for the controller-manager deployment.
+func buildReleaseControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec) *customization.PodOverlay {
+	if spec == nil {
+		return customization.NewPodOverlay()
+	}
+
+	return customization.BuildPodOverlay(
+		customization.DeploymentContext{Replicas: spec.Replicas},
+		customization.WithContainerBuilder(
+			releaseManagerContainerName,
```
gbenhaim, line 166: I think that I've missed it in the pr for integration service, but for most controllers when the number of replicas is >1, leader election should be enabled, thus you would need to add this option as well https://github.com/konflux-ci/konflux-ci/blob/main/operator/pkg/customization/container.go#L98

avi-biton: I added leader election for both integration and release

gbenhaim (approved):

PR #4284, opened 2025-12-29
Dex Configuration Improvements and Stability Fixes

gbenhaim:
This update enhances the stability and flexibility of the Dex identity service integration by automating connector setup and improving configuration handling.

**Key Enhancements**

**Automatic Redirect Management:** The system now automatically sets the redirectURI for Dex connectors if not explicitly provided by the user.

**Environment Variable Support:** Added support for configuring containers via environment variables. This is specifically required for securely passing sensitive credentials, such as GitHub Client IDs and Secrets, to Dex connectors.

**Crash Prevention:** Implemented a safeguard that automatically enables the local password database if no external connectors are defined. This prevents Dex from entering a crash loop when no authentication source is provided.

**Documentation:** Included a new Konflux CR example demonstrating how to properly configure GitHub as an authentication provider.

Assisted-By: Cursor

### DIFF
--- a/operator/config/samples/konflux-empty-cr.yaml
+++ b/operator/config/samples/konflux-empty-cr.yaml
@@ -1,0 +1,8 @@
+apiVersion: konflux.konflux-ci.dev/v1alpha1
+kind: Konflux
+metadata:
+  labels:
+    app.kubernetes.io/name: konflux-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: konflux
+spec: {}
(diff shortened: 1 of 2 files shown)


operator/config/samples/konflux-with-github-auth.yaml
yftacherzog, line 1: I wonder how do we make sure that we don't track samples of dead code? It's good to have examples, but we need something that actually use them to make sure they are still valid.

avi-biton: We might want to add tests for those examples

gbenhaim: I suggest to create a followup ticket on deciding which samples we want to keep and how to test them.

PR #4296, opened 2025-12-30
chore(KFLUXVNGD-608): Reduce RBAC scope

avi-biton:
- Update all controllers RBAC rules to reduce scope.
- Set each controller with the minimal RBAC rules
required to perform its duties.
- Move aggregating ClusterRoles (konflux-*-user-actions) to operator installation
  to avoid requiring cluster-admin at runtime
- Separate RBAC resources into `operator-rbac` and `user-rbac` directories to
avoid prefix addition to the `user-rbac` resources.
- Add `escalate` and `bind` verbs to needed resources.
- Because `escalate` and `bind` verbs are very powerful, restrict them by resource names.

Assisted-By: Cursor


yftacherzog (approved):

operator/internal/controller/konfluxapplicationapi_controller.go
```diff
@@ -44,6 +44,8 @@ type KonfluxApplicationAPIReconciler struct {
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxapplicationapis,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxapplicationapis/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxapplicationapis/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=list
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;watch;patch
```
yftacherzog: Cursor says:

```suggestion
// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=create;get;watch;patch
```

```
Root Cause
The KonfluxApplicationAPI controller uses server-side apply (client.Apply) to manage CRDs. When a CRD doesn't exist, Apply needs create to create it first.
The RBAC marker only had:
verbs=get;watch;patch
It was missing create, so applying new CRDs failed with a permissions error, preventing the controller from setting the Ready condition.
The Fix
Added create to the RBAC marker:
verbs=get;watch;patch
The manifest has been regenerated and now includes create in the role. This should allow the e2e tests to pass, as the controller can now create CRDs when they don't exist.
```

I Didn't verify it actually makes sense.

operator/config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: example.com/konflux-operator
-  newTag: v0.0.1
+  newTag: v0.0.3
```
yftacherzog: Is this intentional?

operator/internal/controller/konfluxrbac_controller.go
```diff
@@ -103,6 +107,28 @@ func (r *KonfluxRBACReconciler) applyManifests(ctx context.Context, owner *konfl
...
+		// Try typed ClusterRole first
+		if cr, ok := obj.(*rbacv1.ClusterRole); ok {
+			if cr.AggregationRule != nil {
+				log.Info("Skipping ClusterRole with aggregationRule (managed by operator installation)",
+					"name", cr.GetName())
+				continue
+			}
+		}
+
+		// Fall back to unstructured for non-typed objects
```
gbenhaim: not sure I understand why you need to use unstructured here. In line 115 you cast the object to ClusterRole, why not using it?

operator/config/manager/kustomization.yaml
avi-biton: Done for the tests, I will revert it

operator/config/rbac/role.yaml
```diff
@@ -63,3 +194,48 @@ rules:
...
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterrolebindings
+  - clusterroles
+  - rolebindings
+  - roles
+  verbs:
+  - bind
```
gbenhaim, line 483: can we limit bind and escalate to specific resource names?

gbenhaim (changes requested):
left a question about making the permissions stricter

operator/internal/controller/konfluxrbac_controller.go
avi-biton: Its a fallback, but now when I think about it, we might not need it.
We can potentially move the `konflux-admin-user*` manifests from `operator/upstream-kustomizations/rbac/core` to `operator/config/rbac` and remove all the code section of skipping these `ClusterRoles`.
I will give it a try

avi-biton: Removed this part as its no longer needed

operator/internal/controller/konfluxapplicationapi_controller.go
avi-biton: Done

operator/config/rbac/role.yaml
avi-biton: Done

operator/config/default/user-rbac/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+# User-facing RBAC resources
+# These ClusterRoles are provided for cluster admins to manage user permissions
+# and must NOT have a namePrefix applied (users reference them by exact name in RoleBindings)
+resources:
```
gbenhaim, line 7: maybe add to `make run` a command for applying this kustomization on the cluster, so it would be still possible to test locally without deploying the operator to the cluster

gbenhaim (approved):
lgtm
small comment

operator/config/default/user-rbac/kustomization.yaml
avi-biton: Done

PR #4299, opened 2025-12-30
OCP deployments fixes

gbenhaim:
Adjust UI for OCP deployment
    
    1. Detect tekton-results hostname during runtime.
    2. Add the DNS name of the proxy service to the certificate.
    3. Move idp location to the nginx config since we are not going
       to make it optional anytime soon.
   
    Make trust bundle CM optional for the build-service
    
    If the trust-bundle doesn't exist there is no reason the build-service
    pod should fail, it can operate without it.

    Adjust ns lister network policy for OCP
    
    1. Add the allow list for the api server pods on OCP.
    2. Allow port 443, which is the API port on OCP 4.20.
    
    Signed-off-by: Gal Ben Haim <gbenhaim@redhat.com>


operator/upstream-kustomizations/ui/core/proxy/proxy.yaml
```diff
@@ -64,6 +64,20 @@ spec:
...
+            results_hostname_k8s=tekton-results-api-service.tekton-pipelines.svc.cluster.local
+            results_hostname_openshift=tekton-results-api-service.openshift-pipelines.svc.cluster.local
+            if nslookup $results_hostname_k8s > /dev/null 2>&1; then
+              results_hostname=$results_hostname_k8s
+            elif nslookup $results_hostname_openshift > /dev/null 2>&1; then
+              results_hostname=$results_hostname_openshift
+            else
+              echo "Error: Could not resolve tekton-results hostname"
+              exit 1
+            fi
```
yftacherzog, line 77: Is results usually up at this point? How likely are we to get into crashloopbackoff which will make UI always start slowly?
I don't have the startup sequence in my head, but if they are happening at the same time and this is likely to fail a few times, maybe it will make sense to add retries here to prevent it from taking a few minutes for the UI to start due to backoff delay.

gbenhaim: Lets implement if if we see an actual issue. Using results is temporary, we will move to use Kubearchive and then removing this code.

PR #4285, opened 2025-12-29
feat(KFLUXVNGD-615): add orphan resource cleanup after upgrades

Omeramsc:
When the operator is upgraded and its manifests change (resources removed or renamed), the old resources would remain in the cluster. This change implements a label-based approach to detect and clean up such orphans.

Flow:
```
                    ┌─────────────────────────────┐
                    │    Operator Starts          │
                    │  (compute binary digest)    │
                    └──────────────┬──────────────┘
                                   │
                                   ▼
                    ┌─────────────────────────────┐
                    │    Reconcile Component      │
                    └──────────────┬──────────────┘
                                   │
                    ┌──────────────┴──────────────┐
                    ▼                             │
         ┌──────────────────────┐                 │
         │  Apply Resource      │                 │
         │  (set labels:        │                 │
         │   component=X        │                 │
         │   operator-version=  │                 │
         │     abc123...)       │                 │
         └──────────┬───────────┘                 │
                    │                             │
                    │ ◄────── repeat ─────────────┘
                    ▼
         ┌──────────────────────┐
         │ PruneOrphanedResources                │
         └──────────┬───────────┘
                    │
                    ▼
    ┌───────────────────────────────────┐
    │  For each resource type:          │
    │  List where component=X           │
    │                                   │
    │  operator-version != current?     │
    │  YES ──► DELETE (orphaned)        │
    │  NO  ──► KEEP                     │
    │  MISSING ──► SKIP                 │
    └───────────────────────────────────┘
```
New files:
- pkg/version/digest.go: Binary digest computation with caching
- internal/controller/orphan_cleanup.go: Orphan detection and cleanup

Integrated with namespace-lister controller as the first implementation. Other controllers can adopt the same pattern.

assisted-by: Cursor

PR #4301, opened 2025-12-31
feat(ui): add ingress configuration support for KonfluxUI

gbenhaim:
- Add IngressSpec with enabled, host, ingressClassName, annotations, and TLS fields
  - Create pkg/ingress package for ingress resource management
  - Auto-detect OpenShift ingress domain when host is not specified
  - Include OpenShift TLS re-encryption annotations by default
  - Delete Ingress resource when ingress is disabled in CR
  - Propagate UI URL to KonfluxUI and Konflux CR status
  - Add kubectl print columns for Ready status and UI URL

requires - https://github.com/konflux-ci/konflux-ci/pull/4300
Assisted-By: Cursor


operator/internal/controller/konfluxui_controller.go
```diff
@@ -386,5 +476,46 @@ func (r *KonfluxUIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	// If ingress is not enabled, delete the Ingress resource if it exists
+	if ui.Spec.Ingress == nil || !ui.Spec.Ingress.Enabled {
+		existingIngress := &networkingv1.Ingress{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ingress.IngressName,
+				Namespace: uiNamespace,
+			},
+		}
+		if err := r.Delete(ctx, existingIngress); err != nil {
```
yftacherzog, line 496: Do we want delete logic inside the reconciler?
How is this different than any other changed object that we'd modify?

This goes back to what we discussed the other day about handling configs we created but don't want anymore. Is there anything special about this case for which we'd like to address it explicitly here rather than solving it in a unified manner later on?

gbenhaim: If ingres is enabled and then disabled the ingress resource should be deleted. I don't see a way to generalize it to all resources it is possible to move this reconcile method to the ingress package so it can be used for other ingress resources (but ATM I don't expect to have any other optional ingress resources).

yftacherzog: This seems to me like a specific case of what we've been trying to address in https://github.com/konflux-ci/konflux-ci/pull/4285. No? Is the specific scenario of an ingress being deleted is something likely that should be addressed specifically, rather than addressing it in a unified way when we try to redo that mentioned PR?

gbenhaim: lets merge this one and then I look into a generic way to cleanup resources.

yftacherzog (approved):

avi-biton (approved):

PR #4634, opened 2026-01-20
Default tenant

gbenhaim:
**Configure ns lister to allow public view**

ns lister was missing the configuration for enabling public view
of namespaces for authenticated users.

**feat(operator): add default tenant support**

Add a default tenant namespace that is automatically created with every
Konflux deployment, accessible by all authenticated users.

Resources created in 'default-tenant' namespace:
- Namespace with konflux-ci.dev/type: tenant label
- trusted-ca ConfigMap for CA bundle injection
- konflux-integration-runner ServiceAccount and RoleBinding
- RoleBinding granting system:authenticated maintainer access

Enabled by default, can be disabled via spec.defaultTenant.enabled: false

### DIFF
--- a/operator/pkg/manifests/namespace-lister/manifests.yaml
+++ b/operator/pkg/manifests/namespace-lister/manifests.yaml
@@ -95,7 +95,9 @@ spec:
         - name: CACHE_NAMESPACE_LABELSELECTOR
           value: konflux-ci.dev/type=tenant
         - name: AUTH_USERNAME_HEADER
-          value: Impersonate-User
+          value: X-User
+        - name: AUTH_GROUPS_HEADER
+          value: X-Group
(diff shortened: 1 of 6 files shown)


operator/config/samples/konflux_v1alpha1_konfluxdefaulttenant.yaml
```diff
@@ -0,0 +1,11 @@
+# Title: Konflux Default Tenant Configuration
+# Description: KonfluxDefaultTenant configuration for default tenant
+apiVersion: konflux.konflux-ci.dev/v1alpha1
+kind: KonfluxDefaultTenant
+metadata:
+  labels:
+    app.kubernetes.io/name: konflux-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: konfluxdefaulttenant-sample
```
yftacherzog, line 9:
```suggestion
  name: konflux-default-tenant
```

PR #4654, opened 2026-01-21
fix: remove notifications from konflux-info CR

yftacherzog:
The konflux-info part of the main top-level PR includes notifications, which caused onboarding of applications to fail. This commit removes this bit.

Assisted-by: Cursor


avi-biton (commented):
Don't we need this part to enable imageController?
```
        integrations:
          image_controller:
            enabled: true
```
Or is it only for the integration of image controller and konflux-info?

gbenhaim (approved):

PR #4665, opened 2026-01-21
chore(KFLUXVNGD-672): GitHub workflow for making sure releases are created

avi-biton:
This workflow verifies that a release was created during the last 7 days in case commits were made to the main branch. 
In case no release was created but commits were made, an issue is created to remind the user to create a release.

This workflow runs periodically a day after the the `create-release` is running

In any case that the workflow fails, an issue is created to remind the user to check the workflow logs and fix the issue.

Assisted-By: Cursor

### DIFF
--- a/.github/scripts/verify-releases.sh
+++ b/.github/scripts/verify-releases.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+set -euo pipefail
+
+# Verify Releases Script
+# This script checks if a release was created in the past 7 days and if commits were made
+#
+# Usage:
+#   verify-releases.sh <repository>
+#
+# Arguments:
+#   repository - Repository in format owner/repo (required)
+#
+# Environment Variables:
+#   GH_TOKEN      - GitHub token for API access (required)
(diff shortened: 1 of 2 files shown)


.github/scripts/verify-releases.sh
```diff
@@ -0,0 +1,119 @@
...
+#   GH_TOKEN      - GitHub token for API access (required)
+#   GITHUB_ENV    - Path to GitHub Actions environment file (automatically set by GitHub Actions)
+#
+# The script sets the following environment variables (via GITHUB_ENV):
+#   COMMIT_COUNT         - Number of commits found in the past 7 days
+#   VERIFICATION_FAILED  - "true" if verification failed (no release but commits exist), unset otherwise
+#
+# Exit Codes:
+#   0 - Success (release exists OR no commits)
+#   1 - Expected failure: No release found but commits exist (should create issue)
```
gbenhaim: I suggest to use 1 as the not expected error since this is what set -e will report.

avi-biton: Done

PR #4687, opened 2026-01-22
add PAC_WEBHOOK_URL env to image-controller

tisutisu:
this change will make PAC_WEBHOOK_URL to be available in image-controller env

### DIFF
--- a/konflux-ci/image-controller/core/image-controller-env-patch.yaml
+++ b/konflux-ci/image-controller/core/image-controller-env-patch.yaml
@@ -1,0 +1,12 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: image-controller-controller-manager
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          env:
+            - name: PAC_WEBHOOK_URL
+              value: http://pipelines-as-code-controller.pipelines-as-code.svc.cluster.local:8180
--- a/konflux-ci/image-controller/core/kustomization.yaml
+++ b/konflux-ci/image-controller/core/kustomization.yaml
@@ -29,3 +29,7 @@ patches:
       kind: CronJob
       name: image-controller-image-pruner-cronjob
     path: replace-pruner-image.yaml
+  - target:
+      kind: Deployment
+      name: image-controller-controller-manager
+    path: image-controller-env-patch.yaml


konflux-ci/image-controller/core/image-controller-env-patch.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Deployment
+metadata:
+  name: image-controller-controller-manager
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          env:
+            - name: PAC_WEBHOOK_URL
```
gbenhaim, line 11: Why image controller needs to know about pac? Also, we are going to change the deployment to use the operator. I think we need the operator to set the url automatically, supporting both upstream tekton and openshift pipelines l.

rcerven: because new model is using pac route, to extract cluster hostname, to include it to namespace robot account https://github.com/konflux-ci/image-controller/pull/267

gbenhaim: Why the namespace robot account needs the clusters host name? How will it work for non openshift clusters which doesn't support routes?

rcerven: because we have multiple clusters, and each namespace SA has to be PER each cluster, so hostname is included in the SA name

rcerven: "How will it work for non openshift clusters which doesn't support routes?"
what do you mean, konflux needs PaC and PaC needs route

gbenhaim: So this is needed in order to create different robot accounts for namespaces with the same name which exists on different clusters?

rcerven: exactly, as the same namespace can be on multiple clusters, ofc with different repos to which it should have access

rcerven: also btw, the same is done for build-service https://github.com/konflux-ci/konflux-ci/blob/main/konflux-ci/build-service/core/build-service-env-patch.yaml

gbenhaim: Is it possible to use the namespace uuid instead?

rcerven: it would be possible, but I don't see reason why, build-service is already using and requires that PaC route

ALSO if I'd use namespace UID, then it wouldn't be clear from namespace SA name to which cluster it belongs, and you would have search all clusters

gbenhaim: 1. Build service is not image controller. While in the build service there is a reason to add dependency in on PAC, I don't se a reason of doing it in image controller as well.

2. The pac url used in this pr is for PAC's service. Different clusters would have the same service url, so you don't get uniqueness. 
Not exposing PAC using a route is a valid use case, for example when running a cluster behind a firewall/vpn and using a solution like smee for pulling the webhooks.

4. You can't assume that you can get the clusters ingress domain from PAC's route, since it may use another host name that doesn't match the cluster's ingress domain.

5. What is the use case of locating the cluster given a robot account?

rcerven: 1. right, but image-controller is used with build-service, I don't think it has much value as standalone controller

2. each konflux cluster which is using the same quay org, HAS to have each PaC route different, because otherwise you wouldn't be able to trigger builds on specific clusters (via webhooks)

3. I am not looking for "correct" ingress name, but "some" cluster identification 

4. well that was example, it just looks way better to have namespace SA with "clustername" instead some chars (like with uid)

I don't object to changing it to to use namespace UID, I just don't see any strong arguments

gbenhaim: The main argument is that the host name is mutable thus imo its preferred to not use it as an identifier.

Another option is to use the uuid of the kube-system namespace, this will also let you identify to which cluster the robot account belongs.

mmorhun: I'd prefer something more readable than UUID of `kube-system` namespace, but probably introducing dependency on PaC doesn't worth it in long run.

PR #4730, opened 2026-01-26
fix(KFLUXVNGD-693): use latest release pipelines

yftacherzog:
Use the latest release-service-catalog revision, providing a token to use the trusted artifacts repo it requires for running pipelines.

Assisted-by: Cursor

### DIFF
--- a/test/e2e/run-e2e.sh
+++ b/test/e2e/run-e2e.sh
@@ -6,10 +6,12 @@ main() {
     echo "Running E2E tests" >&2
     # shellcheck disable=SC1091
     source "${script_path}/vars.sh"
+
     local github_org="${GH_ORG:?A GitHub org where the https://github.com/redhat-appstudio-qe/konflux-ci-sample repo is present/forked should be provided}"
     local github_token="${GH_TOKEN:?A GitHub token should be provided}"
     local quay_dockerconfig="${QUAY_DOCKERCONFIGJSON:?quay.io credentials in .dockerconfig format should be provided}"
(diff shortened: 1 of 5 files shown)


test/e2e/run-e2e.sh
```diff
@@ -9,7 +9,8 @@ main() {
     local github_org="${GH_ORG:?A GitHub org where the https://github.com/redhat-appstudio-qe/konflux-ci-sample repo is present/forked should be provided}"
     local github_token="${GH_TOKEN:?A GitHub token should be provided}"
     local quay_dockerconfig="${QUAY_DOCKERCONFIGJSON:?quay.io credentials in .dockerconfig format should be provided}"
-    local catalog_revision="${RELEASE_SERVICE_CATALOG_REVISION:-development}"
+    local catalog_revision="development"
+    local release_catalog_ta_quay_token="${RELEASE_CATALOG_TA_QUAY_TOKEN:-}"
```
gbenhaim, line 14: Do you think we should pin to a commit and track a branch using renovate? This way we can catch at the pr level breaking change
 Wdyt?

yftacherzog: Yes. You're right. Once I make it work, I'll pin it and see how we get it tracked.

yftacherzog: It's now tracked in https://github.com/konflux-ci/konflux-ci/pull/4730/changes#diff-bd25b6440275685258f506f8e8a0cbb1782b31b2737647b615ba84e0147af52fR7

PR #4734, opened 2026-01-26
chore(KFLUXVNGD-652): Inject console url template to the integration-service

avi-biton:
The console url is injected to the integration-service deployment container spec.

- Added `KonfluxUIIngressStatusChangedPredicate` to filter out status updates that don't affect the ingress configuration.
- Added `filterConsoleURLFromSpec` to remove the CONSOLE_URL environment variable from the user-provided spec.
- Added a watch on the `KonfluxUI` CR to update the console url in the integration-service deployment.
- Added tests for the console url injection.
- Added tests to predicate

Assisted-By: Cursor

### DIFF
--- a/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
+++ b/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
@@ -19,20 +19,24 @@ package integrationservice
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	konfluxv1alpha1 "github.com/konflux-ci/konflux-ci/operator/api/v1alpha1"
 	"github.com/konflux-ci/konflux-ci/operator/internal/condition"
 	"github.com/konflux-ci/konflux-ci/operator/internal/constant"
+	"github.com/konflux-ci/konflux-ci/operator/internal/controller/ui"
 	"github.com/konflux-ci/konflux-ci/operator/internal/predicate"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/customization"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/manifests"
@@ -74,6 +78,7 @@ type KonfluxIntegrationServiceReconciler struct {
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxintegrationservices,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxintegrationservices/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxintegrationservices/finalizers,verbs=update
+// +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxuis,verbs=get;list;watch
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;patch
 // +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;patch
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;create;patch
@@ -119,8 +124,19 @@ func (r *KonfluxIntegrationServiceReconciler) Reconcile(ctx context.Context, req
 		FieldManager:      FieldManager,
 	})
 
+	// Fetch KonfluxUI to get console URL
+	konfluxUI := &konfluxv1alpha1.KonfluxUI{}
+	consoleURL := ""
+	if err := r.Get(ctx, types.NamespacedName{Name: ui.CRName}, konfluxUI); err != nil {
+		// Log warning but don't fail - URL might not be available yet
+		log.Info("KonfluxUI not found, console URL will not be set", "error", err)
+	} else if konfluxUI.Status.Ingress != nil && konfluxUI.Status.Ingress.URL != "" {
+		consoleURL = konfluxUI.Status.Ingress.URL
+		log.Info("Found console URL from KonfluxUI", "url", consoleURL)
+	}
+
 	// Apply all embedded manifests
-	if err := r.applyManifests(ctx, tc, integrationService); err != nil {
+	if err := r.applyManifests(ctx, tc, integrationService, consoleURL); err != nil {
 		return errHandler.HandleApplyError(ctx, err)
 	}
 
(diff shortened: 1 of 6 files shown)


operator/internal/predicate/predicate.go
```diff
@@ -108,3 +110,47 @@ var LabelsOrAnnotationsChangedPredicate = predicate.Funcs{
 		return true
 	},
 }
+
+// KonfluxUIIngressStatusChangedPredicate triggers reconciliation only when the Status.Ingress field changes in KonfluxUI CR.
+// This predicate filters out status updates that don't affect the ingress configuration.
+var KonfluxUIIngressStatusChangedPredicate = predicate.Funcs{
```
gbenhaim, line 116: can you please add a test for this?

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -185,35 +200,104 @@ func (r *KonfluxIntegrationServiceReconciler) applyManifests(ctx context.Context
...
+	return filtered
+}
+
 // buildControllerManagerOverlay builds the pod overlay for the controller-manager deployment.
-func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec) *customization.PodOverlay {
-	if spec == nil {
+func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec, consoleURL string) *customization.PodOverlay {
+	// Build container options
+	var managerSpec *konfluxv1alpha1.ContainerSpec
+	replicas := int32(0)
```
gbenhaim: this looks wrong. default should be 1.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -185,35 +200,104 @@ func (r *KonfluxIntegrationServiceReconciler) applyManifests(ctx context.Context
...
 		return customization.NewPodOverlay()
 	}
 
+	// Build overlay with container options
 	return customization.BuildPodOverlay(
-		customization.DeploymentContext{Replicas: spec.Replicas},
+		customization.DeploymentContext{Replicas: replicas},
 		customization.WithContainerBuilder(
 			managerContainerName,
-			customization.FromContainerSpec(spec.Manager),
```
gbenhaim, line 211: I think that moving the two existing container options to another function adds indirection which is confusing. 
can we keep them here ?

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -185,35 +200,104 @@ func (r *KonfluxIntegrationServiceReconciler) applyManifests(ctx context.Context
...
+	opts = append(opts, customization.WithLeaderElection())
+
+	// Add console URL environment variable if available.
+	// This is added after user customizations to ensure the system value takes precedence.
+	// The URL must match the actual KonfluxUI deployment for PaC links to work correctly.
+	if consoleURL != "" {
+		opts = append(opts, customization.WithEnv(
+			corev1.EnvVar{
+				Name:  "CONSOLE_URL",
+				Value: consoleURL,
```
gbenhaim: console url is a template string. you can see an example here https://github.com/redhat-appstudio/infra-deployments/blob/main/components/integration/production/kflux-ocp-p01/console-url-config-patch.json

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Done

operator/internal/predicate/predicate.go
avi-biton: Tests added

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -185,35 +200,104 @@ func (r *KonfluxIntegrationServiceReconciler) applyManifests(ctx context.Context
...
+	// If no container options, return empty overlay
+	if len(containerOpts) == 0 {
 		return customization.NewPodOverlay()
 	}
 
+	// Build overlay with container options
 	return customization.BuildPodOverlay(
-		customization.DeploymentContext{Replicas: spec.Replicas},
+		customization.DeploymentContext{Replicas: replicas},
 		customization.WithContainerBuilder(
```
gbenhaim, line 237: lets add a new customization option to the customization package.
I suggest the name `customization.WithEnvOverride`, it will add an environment variable to a container, overriding existing env var with the same name. I also think it should be safe to include the variable with an empty string in case we don't have a value yet.

This approach will replace the `filterConsoleURLFromSpec` function which is unique to this reconciler.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: I updated to be the same as the template in the example:
```
	if consoleURL != "" {
		consoleURLTemplate := fmt.Sprintf("%s/ns/{{ .Namespace }}/pipelinerun/{{ .PipelineRunName }}",
			strings.TrimSuffix(consoleURL, "/"))
		opts = append(opts, customization.WithEnv(
			corev1.EnvVar{
				Name:  "CONSOLE_URL",
				Value: consoleURLTemplate,
			},
		))
	}
```
And also updated the tests accordingly

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Done

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Addressed this issue in https://github.com/konflux-ci/konflux-ci/pull/4734#discussion_r2727587418

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -119,8 +124,19 @@ func (r *KonfluxIntegrationServiceReconciler) Reconcile(ctx context.Context, req
 		FieldManager:      FieldManager,
 	})
 
+	// Fetch KonfluxUI to get console URL
+	konfluxUI := &konfluxv1alpha1.KonfluxUI{}
+	consoleURL := ""
+	if err := r.Get(ctx, types.NamespacedName{Name: ui.CRName}, konfluxUI); err != nil {
+		// Log warning but don't fail - URL might not be available yet
+		log.Info("KonfluxUI not found, console URL will not be set", "error", err)
+	} else if konfluxUI.Status.Ingress != nil && konfluxUI.Status.Ingress.URL != "" {
```
gbenhaim, line 133: @avi-biton @yftacherzog   do you think that in this case the status of the integration cr should be not ready?

operator/internal/controller/integrationservice/konfluxintegrationservice_customization_test.go
```diff
@@ -351,3 +356,185 @@ func TestApplyIntegrationServiceDeploymentCustomizations_ResourceMerging(t *test
...
+func findEnvVar(env []corev1.EnvVar, name string) *corev1.EnvVar {
+	for i := range env {
+		if env[i].Name == name {
+			return &env[i]
+		}
+	}
+	return nil
+}
+
+func TestBuildControllerManagerOverlay_ConsoleURL(t *testing.T) {
```
gbenhaim, line 370: We need a test case which checks that the console url gets update when it already has a value.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: I am not that sure, I opened a thread to discuss it:
https://redhat-internal.slack.com/archives/C07MRFX3ULS/p1769509922519699

operator/internal/controller/integrationservice/konfluxintegrationservice_customization_test.go
avi-biton: Test case was added

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: As discussed, we will leave it as it is for now and update it in the future in case we will need to.

PR #4746, opened 2026-01-26
feat: add operator-based local deployment tooling

arewm:
## Summary

Create installation script to work with the operator-based install method and E2E test functionality.

**Key changes:**
- Local development scripts for Kind cluster setup with ARM64 support without the use of qemu/emulation.
- Integrated prerequisite deployment (Tekton, cert-manager, trust-manager).
- Change nginx image from a x86 digest to a multi-arch manifest list.
- Unified script to deploy the operator locally while creating a KinD cluster which is also utilized by e2e tests.
- Call out configuration differences between testing and production deployments including use of default users (the default CR which deploys the users has a banner alerting users to the default passwords).

## Testing

Tested on macOS (Apple Silicon) with Kind + Podman.

Co-authored-by: Claude Code (Sonnet 4.5) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/operator-test-e2e.yaml
+++ b/.github/workflows/operator-test-e2e.yaml
@@ -146,7 +146,7 @@ jobs:
             image_suffix: "-arm64"
             cluster_suffix: "-arm"
             test_scope: "integration"
-            konflux_cr: "operator/config/samples/konflux_v1alpha1_konflux.yaml"
+            konflux_cr: "operator/config/samples/konflux-e2e.yaml"
             job_name: "Run Operator Integration Tests (ARM64)"
     steps:
       - name: Skip test - no operator changes
@@ -178,7 +178,7 @@ jobs:
             image_suffix: "-arm64"
             cluster_suffix: "-arm"
             test_scope: "integration"
-            konflux_cr: "operator/config/samples/konflux_v1alpha1_konflux.yaml"
+            konflux_cr: "operator/config/samples/konflux-e2e.yaml"
             job_name: "Run Operator Integration Tests (ARM64)"
     env:
       # renovate: datasource=github-releases depName=kubernetes-sigs/kind
@@ -254,9 +254,9 @@ jobs:
           GITHUB_APP_ID: ${{ secrets.APP_ID }}
           WEBHOOK_SECRET: ${{ secrets.APP_WEBHOOK_SECRET }}
 
-          # Quay secrets for E2E tests (standardized names)
-          QUAY_TOKEN: ${{ matrix.test_scope == 'integration+e2e' && secrets.QUAY_TOKEN || '' }}
-          QUAY_ORGANIZATION: ${{ matrix.test_scope == 'integration+e2e' && secrets.QUAY_ORG || '' }}
+          # Quay secrets for image-controller (required by konflux-e2e.yaml CR)
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+          QUAY_ORGANIZATION: ${{ secrets.QUAY_ORG }}
 
           # Smee webhook forwarding for E2E tests only
           SMEE_CHANNEL: ${{ matrix.test_scope == 'integration+e2e' && secrets.SMEE_CHANNEL || '' }}


docs/mac-setup.md
```diff
@@ -0,0 +1,395 @@
...
+
+Install required tools using Homebrew:
+
+```bash
+brew install kind kubectl podman
+```
+
+## Podman Installation
+
+Podman replaces Docker on macOS for running Konflux. It provides a lightweight, daemonless container runtime.
```
yftacherzog: Is this required? This is one of those things that can make users u-turn before starting.

docs/mac-setup.md
```diff
@@ -0,0 +1,395 @@
+# Konflux Setup Guide for macOS
+
+This guide covers macOS-specific setup for running Konflux locally using Kind and Podman.
+
+## Table of Contents
```
yftacherzog: Can we manage the table of contents with `<!-- toc --><!-- tocstop -->` here?

docs/operator-deployment.md
```diff
@@ -0,0 +1,470 @@
...
+- [Quick Start (Local Development)](#quick-start-local-development)
+- [Production Deployment](#production-deployment)
+- [Configuration](#configuration)
+- [Authentication](#authentication)
+- [Demo Resources](#demo-resources)
+- [Troubleshooting](#troubleshooting)
+
+## Overview
+
+Konflux can be deployed on any Kubernetes cluster using the Konflux operator. The operator manages the lifecycle of all Konflux components.
```
yftacherzog, line 36: We're retiring any other approach for deploying Konflux. I think we can just have this in the main readme.

docs/operator-deployment.md
```diff
@@ -0,0 +1,470 @@
...
+
+## Quick Start (Local Development)
+
+For local development on Kind clusters (macOS or Linux).
+
+### Prerequisites
+
+- kind: `brew install kind`
+- kubectl: `brew install kubectl`
+- podman: `brew install podman` (macOS) or docker (Linux)
```
yftacherzog: The use of `brew` in multiple places suggests heavy support for mac, while in reality the maintainers don't have macs and we don't have e2e tests on mac (yet maybe), so it would be misleading to display it at the front, in my opinion.

test/resources/demo-users/dex-users.yaml
```diff
@@ -0,0 +1,47 @@
...
+#
+# Default credentials:
+# - user1@konflux.dev / password
+# - user2@konflux.dev / password
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dex
+  namespace: dex
+data:
```
gbenhaim: the dex configmap is generated by the operator. the required configurations for kind already present on this sample https://github.com/konflux-ci/konflux-ci/blob/main/operator/config/samples/konflux_v1alpha1_konflux.yaml which is also used in the ci jobs.

scripts/deploy-local-dev.sh
```diff
@@ -0,0 +1,278 @@
...
+echo ""
+echo "========================================="
+echo "Step 4: Deploying Konflux operator"
+echo "========================================="
+echo "Installing CRDs..."
+make install
+
+echo ""
+echo "Deploying operator..."
+make deploy IMG=localhost/konflux-operator:local
```
gbenhaim: 1. who is the intended audience of this script? for users who want to try out konflux I think we shouldn't ask to compile and build it, but rather point them to install a released version.

2. any script that we add here should be tested in the CI, otherwise we won't able to ensure they are in a good shape.

docs/operator-deployment.md
```diff
@@ -0,0 +1,470 @@
...
+kubectl get konflux konflux -o yaml
+
+# Check component pods
+kubectl get pods -A | grep konflux
+
+# Wait for Ready condition
+kubectl wait --for=condition=Ready konflux/konflux --timeout=15m
+```
+
+## Configuration
```
gbenhaim: we have a site with the api reference https://github.com/konflux-ci/konflux-ci/blob/main/operator/config/samples/konflux_v1alpha1_konflux.yaml it also have examples.

docs/operator-deployment.md
```diff
@@ -0,0 +1,470 @@
...
+For local development on Kind clusters (macOS or Linux).
+
+### Prerequisites
+
+- kind: `brew install kind`
+- kubectl: `brew install kubectl`
+- podman: `brew install podman` (macOS) or docker (Linux)
+- GitHub App with private key
+
+### Setup
```
gbenhaim, line 55: I think that we need to remove the code from the documentation since it's already included in scripts.

scripts/deploy-local.sh
```diff
@@ -0,0 +1,278 @@
...
+# shellcheck disable=SC1090
+source "${ENV_FILE}"
+
+# Export variables so they're available to child scripts
+export KIND_MEMORY_GB PODMAN_MACHINE_NAME REGISTRY_HOST_PORT ENABLE_REGISTRY_PORT
+export INCREASE_PODMAN_PIDS_LIMIT DEPLOY_DEMO_RESOURCES
+export GITHUB_PRIVATE_KEY_PATH GITHUB_APP_ID WEBHOOK_SECRET QUAY_TOKEN
+
+# Validate required secrets
+if [ -z "${GITHUB_PRIVATE_KEY_PATH:-}" ] || [ -z "${GITHUB_APP_ID:-}" ] || [ -z "${WEBHOOK_SECRET:-}" ]; then
```
gbenhaim: In our vision, the creation of those secrets and other admin action should be implemented in a go based CLI https://issues.redhat.com/browse/KFLUXVNGD-636

scripts/deploy-demo-resources.sh
```diff
@@ -0,0 +1,84 @@
...
+echo "👥 Deploying demo user namespaces and RBAC..."
+kubectl apply -k "${REPO_ROOT}/test/resources/demo-users/user/"
+
+# Deploy Dex configuration with demo users
+echo "🔐 Configuring Dex with demo credentials..."
+kubectl apply -f "${REPO_ROOT}/test/resources/demo-users/dex-users.yaml"
+
+# Patch Dex deployment to use the demo ConfigMap
+echo "🔧 Patching Dex deployment to use demo configmap..."
+kubectl patch deployment dex -n dex --type=json \
```
gbenhaim: this won't work since the operator will revert the change.

scripts/deploy-demo-resources.sh
```diff
@@ -0,0 +1,84 @@
...
+    echo "ERROR: Dex deployment not found in 'dex' namespace."
+    echo "Please ensure Konflux is fully deployed before adding demo resources."
+    exit 1
+fi
+
+echo "✓ Konflux detected - proceeding with demo resource deployment"
+echo ""
+
+# Deploy demo user namespaces and RBAC
+echo "👥 Deploying demo user namespaces and RBAC..."
```
gbenhaim: the operator creates a default-tenant and we would like to start using it instead of the current namespaces created for user1 and user2.

docs/mac-setup.md
arewm: Will we have tests for both podman and docker based KinD installations?

docs/operator-deployment.md
arewm: I can remove specific installation commands. We can just state the requirements and let users install them however they are needed.

docs/operator-deployment.md
arewm: What code are you suggesting to delete? The blocks below are just commands required to show how to configure the installation by using template files.

docs/operator-deployment.md
arewm: That's fine. We can link out to the sample configs.

scripts/deploy-demo-resources.sh
arewm: I don't think that the operator should create any namespaces or users by default. It should allow some namespaces/users to be created if explicitly opted in, but we don't want these configurations to be owned by the operator. They should only be used for demo purposes.

scripts/deploy-local.sh
arewm: We can discuss that long term vision later, but I am trying to help solve a problem _today_. This is a simple change that can improve the experience for deploying Konflux. It should be relatively simple to change once we have a different/better approach.

scripts/deploy-local-dev.sh
arewm: The intended audience is anyone who wants to deploy Konflux locally on a KinD cluster. You are right. We shouldn't ask them to compile and build it. I'll change that to use https://quay.io/repository/konflux-ci/konflux-operator?tab=tags.

test/resources/demo-users/dex-users.yaml
arewm: The configmap is generated by the operator but it should not be creating any default users. We cannot have an operator which auto-creates demo users that have static credentials. Even if these users do not have permissions, it is a foothold that can be exploited for an escalation of privilege attack.

CI jobs will need to be changed to explicitly enable demo users.

gbenhaim: The operator doesn't create users by default, it's the cr we use in the ci which makes it create the users.

scripts/deploy-demo-resources.sh
gbenhaim: Creation of the default tenant can be disabled using the konflux cr. I don't see any security concerns with creating a default tenant.

scripts/deploy-local.sh
gbenhaim: It's not a long term vision, the admin cli is something that will be implemented in the upcoming weeks.

docs/operator-deployment.md
gbenhaim: Almost everything. The shell code which creates secrets and also appear in the scripts, the cr definition that should exist in a sample file (that we test for correctness, variable names from the template file can exist only in the template file, duplicating them here can make the documentation out of sync.

scripts/deploy-demo-resources.sh
arewm: I am removing these changes from the PR and opened https://github.com/konflux-ci/konflux-ci/issues/4793 where we can continue to discuss the demo resources, especially the users.

scripts/deploy-local.sh
arewm: That sounds like an interesting tool. Can we not use a script-based approach for now and move towards a different approach once that is supported?

There is a lot of configuration required for deploying. I think we should be doing a reasonable simplification of it sooner if possible.

gbenhaim: @arewm you can bootstrap the cli code. We aim for a cli written in a go so we can enjoy the benefits a programming language provides.

konflux-ci/ui/core/proxy/proxy.yaml
```diff
@@ -84,7 +84,7 @@ spec:
             cpu: 10m
             memory: 64Mi
       containers:
-      - image: registry.access.redhat.com/ubi9/nginx-124@sha256:b924363ff07ee0f8fd4f680497da774ac0721722a119665998ff5b2111098ad1
+      - image: registry.access.redhat.com/ubi9/nginx-124@sha256:a75f10d71969dd2451914df6539dea28f8e2c88506f691ea211768a0bd0746d1
```
gbenhaim, line 87: why do you need to change this image?

scripts/deploy-local.env.template
```diff
@@ -0,0 +1,68 @@
...
+GITHUB_APP_ID=""
+
+# Webhook secret for GitHub webhooks
+# Generate with: openssl rand -hex 20
+WEBHOOK_SECRET=""
+
+# Quay.io robot account token (OPTIONAL - only needed for image-controller)
+# Create at: https://quay.io/ → Account Settings → Robot Accounts
+# Leave empty to skip image-controller deployment
+QUAY_TOKEN=""
```
avi-biton, line 93: I do not see QUAY_TOKEN is in use in the scripts

scripts/deploy-local.sh
```diff
@@ -0,0 +1,286 @@
+#!/bin/bash -eu
+
+# Deploy Konflux for Local Development
```
gbenhaim, line 3: @arewm can you change the e2e tests for the operator to use this script ? this will help to verify it works and to keep us safe from regressions.

konflux-ci/ui/core/proxy/proxy.yaml
arewm: The original digest is not a multi-arch one.
```
$ oras manifest fetch registry.access.redhat.com/ubi9/nginx-124@sha256:b924363ff07ee0f8fd4f680497da774ac0721722a119665998ff5b2111098ad1 | jq -e .mediaType
"application/vnd.oci.image.manifest.v1+json"
$ oras manifest fetch registry.access.redhat.com/ubi9/nginx-124@sha256:a75f10d71969dd2451914df6539dea28f8e2c88506f691ea211768a0bd0746d1 | jq -e .mediaType
"application/vnd.docker.distribution.manifest.list.v2+json"
```

scripts/deploy-local.env.template
arewm: Thanks for the catch. This was leftover from the pre-operator installation.

konflux-ci/ui/core/proxy/proxy.yaml
yftacherzog: Alright. The entire konflux-ci dir should be removed once Fedora will stop using konflux-ci/konflux-ci and we retire anything but the operator.

scripts/deploy-local.env.template
arewm: Actually, I am going to add creating the secret for the operator to the script instead of removing it from this template.

scripts/deploy-local.sh
arewm: I pushed a change to do this. It made the PR bigger though

arewm:
@gbenhaim @yftacherzog @avi-biton , this PR switched CI over to use the common deploy method and CI appears to be passing. 

I created a new CR sample for e2e which is the same as the konflux one we had before but with imagecontroller deploy added. We had talked about renaming the konflux sample to something else (due to the presence of the default password). I didn't do that, but I did add a `danger` banner to it indicating the use of default passwords.

.github/workflows/sanity-test.yml
yftacherzog, line 1: Do we still test the non-operator deployment with this change? We need to keep testing it for a while, as the Fedora cluster is taking manifests from the `konflux-ci` dir.

yftacherzog:
> @gbenhaim @yftacherzog @avi-biton , this PR switched CI over to use the common deploy method and CI appears to be passing.
> 
> I created a new CR sample for e2e which is the same as the konflux one we had before but with imagecontroller deploy added. We had talked about renaming the konflux sample to something else (due to the presence of the default password). I didn't do that, but I did add a `danger` banner to it indicating the use of default passwords.

One thing to note about the e2e CI workflows is that they're using `pull_request_target`, which means that CI is running against the workflow file in the main branch of the upstream repo rather than the PR branch.

It's a limitation that makes those changes a bit annoying and require to either test the change by opening a PR to the fork (which will require secrets created there), or opening the same PR from an upstream branch and temporarily changing the workflow to be triggered on `pull_request` instead of `pull_request_target`.

.github/workflows/sanity-test.yml
arewm: Ah, this proposed to remove. I wasn't sure what we wanted to do there which is why I put that content in a separate commit.

I'll push an update. :)

arewm: Resolving this thread. That file is no longer deleted, but a deprecation notice has been added.

arewm:
Are these checks that passed exhibiting the new workflow changes work? If so, can we merge the PR after I remove the latest commit that changes the event type?

.github/workflows/operator-test-e2e.yaml
```diff
@@ -199,48 +199,35 @@ jobs:
...
+          # Cluster configuration
+          KIND_CLUSTER: ${{ env.KIND_CLUSTER }}
+          OPERATOR_INSTALL_METHOD: "build"
+          OPERATOR_IMAGE: ${{ env.OPERATOR_IMAGE }}
 
-      - name: Build Operator Image (${{ matrix.arch }})
-        run: |
-          make docker-build IMG=$OPERATOR_IMAGE
+          # Use E2E-specific CR with image-controller for E2E tests, default CR for integration-only
+          KONFLUX_CR: ${{ matrix.test_scope == 'integration+e2e' && 'operator/config/samples/konflux-e2e.yaml' || 'operator/config/samples/konflux_v1alpha1_konflux.yaml' }}
```
yftacherzog: Can't the file be referenced by something like `matrix.konflux_cr`?

.github/workflows/sanity-test.yml
```diff
@@ -145,6 +151,8 @@ jobs:
           kubectl get namespace
 
       - name: Deploy test resources
+        env:
+          SKIP_SAMPLE_COMPONENTS: "true"
```
yftacherzog, line 155: @arewm Seeing that we change what we'er doing in sanity-test.yaml, did we move it temporarily to work on pull_request to see it's passing?

@gbenhaim @avi-biton  I'm tempted that we keep the workflows in `pull_request` rather than `pull_request_target` and then later on make sure that our commenting trigger will work for PRs from fork.

docs/operator-deployment.md
yftacherzog, line 1: It seems to me that we're kind of documenting similar things in here and in the main readme. I think we should keep it there, or reduce much of the context there (I prefer the former).

scripts/deploy-local.env.template
```diff
@@ -0,0 +1,86 @@
...
+# Leave empty to skip image-controller secret creation
+QUAY_TOKEN=""
+QUAY_ORGANIZATION=""
+
+# Smee webhook forwarding (OPTIONAL - for webhook testing)
+# Smee forwards GitHub webhooks to non-publicly-accessible clusters
+# Create a channel at: https://smee.io/
+# If not set, a random channel will be generated automatically
+# See docs for configuring your GitHub App to use this webhook URL
+SMEE_CHANNEL=""
```
yftacherzog, line 103: Can we add an example above to make it clear whether we're expecting just the channel ID or the full URL?

README.md
```diff
@@ -75,9 +79,78 @@ forwarding is needed for accessing Konflux.
 the remote machine to port `9443` on your local machine to be able to access the UI from
 your local machine.
 
+## Operator-Based Deployment
+
+Konflux can be deployed using the operator-based installer, which supports both local development (Kind) and production deployments on any Kubernetes cluster.
+
+### Local Development (Kind)
```
yftacherzog, line 86: When initially working on this document, we added the use of the gear icon to make it a little easier to track the steps that require action rather than just providing information.

We should stick to it.

README.md
```diff
@@ -75,9 +79,78 @@ forwarding is needed for accessing Konflux.
...
+```bash
+# 1. Create configuration from templates
+cp scripts/deploy-local.env.template scripts/deploy-local.env
+
+# 2. Edit scripts/deploy-local.env with your GitHub App credentials
+# See the template file for instructions
+
+# 3. Deploy Konflux
+./scripts/deploy-local.sh
+```
```
yftacherzog, line 103: If we split the code blocks to include batches that can be run together, we allow users to use the little button at the top-right corner of the box that copies the entire content.

I.e. in this case, they could have clicked it once to copy the file, and then click it another time to run the script, rather than highlighting different lines.

README.md
```diff
@@ -75,9 +79,78 @@ forwarding is needed for accessing Konflux.
...
+
+# 3. Create required secrets (GitHub App, Quay tokens, etc.)
+# See deployment guide for details
+```
+
+**Important:** Do not use the sample with demo users (`konflux_v1alpha1_konflux.yaml`) for production - configure OIDC authentication instead.
+
+See [Operator Deployment Guide](docs/operator-deployment.md) for complete instructions and [sample configurations](operator/config/samples/).
+
+### Key Differences from Legacy Deployment
```
yftacherzog, line 136: I think we should drop the legacy deployment from the docs altogether.

docs/troubleshooting.md
```diff
@@ -179,7 +182,49 @@ The symptoms may include:
   ```
 * Pipelines fail at inconsistent stages.
 
-:gear: For mitigation steps, consult the notes at the top of the
+### For Podman Users (macOS/Linux)
+
+**Check Podman machine memory:**
+
+```bash
+podman machine inspect | grep Memory
```
yftacherzog, line 193: On Linux, this gives:
```bash
→ podman machine inspect | grep Memory
Error: podman-machine-default: VM does not exist
```

docs/troubleshooting.md
```diff
@@ -179,7 +182,49 @@ The symptoms may include:
...
+```
+
+**If insufficient, create new machine with more resources:**
```
yftacherzog, line 196: Not sure if `podman machine` is relevant for typical linux users. Might need to move things under different titles to make it clear what's relevant to only some of the platforms.

operator/config/samples/README.md
```diff
@@ -2,10 +2,24 @@
...
+- [Production Considerations](#production-considerations)
+  * [Authentication](#authentication)
+  * [Default Tenant](#default-tenant)
+- [Related Documentation](#related-documentation)
+
+<!-- tocstop -->
+
+## Functional Samples
+
+**konflux_v1alpha1_konflux.yaml** - Used in CI tests and local development. Represents a complete Konflux CR configuration with all components, realistic resource limits, and demo users for testing. Includes helpful comments for common configurations.
```
yftacherzog: Let's wrap the lines shorter.

scripts/deploy-local.sh
yftacherzog, line 1: I tried it locally now, and it timed out for legitimate reasons (took too long for some things to be ready).
I wanted to see whether I could run the script again to verify that it's reentrant, but then I realized that it deletes the cluster automatically.

I think we should avoid this -- at least as default behavior.
It's typical for slower environments to time out, and it would be best if users can run the script again and it won't replace things.

CONTRIBUTING.md
```diff
@@ -50,6 +53,58 @@ It may be also recommended to create a configuration file. To do so please check
...
+OPERATOR_INSTALL_METHOD=build ./scripts/deploy-local.sh
+```
+
+This will:
+1. Build the operator image from your local changes
+2. Load the image into the Kind cluster
+3. Install CRDs using `make install`
+4. Deploy the operator using `make deploy`
+
+## Operator Development Commands
```
yftacherzog: The operator dir has a readme that shares some of the content we have here. Might be better if we point there instead.

yftacherzog: (and enhance the doc there if needed)

scripts/deploy-local.env.template
arewm: I'll push this with the next commit.

README.md
arewm: I think we should keep these until legacy deployments are not possible, but we can restructure the documentation in a future commit to make them less obvious.

docs/operator-deployment.md
arewm: I think we should adjust the documentation in a future PR which includes removing the legacy documentation. When we do that, I was thinking about putting a high-level overview in README and the smaller details/configuration in the separate documentation. But regardless, we can have that conversation with a future PR as needed.

.github/workflows/sanity-test.yml
arewm: This is set to 

```
on:
  pull_request_target:
[...]
    - name: Check prerequisites
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        PR_AUTHOR: ${{ github.event.pull_request.user.login }}
```

yftacherzog: Not sure what you meant here in this comment. :)

scripts/deploy-local.env.template
```diff
@@ -0,0 +1,88 @@
...
+# Operator installation method:
+#   local (default) - Install from current commit using kustomize
+#   build           - Build operator image locally (for operator development)
+#   release         - Install from latest GitHub release
+OPERATOR_INSTALL_METHOD=local
+
+# Operator image (used with 'local' and 'build' methods)
+# Default: quay.io/konflux-ci/konflux-operator:latest
+# Override to use a different image with 'local' method
+OPERATOR_IMAGE=""
```
yftacherzog, line 28: I have some reservations about the default combination. I think it's neither something that is promised to be a working combination nor it's something that is completely dev-oriented as it's using the latest released image.

I think a better default would be release manifests + released image.

.github/workflows/operator-test-e2e.yaml
```diff
@@ -145,6 +146,7 @@ jobs:
             image_suffix: "-arm64"
             cluster_suffix: "-arm"
             test_scope: "integration"
+            konflux_cr: "operator/config/samples/konflux_v1alpha1_konflux.yaml"
```
gbenhaim: can we use the same CR for both arm and amd?

.github/workflows/operator-test-e2e.yaml
```diff
@@ -235,48 +239,34 @@ jobs:
...
+          # GitHub App secrets (standardized names)
+          GITHUB_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
+          GITHUB_APP_ID: ${{ secrets.APP_ID }}
+          WEBHOOK_SECRET: ${{ secrets.APP_WEBHOOK_SECRET }}
 
-      - name: Deploy Operator Controller
-        run: |
-          make deploy IMG=$OPERATOR_IMAGE
+          # Quay secrets for E2E tests (standardized names)
+          QUAY_TOKEN: ${{ matrix.test_scope == 'integration+e2e' && secrets.QUAY_TOKEN || '' }}
```
gbenhaim: why do did you add " || '' "? if the secret is missing the action should fail

scripts/deploy-local.sh
```diff
@@ -0,0 +1,351 @@
...
+echo ""
+echo "========================================="
+echo "Step 3: Deploying Konflux operator"
+echo "========================================="
+
+INSTALL_METHOD="${OPERATOR_INSTALL_METHOD:-local}"
+echo "Using installation method: ${INSTALL_METHOD}"
+
+case "${INSTALL_METHOD}" in
+    local)
```
gbenhaim, line 158: can we use make here? the commands here duplicates what we have in the makefile.

scripts/deploy-local.sh
```diff
@@ -0,0 +1,351 @@
...
+fi
+
+# Step 7: Wait for Konflux to be ready
+echo ""
+echo "========================================="
+echo "Step 7: Waiting for Konflux to be ready"
+echo "========================================="
+echo "This may take several minutes..."
+
+if ! kubectl wait --for=jsonpath='{.status.conditions[?(@.type=="Ready")].status}'=True \
```
gbenhaim: this can be simplified `kubectl wait --for=condition=Ready=True konflux konflux`

scripts/deploy-local.sh
```diff
@@ -0,0 +1,351 @@
...
+    konflux/konflux \
+    --timeout=15m 2>/dev/null; then
+    echo ""
+    echo "WARNING: Konflux CR did not become Ready within 15 minutes"
+    echo "         This may be normal if deploying all components"
+    echo "         Check status with: kubectl get konflux konflux -o yaml"
+    echo ""
+    echo "To monitor progress:"
+    echo "  kubectl get pods -A"
+    echo "  kubectl get konflux konflux -o jsonpath='{.status.conditions}' | jq"
```
gbenhaim: we don't have a  dependency on jq today. is it really needed?

operator/config/samples/README.md
```diff
@@ -2,10 +2,27 @@
...
+## Functional Samples
+
+**konflux_v1alpha1_konflux.yaml** - Used in CI tests and local development.
+Represents a complete Konflux CR configuration with all components, realistic
+resource limits, and demo users for testing. Includes helpful comments for
+common configurations.
+
+**konflux-empty-cr.yaml** - Minimal empty spec using all default values.
+
+**konflux-with-github-auth.yaml** - Shows GitHub authentication connector configuration for production use.
```
avi-biton, line 29: We might want to add the `konflux-e2e.yaml` description as well as it not a part of the other sample files that are provided to demonstrate the CRD structure and available
fields.

.github/workflows/operator-test-e2e.yaml
arewm: Removed.

.github/workflows/operator-test-e2e.yaml
arewm: The amd64 job runs E2E tests which require image-controller; the arm64 job only runs integration tests which don't. `konflux-e2e.yaml` is identical to `konflux_v1alpha1_konflux.yaml` except it adds

``` yaml
imageController:
    enabled: true.
```

We could use `konflux-e2e.yaml` for both as the image-controller being enabled shouldn't break integration-only tests. It just deploys an unnecessary component on arm64. Do you want that simplification?

scripts/deploy-local.sh
arewm: Sounds good. I'll also make similar changes elsewhere (i.e. the GitHub workflows).

scripts/deploy-local.sh
arewm: This isn't a dependency as it is just a suggested command that users might want to run, but we can remove it regardless.

scripts/deploy-local.env.template
arewm: We can change the default to the release configuration. The use case that I had though of for the local is if a user has cloned it locally and wants to modify the CR before applying them (the alternative would be to use release and either modify the CR after applying or modifying it locally somehow and re-applying it). We can caution users about this mismatch for local and instead suggest that they run the tool from a specific tag?

arewm: I am pushing this change so I will resolve the comment.

docs/troubleshooting.md
arewm: Resolving this comment. This is in the macOS section so it wouldn't be relevant for Linux.

.github/workflows/sanity-test.yml
arewm: I can change sanity-test to use a pull request if you want.

README.md
arewm: I am also adding this convention to CONTRIBUTING.md.

.github/workflows/sanity-test.yml
yftacherzog: Yes. Please do.

README.md
yftacherzog: It's mentioned at the top of the readme if I recall correctly.

.github/workflows/sanity-test.yml
yftacherzog: We'll merge it with `pull_request` trigger.
We'll start creating PRs from the upstream repo, and the mechanism we have for allowing walk-ins to open PRs from forks should still allow triggering those workflows with PR comments by maintainers.

test/resources/demo-users/user/kustomization.yaml
```diff
@@ -3,4 +3,6 @@ kind: Kustomization
 resources:
   - ns1
   - ns2
+  - managed-ns1
+  - managed-ns2
```
yftacherzog: I'm not sure that those should be included here. The main reason is that the managed resources (mainly the RPA I think) need to be modified by the user. 

That's the reason they were left to be added as part of the onboarding process.

CONTRIBUTING.md
```diff
@@ -15,6 +17,30 @@ Contributing Guidelines
 
 <!-- tocstop -->
 
+# Documentation Conventions
+
+When writing user-facing documentation (README, deployment guides, etc.):
+
+- **Gear icon (:gear:)**: Use `:gear:` to mark steps where the user must take action. This makes action items visually distinct from explanatory text.
+- **One command per code block**: Split bash code blocks so each contains a single copyable command. This allows users to use the copy button on each block rather than manually selecting lines.
+- **Descriptive text between blocks**: Add a brief description before each code block explaining what the command does.
```
yftacherzog, line 26: Let's wrap the lines shorter.

test/resources/demo-users/user/kustomization.yaml
arewm: The RPA can be modified by a user if they want or the user can just create a new RPA. 

I added this in here to simplify the deployment steps that are required for CI, but I think that it also makes sense to include it. Part of the value of Konflux comes from the concept of having specific user/managed namespaces. If we auto-create these namespaces when we deploy the demo users then we can easily refer to them in any documentation/description including _how_ these can be utilized for improving the supply chain security.

.github/workflows/operator-test-e2e.yaml
gbenhaim: Yes please. We want to run e2e tests on arm as well. The reason we still didn't do it yet is because we are waiting for a fix from another team.

.github/workflows/operator-test-e2e.yaml
gbenhaim: I still see the || when reading the quay token.

gbenhaim: Also the smell channel below

test/resources/demo-users/user/kustomization.yaml
yftacherzog: If i recall correctly,  the e2e tests are using their own rpa. I don't think the managed resources are used in ci.

scripts/setup-kind-local-cluster.sh
```diff
@@ -0,0 +1,221 @@
...
+ENABLE_REGISTRY_PORT="${ENABLE_REGISTRY_PORT:-1}"
+REGISTRY_HOST_PORT="${REGISTRY_HOST_PORT:-5001}"
+
+# Increase inotify limits on Linux
+if [[ "$(uname)" == "Linux" ]]; then
+    echo "Increasing inotify limits for Kind cluster..."
+    echo "You may be prompted for your password."
+    sudo sysctl fs.inotify.max_user_watches=524288
+    sudo sysctl fs.inotify.max_user_instances=512
+fi
```
yftacherzog: Found myself waiting for the script to finish on rerun and then realized it's waiting for sudo password :)

```suggestion
# Increase inotify limits on Linux (only if current values are lower than required)
if [[ "$(uname)" == "Linux" ]]; then
    WATCHES_REQUIRED=524288
    INSTANCES_REQUIRED=512
    WATCHES_CURRENT=$(cat /proc/sys/fs/inotify/max_user_watches 2>/dev/null || echo 0)
    INSTANCES_CURRENT=$(cat /proc/sys/fs/inotify/max_user_instances 2>/dev/null || echo 0)
    if [[ "$WATCHES_CURRENT" -lt "$WATCHES_REQUIRED" ]] || [[ "$INSTANCES_CURRENT" -lt "$INSTANCES_REQUIRED" ]]; then
        echo "Increasing inotify limits for Kind cluster..."
        echo "You may be prompted for your password."
        sudo sysctl fs.inotify.max_user_watches="$WATCHES_REQUIRED"
        sudo sysctl fs.inotify.max_user_instances="$INSTANCES_REQUIRED"
    else
        echo "inotify limits already sufficient (watches=$WATCHES_CURRENT, instances=$INSTANCES_CURRENT)."
    fi
fi
```

scripts/deploy-local.sh
```diff
@@ -0,0 +1,348 @@
...
+OPERATOR_INSTALL_METHOD="${OPERATOR_INSTALL_METHOD:-release}"
+OPERATOR_IMAGE="${OPERATOR_IMAGE:-quay.io/konflux-ci/konflux-operator:latest}"
+
+# Export variables for child scripts
+export KIND_CLUSTER KIND_MEMORY_GB PODMAN_MACHINE_NAME REGISTRY_HOST_PORT ENABLE_REGISTRY_PORT
+export INCREASE_PODMAN_PIDS_LIMIT
+export GITHUB_PRIVATE_KEY GITHUB_APP_ID WEBHOOK_SECRET QUAY_TOKEN QUAY_ORGANIZATION
+
+# Get Konflux CR file path (env var, command-line arg, or default)
+KONFLUX_CR="${KONFLUX_CR:-${1:-${REPO_ROOT}/operator/config/samples/konflux_v1alpha1_konflux.yaml}}"
```
yftacherzog, line 96: We should have `KONFLUX_CR` in the env file template. Otherwise, it includes the image controller secrets, but does not suggest that image controller is not going to be deployed (while the script is going to wait on it) unless it's enabled in the CR.

yftacherzog: I noticed now that the script allows it as input, so I think we should mention it as an optional input in the readme when mentioning the script, and add to the comment of the image-controller parameters that it needs to be enabled in the CR.

scripts/deploy-local.env.template
yftacherzog: I think it's documented incorrectly now. It should be the full URL as far as I understand the way the script is implemented.

scripts/setup-kind-local-cluster.sh
arewm: I pushed a change for this providing text if they want to run the commands instead of letting the script have sudo access

scripts/deploy-local.sh
arewm: I added this but also added text only in the template file about the inter-relationship.

yftacherzog (approved):
@gbenhaim @arewm I think I'm good with merging this :)

scripts/deploy-local.env.template
arewm: Bah. Fixed. I thought I had it implemented the other way.

gbenhaim (approved):

PR #4749, opened 2026-01-26
operator: Add merge-based pipelineConfig for build pipeline customization

arewm:
## Summary

Add a `pipelineConfig` field to `KonfluxBuildServiceSpec` that gives users
fine-grained control over the `build-pipeline-config` ConfigMap via merge
semantics, rather than an all-or-nothing boolean. Users can:

- Override individual default pipelines by specifying the same name with a different bundle
- Remove specific defaults with `removed: true`
- Discard all defaults with `removeDefaults: true`
- Add custom pipelines alongside defaults

The operator always applies the ConfigMap. When `pipelineConfig` is unset,
defaults pass through unchanged.

Example CR:
```yaml
spec:
  pipelineConfig:
    pipelines:
      - name: fbc-builder
        removed: true
      - name: my-custom
        bundle: quay.io/myorg/pipeline:latest
```

## Test plan

- [x] Unit tests for `mergePipelines` and `applyPipelineConfigMerge` covering all merge scenarios
- [x] Integration tests verifying ConfigMap content after reconciliation
- [x] CRD validation: `MinLength=1` on pipeline name, CEL rule preventing `bundle` + `removed` together
- [x] `make test` passes (all suites)


arewm:
@gbenhaim Thank you for this suggestion - I agree that the current boolean approach is limited.

Your merge-based design addresses the key limitations you identified. Before implementing it, I want to clarify a few design details given the evolving pipeline configuration.

## ConfigMap Structure Evolution

With the new component model, the `build-pipeline-config` ConfigMap may support more than just bundle resolvers (e.g., git references as well). Because of this potential complexity, I think your point #4 would be clearer with an explicit field rather than relying on an empty bundle value.

**Proposed:**
```yaml
spec:
  pipelineConfig:
    pipelines:
      - name: fbc-builder
        removed: true  # or "excluded: true"
      - name: my-custom
        bundle: quay.io/myorg/pipeline:latest
```

This works regardless of resolver type and makes the intent explicit.

## Removing All Defaults

Your suggestion also handles adding/overriding individual pipelines well, but doesn't address the case where users want zero defaults. I think we should add a top-level configuration to disable all defaults:

```yaml
spec:
  pipelineConfig:
    removeDefaults: true  # Disable all operator-provided pipelines
    pipelines:
      - name: my-custom
        bundle: quay.io/myorg/pipeline:latest
```

This is important to protect against unvetted pipelines being added (i.e. in future operator upgrades) without explicit approval from the installer/platform engineer.

Does this sound reasonable? If so, I can ask Claude to implement this approach.

gbenhaim:
> @gbenhaim Thank you for this suggestion - I agree that the current boolean approach is limited.
> 
> Your merge-based design addresses the key limitations you identified. Before implementing it, I want to clarify a few design details given the evolving pipeline configuration.
> 
> ## ConfigMap Structure Evolution
> With the new component model, the `build-pipeline-config` ConfigMap may support more than just bundle resolvers (e.g., git references as well). Because of this potential complexity, I think your point #4 would be clearer with an explicit field rather than relying on an empty bundle value.
> 
> **Proposed:**
> 
> ```yaml
> spec:
>   pipelineConfig:
>     pipelines:
>       - name: fbc-builder
>         removed: true  # or "excluded: true"
>       - name: my-custom
>         bundle: quay.io/myorg/pipeline:latest
> ```
> 
> This works regardless of resolver type and makes the intent explicit.
> 
> ## Removing All Defaults
> Your suggestion also handles adding/overriding individual pipelines well, but doesn't address the case where users want zero defaults. I think we should add a top-level configuration to disable all defaults:
> 
> ```yaml
> spec:
>   pipelineConfig:
>     removeDefaults: true  # Disable all operator-provided pipelines
>     pipelines:
>       - name: my-custom
>         bundle: quay.io/myorg/pipeline:latest
> ```
> 
> This is important to protect against unvetted pipelines being added (i.e. in future operator upgrades) without explicit approval from the installer/platform engineer.
> 
> Does this sound reasonable? If so, I can ask Claude to implement this approach.

Looks good!

yftacherzog:
Hi @arewm, I see this one had no activity for a while. Is it still relevant?

yftacherzog:
Closing for now. You can open later if this becomes relevant again.

arewm:
/review
Thanks for closing this -- I had forgotten about it. I wonder, is it possible for fullsend to review this even though the code is old and some conflicts might exist?

PR #4776, opened 2026-01-27
fix(KFLUXVNGD-694): Publish operator workflow isn't triggerred

avi-biton:
The `create-release.yaml` workflow was using the default `GITHUB_TOKEN` to create releases.
GitHub Actions prevents events triggered by `GITHUB_TOKEN` from triggering other workflows to avoid infinite loops. 
This prevented the `community-operator-pr.yaml` workflow (which listens for `release` events) from running automatically. 

This commit updates the workflow to use `KONFLUX_CI_BOT_PAT`, which will enable triggering the downstream community operator PR workflow.

Assisted-By: Cursor

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -119,7 +119,7 @@ jobs:
       - name: Create GitHub Release
         working-directory: ${{ github.workspace }}
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.KONFLUX_CI_BOT_PAT }}
         run: |
           .github/scripts/create-release.sh \
             "${{ steps.version.outputs.version }}" \


.github/workflows/create-release.yaml
```diff
@@ -119,7 +119,7 @@ jobs:
       - name: Create GitHub Release
         working-directory: ${{ github.workspace }}
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.KONFLUX_CI_BOT_PAT }}
```
yftacherzog, line 122: In other cases, we had to generate the token on the fly from a private key, but this might be a slightly different kind of secret.

avi-biton: We are using the same secret in https://github.com/konflux-ci/konflux-ci/blob/main/.github/workflows/community-operator-pr.yaml#L36 which should be triggered by the `create-release` workflow

yftacherzog: ok. I'm not sure to be honest but let's give it a try. [Here](https://github.com/konflux-ci/konflux-ci/blob/main/.github/workflows/update-upstream-manifests.yaml#L34-L35) we use a secret to generate a temp token, but it might be that what we have in this case is something which is more similar to a personal PAT.

PR #5222, opened 2026-02-12
feat: add upstream changelog generator for release notes (POC)

ergonzal22:
## Summary

- Adds a curated "Upstream Changes" section to release notes showing only `feat`/`fix` conventional commits from upstream component repos
- Covers **4 components**: release-service, UI (via `newTag`), application-api, and enterprise-contract (via `?ref=` URLs)
- GitHub's `--generate-notes` auto-generated changelog is kept as-is — the curated section is additive

### How it works

Between operator release tags, upstream commit SHAs change in the kustomization files under `operator/upstream-kustomizations/` (via Renovate digest updates). The new `generate-changelog.sh`:

1. Extracts upstream SHAs at each tag using `yq` — either from `images[].newTag` or from `resources[]` `?ref=` URLs, depending on the component
2. Queries the GitHub compare API for commits between those SHAs
3. Filters to `feat`/`fix` conventional commits, excluding bot authors
4. Outputs formatted markdown to stdout

### Component configuration

Components are configured in a single associative array with 5 pipe-delimited fields:

```bash
declare -A COMPONENT_CONFIG=(
  ["release-service"]="...kustomization.yaml|quay.io/konflux-ci/release-service|konflux-ci/release-service|release-service|newTag"
  ["ui"]="...kustomization.yaml|quay.io/konflux-ci/konflux-ui|konflux-ci/konflux-ui|UI|newTag"
  ["application-api"]="...kustomization.yaml|redhat-appstudio/application-api|konflux-ci/application-api|Application API|ref"
  ["enterprise-contract"]="...kustomization.yaml|conforma/crds|conforma/crds|Enterprise Contract|ref"
)
```

Adding a new component is a single line. The `extraction_type` field (`newTag` or `ref`) dispatches to the appropriate extraction function.

### Files changed

| File | Change | Why |
|------|--------|-----|
| `.github/scripts/generate-changelog.sh` | **New** | Core changelog logic. Two extraction methods (`newTag` via `yq` for images, `ref` via `yq`+`sed` for resource URLs). Handles new/removed component edge cases. Deterministic output order. |
| `.github/scripts/prepare-release-notes.sh` | Modified | Calls `generate-changelog.sh` with `$GIT_REF`. Uses `git tag --merged "$GIT_REF"` with strict semver regex for previous tag lookup — scopes to branch ancestry and excludes pre-release tags (`-rc`, `-beta`, etc.). |
| `.github/workflows/create-release.yaml` | Modified | Adds `GH_TOKEN` env var to the "Prepare release notes" step so `gh api` can authenticate. |
| `.github/workflows/test-changelog.yaml` | **New** | CI workflow with 8 tests run on PRs touching these scripts. Check the job summary to see live results. |

### Example output (v0.0.8 → v0.0.9)

```markdown
## Upstream Changes

### release-service
- *feat*: add componentGroup support for rp and rpa
```

### Edge cases handled

- **New component** (didn't exist in old tag): Outputs `- *new*: component added in this release`
- **Removed component** (existed in old tag but not new): Outputs `- *removed*: component removed in this release`
- **Same SHA**: Component silently skipped (no changes)
- **Missing kustomization file**: Warning logged to stderr, component skipped
- **Unknown extraction type**: Error logged, component skipped
- **Multiple release branches**: `git tag --merged "$GIT_REF"` scopes to branch ancestry; strict `^v[0-9]+\.[0-9]+\.[0-9]+$` regex excludes RC/pre-release tags
- **yq returning `"null"`**: Filtered to empty string
- **API truncation**: Warning logged when GitHub's 250-commit compare limit is hit

### Risk assessment

- **Zero risk to existing releases**: The changelog call is wrapped in `|| true`. If anything goes wrong, releases proceed exactly as before.
- **No behavior change to existing workflow**: The only workflow change is adding `GH_TOKEN` to one step. `--generate-notes` remains enabled.
- **Additive only**: The curated changelog is appended after the existing static template. Nothing is removed or reordered.

## Test plan

- [x] Tested in `ubuntu:latest` container with authenticated GitHub API (all 8 CI tests + 3 additional local tests pass)
- [x] Verified against real tags: `v0.0.8 → v0.0.9`, `v0.0.6 → v0.0.9`
- [x] Both extraction types: `newTag` (release-service, UI) and `ref` (application-api, enterprise-contract)
- [x] UI-specific test with real changelog output (v0.0.6 → v0.0.7, newTag actually changed)
- [x] Enterprise Contract new-component detection for ref-type extraction
- [x] Edge cases: same tag, new component (first tag), unknown component, deterministic ordering, yq null handling
- [x] Full integration via `prepare-release-notes.sh` with `$GIT_REF`-based previous tag lookup
- [ ] Check the **"Demo Changelog Output"** CI job summary on this PR for live results

Ref: KFLUXVNGD-726

### DIFF
--- a/.github/scripts/prepare-release-notes.sh
+++ b/.github/scripts/prepare-release-notes.sh
@@ -30,13 +30,15 @@ OUTPUT_FILE="$4"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
-# get_previous_tag finds the nearest tag reachable from the current commit.
-# git describe walks the commit graph backwards and only finds tags that are
-# ancestors of the current ref. Unlike --sort=-creatordate, this respects
-# branch ancestry, so it works correctly when releasing from multiple branches
-# (e.g., release-0.x and release-1.0 simultaneously).
+# get_previous_tag finds the most recent stable release tag reachable from
+# $GIT_REF. Uses --merged to scope to ancestor tags (respects branch topology),
+# --sort=-creatordate to order by git history, and a strict regex to match only
+# stable semver tags (vX.Y.Z) — excluding pre-release suffixes like -rc, -beta,
+# -alpha, etc.
 get_previous_tag() {
-  git describe --tags --abbrev=0 2>/dev/null || true
+  git tag --merged "$GIT_REF" --sort=-creatordate 2>/dev/null \
+    | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' \
+    | head -1 || true
 }
 
 # Generate release notes
--- a/.github/workflows/test-changelog.yaml
+++ b/.github/workflows/test-changelog.yaml
@@ -83,7 +83,8 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          NEXT_VERSION="v0.0.$(( $(git tag -l 'v0.0.*' | sed 's/v0\.0\.//' | sort -n | tail -n 1) + 1 ))"
+          LATEST_TAG=$(git tag --merged HEAD --sort=-creatordate | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1)
+          NEXT_VERSION=$(echo "$LATEST_TAG" | awk -F. '{printf "%s.%s.%d", $1, $2, $3+1}')
           echo "### Test 4: Full integration (\`prepare-release-notes.sh ${NEXT_VERSION}\`)" >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "Simulates a release of \`${NEXT_VERSION}\` from HEAD. This is the content written to the release notes file by our scripts (static template + curated upstream changes). In a real release, GitHub's auto-generated commit history would also be appended below by the \`--generate-notes\` flag." >> "$GITHUB_STEP_SUMMARY"
@@ -126,11 +127,12 @@ jobs:
           echo "</details>" >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
 
-          # Test 6: All components (both extraction types)
-          echo "### Test 6: \`v0.0.8\` → \`v0.0.9\` (all components)" >> "$GITHUB_STEP_SUMMARY"
+          # Test 6: New component edge case (first tag, component may not exist)
+          FIRST_TAG=$(git tag --sort=creatordate | head -n 1)
+          echo "### Test 6: \`${FIRST_TAG}\` → \`v0.0.9\` (edge case: UI may not exist in old tag)" >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo '```markdown' >> "$GITHUB_STEP_SUMMARY"
-          .github/scripts/generate-changelog.sh v0.0.8 v0.0.9 2>/tmp/stderr6.log >> "$GITHUB_STEP_SUMMARY" || true
+          .github/scripts/generate-changelog.sh "$FIRST_TAG" v0.0.9 ui 2>/tmp/stderr6.log >> "$GITHUB_STEP_SUMMARY" || true
           echo '```' >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "<details><summary>Diagnostics</summary>" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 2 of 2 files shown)


.github/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,197 @@
...
+
+OLD_TAG="$1"
+NEW_TAG="$2"
+shift 2
+
+# Component configuration
+# Format: "kustomization_path|github_org/repo|display_name"
+# Add new components here to extend changelog coverage.
+declare -A COMPONENT_CONFIG=(
+  ["release-service"]="operator/upstream-kustomizations/release/core/kustomization.yaml|konflux-ci/release-service|release-service"
```
yftacherzog: Not all our upstream sources have references to git.

For example, UI doesn't have a reference to manifests in git, and the only thing we have for it is the image. In that case, I suppose we will need to correlate the images that were used in previous release and this release in order to find the git commit.

So, we'd need to decide whether we can have a unified approach that will work for all (maybe the image approach can be used for all -- need to verify).

Regardless, if we need image-based approach for some or all, we might be able to make our lives easier by verifying that we're using git sha references for all images (I think we should be able to do that). If so, then we won't need to actually fetch any image, but we would need to have some sort of mapping between images and git repos.

ergonzal22: I changed the code to use the `newTag` in the kustomization files and added the UI component to the list. You can see the changes in the new commit and in here: https://github.com/konflux-ci/konflux-ci/actions/runs/21950224000

yftacherzog: Cool. Looking!

.github/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,211 @@
...
+  local content
+  content=$(git show "${ref}:${kustomization_path}" 2>/dev/null) || {
+    echo "" # Return empty on failure
+    return
+  }
+
+  # Find the image entry by name, then extract the newTag value.
+  # Uses -F for fixed-string matching (image names contain dots).
+  # -A3 captures the newTag line which follows name/newName.
+  echo "$content" | grep -F -A3 "$image_name" | grep -oP 'newTag:\s*\K[a-f0-9]+' | head -1 || true
```
yftacherzog: I think we can assume the runner has yq installed, which can make this more human-readable.

.github/scripts/prepare-release-notes.sh
```diff
@@ -51,4 +63,18 @@ kubectl apply -f https://github.com/konflux-ci/konflux-ci/releases/download/${VE
 - [README.md](https://github.com/konflux-ci/konflux-ci/blob/main/README.md) - Installation and usage instructions
 EOF
 
+# Append upstream changelog (failures here must never block the release)
+PREVIOUS_TAG=$(get_previous_tag "$VERSION")
+if [ -n "$PREVIOUS_TAG" ]; then
+  # Use HEAD (not $VERSION) as the new ref because the version tag has not been
```
yftacherzog: Should we use HEAD or $GIT_REF?
If I'm not mistaken HEAD might still be pointing to the top of main at this point.

.github/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,211 @@
...
+#
+# Args: $1=ref, $2=kustomization_path, $3=image_name
+extract_sha_from_kustomization() {
+  local ref="$1"
+  local kustomization_path="$2"
+  local image_name="$3"
+
+  local content
+  content=$(git show "${ref}:${kustomization_path}" 2>/dev/null) || {
+    echo "" # Return empty on failure
```
gbenhaim, line 85: Why not fail in case of an error?

.github/scripts/generate-changelog.sh
gbenhaim: We have components without images, for example application-api and enterprise contract.

.github/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,211 @@
...
+  IFS='|' read -r kustomization_path image_name github_repo display_name <<< "$config"
+
+  echo "Processing component: ${display_name} (${github_repo})" >&2
+
+  # Extract SHAs at each tag
+  local old_sha new_sha
+  old_sha=$(extract_sha_from_kustomization "$OLD_TAG" "$kustomization_path" "$image_name")
+  new_sha=$(extract_sha_from_kustomization "$NEW_TAG" "$kustomization_path" "$image_name")
+
+  if [ -z "$old_sha" ]; then
```
gbenhaim, line 188: We have an edge case where a new component was added or removed after the previous release.

.github/scripts/prepare-release-notes.sh
```diff
@@ -28,6 +28,18 @@ IMAGE_TAG="$2"
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# get_previous_tag finds the most recent existing tag by creation date.
+# The current release's tag does not exist yet (it is created by the "Create
+# GitHub Release" step later in the workflow), so the latest existing tag is
+# the previous release. We sort by creation date rather than version number
+# because the repo has legacy tags (v0.1-v0.4) that sort higher than the
+# current v0.0.x series in version comparison.
+get_previous_tag() {
+  git tag --sort=-creatordate | head -1
```
gbenhaim: Will this work if we do release from multiple branches?

.github/scripts/generate-changelog.sh
ergonzal22: Done! Changed using yq

.github/scripts/prepare-release-notes.sh
ergonzal22: Changed to $GIT_REF

.github/scripts/prepare-release-notes.sh
yftacherzog: @ernesgonzalez33 we're now working on the ability to maintain multiple releases at the same time.

Development is going to be done in main. It's going to be tagged with release-candidate tags (I'm proposing vX.Y.0-rc.W, when W is incremented on each release, but it's not settled yet).

Once we make a release, we're going to create branch release-X.Y and there will get tags such as the one we've been having on main up until now (vX.Y.0, vX.Y.1...).

When creating the release notes, we should start with including what's new since the last stable release. Examples:

1. v0.2.0-rc.2: what changed since v0.1.0
2. v0.2.0: what changed since v0.1.0
3. v0.2.1: what changed since v0.2.0

Later on, but can be out of scope for this time, for RCs, we might also want to include what changed since the last RC.

All in all, need to understand how we identify which tag are we looking for as the starting point to be covered in the release notes, and need to see if this can be inferred from the things the release workflow has access to at the moment.

ergonzal22: @yftacherzog so basically getting the last `semver` version in the git history? I'll push a change with that for review. Also, we could filter `-rc` from tags. But I don't like that because maybe there are non stable tags that have a different appendix

.github/scripts/generate-changelog.sh
ergonzal22: I added support for that. It will show up in the changelog when there are new components or deleted others. Should I add a `BREAKING CHANGE` note when removing something?

.github/scripts/generate-changelog.sh
ergonzal22: Addressed. Added a second extraction type (`ref`) alongside the existing `newTag` type. The component config now has a 5th field (`extraction_type`) that dispatches to the appropriate function:

- `newTag`: Uses `yq` to extract `images[].newTag` (release-service, UI)
- `ref`: Uses `yq` to find matching `resources[]` URLs, then `sed -n` to extract the `?ref=` SHA (application-api, enterprise-contract)

Both application-api (`konflux-ci/application-api`) and enterprise-contract (`conforma/crds`) are now configured and tested.

I'm open to making this simpler if we found a way or adding it to the logic instead of selecting the option in the component config.

.github/scripts/prepare-release-notes.sh
ergonzal22: updated

yftacherzog: > so basically getting the last semver version in the git history

Not necessarily. When doing x.y.0, you need to compute it from x.y-1.0 if y is not 0.
For y=0, I don't know what's the expected outcome yet, but I think it's going to be a while before we run into it and test it, so we can put some sort of placeholder for now.

PR #5256, opened 2026-02-12
fix(operator): use explicit local path for go build and tidy go.sum

gbenhaim:
Use ./cmd/main.go instead of cmd/main.go in the Containerfile and Makefile go build commands. Without the ./ prefix, Go can ambiguously interpret cmd/main.go as a standard library package path rather than a local file path, causing "package cmd/main.go is not in std" build failures in CI.

Also run go mod tidy to remove 14 stale go.sum entries left behind by renovate dependency updates, and add a CI step using `go mod tidy -diff` to verify go.sum stays up to date.

Assisted-By: Cursor


operator/Containerfile
```diff
@@ -24,7 +24,7 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} \
     go build -a \
     -ldflags "-X github.com/konflux-ci/konflux-ci/operator/pkg/version.Version=${OPERATOR_VERSION:-unknown} \
               -X github.com/konflux-ci/konflux-ci/operator/pkg/version.GitCommit=${GIT_COMMIT:-unknown}" \
-    -o /opt/app-root/manager cmd/main.go
+    -o /opt/app-root/manager ./cmd/main.go
```
yftacherzog, line 27: Fails now on:
```
Error: #16 0.153 stat /workspace/cmd/main.go: permission denied
```

gbenhaim: Yeah, which is strange since we set the permissions on line 16.

yftacherzog: I took your commit and made a small change in permissions: https://github.com/konflux-ci/konflux-ci/pull/5263.
It seems to be passing the build. Are we ok with such permissions?

Also, might not be related, but it might make sense to build before we start the cluster.

yftacherzog: I added this in another PR that include both changes in different commits. https://github.com/konflux-ci/konflux-ci/pull/5264

yftacherzog: We can merge either I think.

Issue #5278, opened 2026-02-13
Deploy-local.sh not enabling image-controller by default


hmagph:
Following default guidance to install Konflux locally with kind, it didn't come with image controller out of the box.

```
[...]
=========================================
Step 6: Creating GitHub integration secrets
=========================================
Creating Pipelines-as-Code secrets...
Waiting for namespace: pipelines-as-code
Creating secret in pipelines-as-code...
secret/pipelines-as-code-secret created
Waiting for namespace: build-service
Creating secret in build-service...
secret/pipelines-as-code-secret created
Waiting for namespace: integration-service
Creating secret in integration-service...
secret/pipelines-as-code-secret created
✓ Secrets created

Creating image-controller Quay secret...
Waiting for namespace: image-controller
WARNING: Namespace image-controller not created after 60 seconds
         Secret will need to be created manually
[...]
```

After digging a bit, realized the default CR (at ./operator/config/samples/konflux_v1alpha1_konflux.yaml) should enable it in operator setup:

```
  imageController:
    enabled: true
```

This would save having to later do it manually, and confuse new users.

PR #5297, opened 2026-02-14
fix: auto-detect image-controller requirement in deploy-local.sh

arewm:
When QUAY_TOKEN and QUAY_ORGANIZATION are configured but no CR is explicitly specified, auto-select konflux-e2e.yaml which enables image-controller. This eliminates the confusing "namespace not created" warning users encounter when expecting Quay integration to work.

Users can still override by setting KONFLUX_CR or passing a CR argument.

Fixes: https://github.com/konflux-ci/konflux-ci/issues/5278

Assisted-by: Claude Code (Sonnet 4.5)

### DIFF
--- a/scripts/deploy-local.sh
+++ b/scripts/deploy-local.sh
@@ -92,8 +92,20 @@ export KIND_CLUSTER KIND_MEMORY_GB PODMAN_MACHINE_NAME REGISTRY_HOST_PORT ENABLE
 export INCREASE_PODMAN_PIDS_LIMIT
 export GITHUB_PRIVATE_KEY GITHUB_APP_ID WEBHOOK_SECRET QUAY_TOKEN QUAY_ORGANIZATION
 
-# Get Konflux CR file path (env var, command-line arg, or default)
-KONFLUX_CR="${KONFLUX_CR:-${1:-${REPO_ROOT}/operator/config/samples/konflux_v1alpha1_konflux.yaml}}"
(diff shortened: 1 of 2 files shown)


scripts/deploy-local.sh
```diff
@@ -93,7 +93,17 @@ export INCREASE_PODMAN_PIDS_LIMIT
 export GITHUB_PRIVATE_KEY GITHUB_APP_ID WEBHOOK_SECRET QUAY_TOKEN QUAY_ORGANIZATION
 
 # Get Konflux CR file path (env var, command-line arg, or default)
-KONFLUX_CR="${KONFLUX_CR:-${1:-${REPO_ROOT}/operator/config/samples/konflux_v1alpha1_konflux.yaml}}"
+# Auto-select e2e CR when Quay credentials are configured but no explicit CR specified
```
gbenhaim, line 98: you can simplify this condition by

```
KONFLUX_CR="$1"
```

and removing 
```
&& [ -z "${1:-}" ];
```

arewm: We can simplify it, but that proposal would prevent us from being able to set that variable directly before calling the script (i.e. in the e2e tests or the env file). I'll change it to

```
KONFLUX_CR="${1:-$KONFLUX_CR}"
```

to use the command line argument if set, otherwise use the environment variable. This means that if a user has an environment variable, a command line argument will take precedence.

gbenhaim (approved):

PR #5450, opened 2026-02-21
feat: add reusable GitHub Action for deploying Konflux on kind (#2326)

### LINKED ISSUE #2326: Create a GitHub action for deploying Konflux on a kind cluster
Deploying Konflux is a multi-step process that can be cumbersome to many (although it should get better once we have Helm charts). We want to encourage developers of the various Konflux services to use the upstream edition ([konflux-ci/konflux-ci](https://github.com/konflux-ci/konflux-ci)) when testing their components. To assist we should make it simple to deploy Konflux on a provided kind cluster in existing CI systems. Many of our contributors use a mixture of Tekton and GitHub Actions for their CI. This enhancement is specifically for the latter.

abhayrajjais01:
## Summary

Closes #2326

Adds a reusable composite action at `.github/actions/deploy-konflux/` that
wraps the existing `scripts/deploy-local.sh` deployment logic, making it
easy for any Konflux service developer to spin up a full Konflux stack in
their GitHub Actions CI with a single `uses:` step:

```yaml
- uses: konflux-ci/konflux-ci/.github/actions/deploy-konflux@main
  with:
    github-app-id: ${{ secrets.GITHUB_APP_ID }}
    github-private-key: ${{ secrets.APP_PRIVATE_KEY }}
    webhook-secret: ${{ secrets.APP_WEBHOOK_SECRET }}

Changes
.github/actions/deploy-konflux/action.yml — composite action that:

Installs kind and kubectl with automatic amd64/arm64 detection
Disables AppArmor and pre-configures inotify limits for Ubuntu 24.04+
Exposes all deploy-local.sh environment variables as typed inputs with sensible defaults (renovate-managed version pins for kind/kubectl)
Supports all three operator install methods: release, local, build
Accepts a konflux-ci-path input to skip auto-checkout when called from within the repo itself
Outputs kubeconfig-path and cluster-name for downstream steps
.github/workflows/example-deploy-konflux.yml — manually-triggered
example workflow showing how external consumers use the action

.github/workflows/operator-test-e2e.yaml — refactored to use the
new action (replaces ~25 manual setup lines with a single uses: step);
also adds .github/actions/deploy-konflux/** to the changed-files filter
so action changes trigger E2E tests

Test plan
 YAML validation: python -c "import yaml; yaml.safe_load(open('.github/actions/deploy-konflux/action.yml'))"
 Shellcheck runs automatically via differential-shellcheck.yaml on the PR
 The refactored operator-test-e2e.yaml passes its full E2E suite (proves the action works end-to-end)
 example-deploy-konflux.yml can be triggered manually via gh workflow run with real secrets


gbenhaim:
Thanks for the contribution, but since we moved to an operator based installation the issue is no longer relevant.

Thanks again !

PR #5469, opened 2026-02-23
feat: scale changelog generation to all upstream projects

ergonzal22:
## Summary
- Adds scripts and CI workflow to generate curated upstream changelogs for Konflux releases
- Extracts upstream commit SHAs from kustomization files at each release tag using `ref` (from `?ref=` resource URLs) or `digest` (via `skopeo inspect` OCI labels) extraction
- Covers 9 upstream components in parallel: application-api, build-service, enterprise-contract, image-controller, integration-service, internal-services, namespace-lister, release-service, UI
- Includes fallback format for repos without conventional commits: `> N commits since last release - [view diff](...)`
- Uses semantic version sorting (`--sort=-version:refname`) and `git merge-base` for tag detection, supporting release-branch workflows

Supersedes #5222 (POC with 4 components) and #5429 (previous iteration from fork).

Ref: [KFLUXVNGD-727](https://issues.redhat.com/browse/KFLUXVNGD-727), [KFLUXVNGD-726](https://issues.redhat.com/browse/KFLUXVNGD-726)
Epic: [KFLUXVNGD-653](https://issues.redhat.com/browse/KFLUXVNGD-653)

## Verification
- [x] 14 CI tests pass (ranges, edge cases, ref/digest extraction, fallback format, parallel ordering)
- [x] All 9 components verified locally in `ubuntu:latest` container
- [x] Digest extraction works for UI and namespace-lister (skopeo inspect OCI labels)
- [x] Backwards-compat fallback to `newTag` works on old tags (v0.0.6 -> v0.0.7)
- [x] Parallel execution produces deterministic, alphabetically ordered output

### DIFF
--- a/.github/scripts/generate-changelog.sh
+++ b/.github/scripts/generate-changelog.sh
@@ -1,0 +1,352 @@
+#!/bin/bash
+set -euo pipefail
+
+# Generate Changelog Script
+# Generates a markdown changelog from upstream conventional commits between two operator tags.
+#
+# For each configured component, this script:
+# 1. Extracts the upstream commit SHA from the kustomization file at each tag
+# 2. Queries the GitHub API for commits between those SHAs
+# 3. Filters to conventional commit types (feat, fix) excluding bot authors
+# 4. Outputs formatted markdown to stdout
+#
+# Usage:
+#   generate-changelog.sh <old_tag> <new_tag> [component...]
+#
+# Arguments:
+#   old_tag       - Previous release tag (e.g., v0.0.8)
+#   new_tag       - Current release tag (e.g., v0.0.9)
+#   component...  - Optional list of components to process (default: all configured)
+#
+# Environment:
+#   GH_TOKEN      - GitHub token for API access (required)
+#
+# Dependencies: gh, jq, git, yq, skopeo (for digest extraction)
+#
+# Example:
+#   export GH_TOKEN=$(gh auth token)
+#   generate-changelog.sh v0.0.8 v0.0.9
+#   generate-changelog.sh v0.0.8 v0.0.9 release-service
+
+if [ "$#" -lt 2 ]; then
+  echo "Error: Invalid number of arguments" >&2
+  echo "Usage: $0 <old_tag> <new_tag> [component...]" >&2
+  exit 1
+fi
+
+OLD_TAG="$1"
+NEW_TAG="$2"
+shift 2
+
+# Component configuration
+# Format: "kustomization_path|image_or_pattern|github_org/repo|display_name|extraction_type"
+#
+# - kustomization_path: Path to the kustomization file containing the image or resource reference
+# - image_or_pattern:   For ref: repo pattern to match in `resources:` URLs
+#                        For digest: image name in `images:` section (used with skopeo inspect)
+# - github_org/repo:    GitHub repository for the Compare API call
+# - display_name:       Human-readable name for the changelog section
+# - extraction_type:    How to extract the SHA:
+#                        "ref"    — from ?ref= parameter in resources[] URLs
+#                        "digest" — from images[].digest via skopeo inspect OCI labels
+#
+# Conventional Commits Audit (2026-02)
+# =====================================
+# Strong adoption:  release-service, integration-service, ui (konflux-ui)
+# Partial adoption: build-service, image-controller, application-api
+# Not adopted:      enterprise-contract (conforma/crds), internal-services
+#
+# Components without upstream git references (no changelog possible):
+#   rbac, info, cert-manager, default-tenant, registry
+#
+# The script uses runtime detection: if no feat/fix commits are found,
+# it falls back to "N commits - [view diff]" format automatically.
+#
+# Add new components here to extend changelog coverage.
+declare -A COMPONENT_CONFIG=(
+  ["application-api"]="operator/upstream-kustomizations/application-api/kustomization.yaml|redhat-appstudio/application-api|konflux-ci/application-api|Application API|ref"
+  ["build-service"]="operator/upstream-kustomizations/build-service/core/kustomization.yaml|konflux-ci/build-service|konflux-ci/build-service|Build Service|ref"
+  ["enterprise-contract"]="operator/upstream-kustomizations/enterprise-contract/core/kustomization.yaml|conforma/crds|conforma/crds|Enterprise Contract|ref"
+  ["image-controller"]="operator/upstream-kustomizations/image-controller/core/kustomization.yaml|konflux-ci/image-controller|konflux-ci/image-controller|Image Controller|ref"
+  ["integration-service"]="operator/upstream-kustomizations/integration/core/kustomization.yaml|konflux-ci/integration-service|konflux-ci/integration-service|Integration Service|ref"
+  ["internal-services"]="operator/upstream-kustomizations/release/internal-services/kustomization.yaml|redhat-appstudio/internal-services|redhat-appstudio/internal-services|Internal Services|ref"
+  ["namespace-lister"]="operator/upstream-kustomizations/namespace-lister/kustomization.yaml|quay.io/konflux-ci/namespace-lister|konflux-ci/namespace-lister|Namespace Lister|digest"
+  ["release-service"]="operator/upstream-kustomizations/release/core/kustomization.yaml|konflux-ci/release-service|konflux-ci/release-service|Release Service|ref"
+  ["ui"]="operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml|quay.io/konflux-ci/konflux-ui|konflux-ci/konflux-ui|UI|digest"
+)
+
+# Bot authors to exclude from changelog
+BOT_AUTHORS='["dependabot[bot]", "renovate[bot]", "github-actions[bot]", "konflux-internal-p02[bot]", "red-hat-konflux[bot]"]'
+
+# Determine which components to process.
+# Sort keys for deterministic output order (bash associative arrays are unordered).
+if [ "$#" -gt 0 ]; then
+  COMPONENTS=("$@")
+else
+  IFS=$'\n' read -r -d '' -a COMPONENTS < <(printf '%s\n' "${!COMPONENT_CONFIG[@]}" | sort && printf '\0') || true
+fi
+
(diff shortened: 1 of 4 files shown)


.github/scripts/prepare-release-notes.sh
```diff
@@ -28,6 +28,19 @@ IMAGE_TAG="$2"
...
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# get_previous_tag finds the most recent stable release tag reachable from
+# $GIT_REF. Uses --merged to scope to ancestor tags (respects branch topology),
+# --sort=-creatordate to order by git history, and a strict regex to match only
+# stable semver tags (vX.Y.Z) — excluding pre-release suffixes like -rc, -beta,
+# -alpha, etc.
+get_previous_tag() {
+  git tag --merged "$GIT_REF" --sort=-creatordate 2>/dev/null \
```
yftacherzog: I'm not sure this will work.
Suppose this progression:

day 1: we work on 0.1.0 on branch main.
day 2: we create branch release-0.1 and create a release candidate tag on the point of divergence (`v0.1.0-rc.1`)
day 3: we make a commit to main and cherry-pick it to release-0.1
day 4: we make some more commits to main without cherry picking.
day 5: we tag the cherry-picked commit in release-0.1 with `v0.1.0`.

Eventually, we create release-0.2 and eventually tag `v0.2.0` there. The release notes should compare `v0.2.0` and `v0.1.0`, but 0.1.0 is not reachable from 0.2.0.

```bash
|
|--v0.1.0-rc.1--v0.1.0--[release-0.1]
|
|
|--v0.2.0-rc.1--v0.2.0-rc.2--v0.2.0--[release-0.2]
|
[main]
```

yftacherzog: When z=0, we need to find the x.y.0 that is semantically right before the release we want to make.

ergonzal22: Changed to using `--sort=-version:refname` across all tags. So right now it is returning the highest stable one. So for v0.2.0 it will return v0.1.2. Why would you like it to return v0.1.0? Because that's what I get from your last comment

yftacherzog: It's a delicate point. There should not be a direct pass from v0.1.2 to v0.2.0, but this is most probably be correct for v0.1.0 to v0.2.0 as well.

We plan it so that v0.1.1 and v0.1.2 should contain commits that were back-ported from the main branch (which at this time is used for working on v0.2.0), so those commits should be redundant as far as I see it.

Suppose that one main we have:
```bash
| feat0
| -- fix1 -- fix2 [release-0.1]
| feat1
| fix1
| feat2
| fix2
| feat3
[main]
```
We need the release notes for the release that will include feat3, to also mention feat1 and feat2, even if there was v0.1.z that included fix1 and fix2.

I'm not familiar with the command you use here, but I think what we need is take everything from the common ancestor of the z=0 that we're releasing now and the previous y-stream. I.e. from the point of divergence to the tag that we're releasing now.

yftacherzog: There are commands for getting that common ancestor.

ergonzal22: Changed! Feel free to check it. Also, I updated the output of the notes. Now it is something like this:

### Integration Service

#### Features
- whatever feature

#### Bug Fixes
- whatever bug fix

I think it will look nicer than the *feat* and *fix* list

yftacherzog:
> @yftacherzog btw, regarding the changelog test that I created, do you want to keep it as it is, remove it entirely or maybe leave only the part that shows how the output would look with the current status. I vote for option 3, but I wanted to know your thoughts

How should it be used?

.github/scripts/prepare-release-notes.sh
```diff
@@ -28,6 +28,33 @@ IMAGE_TAG="$2"
...
+# First finds the highest previous stable release tag (semantic version sort,
+# strict vX.Y.Z regex excluding pre-release suffixes). Then returns the common
+# ancestor (merge-base) between that tag and GIT_REF.
+#
+# This handles both release models correctly:
+# - Patch releases (z>0): merge-base of two commits on the same branch returns
+#   the older one, equivalent to using the tag directly.
+# - Minor releases (z=0): merge-base returns the divergence point between
+#   release branches, capturing all changes since the branch fork — including
+#   feats/fixes that were also backported to patch releases on the old branch.
```
yftacherzog, line 43: Sounds right!
@gbenhaim @avi-biton can you take a look to see that you agree with this approach for setting the starting point from which to evaluate changes.

.github/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,338 @@
...
+
+# Component configuration
+# Format: "kustomization_path|image_or_pattern|github_org/repo|display_name|extraction_type"
+#
+# - kustomization_path: Path to the kustomization file containing the image or resource reference
+# - image_or_pattern:   For newTag: image name in the `images:` section
+#                        For ref: repo pattern to match in `resources:` URLs
+# - github_org/repo:    GitHub repository for the Compare API call
+# - display_name:       Human-readable name for the changelog section
+# - extraction_type:    How to extract the SHA: "newTag" (from images[].newTag) or "ref" (from resources[] ?ref=)
```
yftacherzog: I noticed that we have at least one case in which we have `digest` instead of `newTag`.
We need to see if we can change cases like this to use `newTag` (don't recall what's the reason it's that way). Otherwise, we'll need to handle that case as well (I'd prefer we have it unified instead of another case to handle here).

yftacherzog: https://github.com/konflux-ci/konflux-ci/blob/main/operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml#L15

yftacherzog: I ran into that when trying to figure out an issue with renovate updates. I think we need to be able to support Digest, but it might be that if we only have an image then we should only use Digest. So it's either ref or digest, but I'm not sure yet.

ergonzal22: But that's a ubi9 nginx image. Do we really need to add that to our changelog?

ergonzal22: `namespace-lister` also uses digest: https://github.com/konflux-ci/konflux-ci/blob/main/operator/upstream-kustomizations/namespace-lister/kustomization.yaml#L14 

Can that be changed to `newTag` instead?

.github/scripts/prepare-release-notes.sh
gbenhaim: Looks great !

.github/scripts/prepare-release-notes.sh
gbenhaim: Lgtm

.github/scripts/generate-changelog.sh
gbenhaim: Yes

yftacherzog: We have several new tags but they should also have git ref with the same tag in the same file. If not then we need to change that to use digest as renovate can not make order out of this arbitrary image tag. It needs a digest.

yftacherzog: > But that's a ubi9 nginx image. Do we really need to add that to our changelog?

I dont think we need the ubi image in our change log v

yftacherzog: The problem is with the other ui images. We need to have them use digest so we need to be able to use it. We don't need to be able to use newtag as wherever we have newtag we should also have git ref.

yftacherzog: I'm trying to think if there's a way to bridge this, as I think finding the git sha tag with the specific digest will require `skopeo inspect` call for every potential tag.

yftacherzog: Might be that the quay api can help us here, but we need to take pagination into consideration.

yftacherzog: Claude is suggesting something like this:
```bash
  #!/bin/bash
  find_git_sha_for_digest_fast() {
      local repo="$1"
      local target_digest="$2"
      local page=1

      # Start with newest tags (default API behavior)
      while true; do
          local response=$(curl -s "https://quay.io/api/v1/repository/${repo}/tag/?page=${page}&limit=100")

          # Check git SHA tags on this page (newest first)
          local found_tag=$(echo "$response" | jq -r --arg target "$target_digest" \
              '.tags[] | select(.name | test("^[a-f0-9]{40}$")) | select(.manifest_digest == $target) | .name')

          if [ -n "$found_tag" ]; then
              echo "$found_tag"
              return 0
          fi

          # Early exit optimization: if no git SHA tags on this page, 
          # and we're past page 1, likely won't find any newer ones
          local git_sha_count=$(echo "$response" | jq -r '.tags[] | select(.name | test("^[a-f0-9]{40}$")) | .name' | wc -l)
          if [ "$git_sha_count" -eq 0 ] && [ "$page" -gt 1 ]; then
              break
          fi

          local has_additional=$(echo "$response" | jq -r '.has_additional')
          if [ "$has_additional" != "true" ]; then
              break
          fi

          ((page++))
      done

      return 1
  }
```

I think this will give us the result in much fewer iterations (100 seems to be the max page size). We probably should consider retries.

yftacherzog: We had a call around this just now. It might be simpler and the answer is available in the labels in the skopeo inspect call. Just need to be sure that what we get there for the UI is valid. If it's not (i.e. points to the real revision). If it doesn't, then we can look into fixing the UI image build process.

gbenhaim:
> > > regarding the changelog test
> > > > How should it be used?
> > 
> > 
> > I basically created it to test that the features here were working. Maybe if there are changes to this in the future, it could work to test those. But specifically the one that creates the preview of the changelog I think could be interesting to see how would it look before creating it.
> > If we decide to keep something, it should be the test 4 (the full integration preview — shows what a release would look like right now) and maybe test 3 + test 13 as minimal regression checks (same-tag negative test, parallel ordering)
> 
> Yes I think that keeping the preview is useful. I checked the github action. I can't say I understand it's output and I also didn't see a full changelog.

I was looking at the wrong place. I saw now https://github.com/konflux-ci/konflux-ci/actions/runs/22401531405?pr=5469 which is great.

PR #5499, opened 2026-02-24
chore(KFLUXVNGD-690): KonfluxSegmentBridge controller & Konflux wiring

Omeramsc:
Add the KonfluxSegmentBridgeReconciler that deploys and manages the
segment-bridge telemetry CronJob, its RBAC, and Segment configuration
for both the CronJob and the UI.

depends on https://github.com/konflux-ci/konflux-ci/pull/5099

PR #5551, opened 2026-02-25
chore: expose Segment telemetry config to UI

Omeramsc:
Extend the KonfluxInfo controller to create UI-facing resources for
Segment telemetry, replacing the cross-namespace writes
that were removed from the SegmentBridge controller per review feedback
on PR #5499.


gbenhaim:
@Omeramsc I'll take care of making the key available for the front-end, so you can focus on completing the other prs.

PR #5600, opened 2026-02-26
feat(KFLUXVNGD-624): Add OCP deployment script for Konflux.

kelchen123:
Add OCP deployment script for deploy-konflux-on-ocp.sh.

Assisted by: Cursor

### DIFF
--- a/dependencies/tekton-chains-rbac-ocp/tekton-chains.yaml
+++ b/dependencies/tekton-chains-rbac-ocp/tekton-chains.yaml
@@ -1,0 +1,168 @@
+# Tekton Chains RBAC for OpenShift Pipelines
+# This is a modified version of tekton-chains-rbac for OpenShift Pipelines,
+# which uses 'openshift-pipelines' namespace instead of 'tekton-pipelines'.
+#
+# Give authenticated users read permission for tekton-chains public key
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "0"
+  name: tekton-chains-public-key-viewer
+rules:
+- apiGroups:
+  - ""
+  resourceNames:
+  - public-key
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "0"
+  name: tekton-chains-public-key-viewer
+  namespace: openshift-pipelines
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: tekton-chains-public-key-viewer
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "0"
+  name: chains-secrets-admin
+  namespace: openshift-pipelines
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "0"
+  name: chains-secret-admin
+  namespace: openshift-pipelines
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - list
+  - create
+  - get
+  - update
+  - patch
+  - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "0"
+  name: chains-secret-admin
+  namespace: openshift-pipelines
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: chains-secret-admin
+subjects:
+- kind: ServiceAccount
+  name: chains-secrets-admin
+  namespace: openshift-pipelines
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "1"
+  name: tekton-chains-signing-secret
+  namespace: openshift-pipelines
+spec:
+  ttlSecondsAfterFinished: 7200
+  template:
+    metadata:
+      annotations:
+        argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    spec:
+      containers:
+      - command:
+        - /bin/bash
+        - -c
+        - |
+          set -o errexit
+          set -o nounset
+          set -o pipefail
+
+          # For OpenShift Pipelines, use openshift-pipelines namespace
+          namespace="openshift-pipelines"
+          secret="signing-secrets"
+
+          cd /tmp
+
+          if [ "$(kubectl get secret "$secret" -n "$namespace" -o jsonpath='{.data}' --ignore-not-found --allow-missing-template-keys)" != "" ]; then
+            echo "Signing secret exists and is non-empty."
+          else
+            # Delete secret/signing-secrets if already exists since by default cosign creates immutable secrets
+            kubectl delete secrets "$secret" -n "$namespace" --ignore-not-found=true
+
+            # To make this run conveniently without user input let's create a random password
+            RANDOM_PASS=$( openssl rand -base64 30 )
+
+            # Generate the key pair secret directly in the cluster.
+            # The secret should be created as immutable.
+            echo "Generating k8s secret/$secret in $namespace with key-pair"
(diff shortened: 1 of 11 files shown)


deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,145 @@
...
+oc wait --for=condition=Ready tektonconfig/config --timeout=600s
+
+# Wait for Tekton webhook services to be available (TektonConfig Ready doesn't guarantee this)
+echo "Waiting for Tekton webhook services..."
+until oc get service tekton-operator-proxy-webhook -n openshift-pipelines &>/dev/null; do
+    echo "  Waiting for tekton-operator-proxy-webhook service..."
+    sleep 10
+done
+oc wait --for=condition=Available deployment/tekton-operator-proxy-webhook -n openshift-pipelines --timeout=120s
+echo "OpenShift Pipelines is ready!"
```
amisstea: Logically, since the pipelines operator is still a dependency, it would make sense for this to reside in the `deploy-deps.sh` script. We should then rename the `SKIP_TEKTON` var since that script would always install it, just using a different approach depending on the platform.

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,136 @@
...
+# The operator needs 'delete' permission on RBAC resources to set ownerReferences
+# This is required for UI and info components to work correctly
+echo ""
+echo "=== Step 5/6: Patching operator RBAC for OCP compatibility ==="
+oc patch clusterrole konflux-operator-manager-role --type='json' -p='[
+  {"op": "add", "path": "/rules/-", "value": {"apiGroups": ["rbac.authorization.k8s.io"], "resources": ["clusterroles", "clusterrolebindings", "roles", "rolebindings"], "verbs": ["delete"]}}
+]'
+echo "RBAC patched. Restarting operator to pick up new permissions..."
+oc rollout restart deployment/konflux-operator-controller-manager -n konflux-operator
+oc rollout status deployment/konflux-operator-controller-manager -n konflux-operator --timeout=120s
```
amisstea: Do you think we could add the delete permission to the clusterrole directly to avoid this step? We can do this in a followup PR.

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,136 @@
...
+echo "Git log:"
+git log --oneline -3
+echo ""
+
+# Configuration
+KONFLUX_OPERATOR_REPO="${KONFLUX_OPERATOR_REPO:-quay.io/redhat-user-workloads/konflux-vanguard-tenant/konflux-operator}"
+
+# Get the script's directory
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "${SCRIPT_DIR}"
```
amisstea: I don't think we need this `SCRIPT_DIR` variable.

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,136 @@
...
 
-if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
-    main "$@"
+# On OCP, use 'oc' as kubectl if kubectl is not available
+# oc is a superset of kubectl and works for all kubectl commands
+if ! command -v kubectl &>/dev/null && command -v oc &>/dev/null; then
+    echo "kubectl not found, using 'oc' as kubectl"
+    function kubectl() { oc "$@"; }
+    export -f kubectl
 fi
```
amisstea, line 29: `kubectl` isn't used in this script. You can remove this block.

kelchen123: It caused some error when I didn't have this block, I think the deploydeps script uses it internally: "Some requirements are missing, please install the following requirements first:
kubectl (v1.31.4 or newer)"

amisstea: You're right. I thought I saw this repeated in `deploy-deps.sh` as well but that's clearly not the case. Nothing to see here :smile:.

deploy-konflux-on-ocp.sh
kelchen123: good idea, noted for followup pr!

deploy-deps.sh
```diff
@@ -101,7 +101,57 @@ test_pvc_binding(){
...
+  channel: latest
+  name: openshift-pipelines-operator-rh
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
+EOF
+
+    # Wait for TektonConfig CRD to be available
+    echo "  ⏳ Waiting for TektonConfig CRD..." >&2
+    until kubectl get crd tektonconfigs.operator.tekton.dev &>/dev/null; do
+        sleep 10
```
gbenhaim, line 129: Can this loop forever?

deploy-deps.sh
```diff
@@ -101,7 +101,57 @@ test_pvc_binding(){
...
+    done
+
+    # Wait for TektonConfig resource to be created and ready
+    echo "  ⏳ Waiting for TektonConfig to be ready..." >&2
+    until kubectl get tektonconfig config &>/dev/null; do
+        sleep 10
```
gbenhaim, line 142: Same comment as above

deploy-deps.sh
```diff
@@ -101,7 +101,57 @@ test_pvc_binding(){
...
+          "TektonConfig did not become ready within the allocated time"
+
+    # Wait for Tekton webhook services (TektonConfig Ready doesn't guarantee this)
+    echo "  ⏳ Waiting for Tekton webhook services..." >&2
+    until kubectl get service tekton-operator-proxy-webhook -n openshift-pipelines &>/dev/null; do
+        sleep 10
+    done
+    retry "kubectl wait --for=condition=Available deployment/tekton-operator-proxy-webhook -n openshift-pipelines --timeout=120s" \
+          "Tekton webhook did not become available within the allocated time"
+
```
gbenhaim, line 162: ${script_path}/dependencies/tekton-chains-rbac

This should be applied as well.

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,131 @@
...
+    echo "Operator image: ${OPERATOR_IMAGE}"
+fi
+
+# Wait for operator image to be available (Konflux CI may still be building it)
+echo "Checking if operator image is available..."
+MAX_WAIT=900  # 15 minutes
+WAIT_INTERVAL=30
+WAITED=0
+# Use --filter-by-os for multi-arch images
+while ! oc image info --filter-by-os=linux/amd64 "${OPERATOR_IMAGE}" &>/dev/null; do
```
gbenhaim, line 66: Nice

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,131 @@
...
+echo "=== Step 2/6: Installing Operator CRDs ==="
+cd operator
+# Clear GOFLAGS to allow downloading tools (CI may have -mod=vendor set)
+GOFLAGS="" make install
+
+# Step 3: Deploy the operator using the Konflux-built image
+echo ""
+echo "=== Step 3/6: Deploying Operator ==="
+echo "Image: ${OPERATOR_IMAGE}"
+GOFLAGS="" make deploy IMG="${OPERATOR_IMAGE}"
```
gbenhaim, line 99: Why do we need to care about GOFLAGS here?

deploy-konflux-on-ocp.sh
```diff
@@ -1,9 +1,131 @@
...
+echo "=== Step 4/6: Waiting for Operator to be ready ==="
+oc wait --for=condition=Available deployment/konflux-operator-controller-manager \
+    -n konflux-operator --timeout=300s
+echo "Operator is ready!"
+
+# Step 5: Fix operator RBAC permissions for OCP
+# The operator needs 'delete' permission on RBAC resources to set ownerReferences
+# This is required for UI and info components to work correctly
+echo ""
+echo "=== Step 5/6: Patching operator RBAC for OCP compatibility ==="
```
gbenhaim: Interesting
I wonder how we didn't notice this lack of permissions. I deployed on of 4.20 many time and it worked.

Regardless, permission issues should be fixed in the operator code by adding kubebuilder annotation on the reconcilers.

deploy-deps.sh
gbenhaim, line 192: Cert manager should be deployed using the redhat operator as well.

deploy-deps.sh
```diff
@@ -101,7 +101,57 @@ test_pvc_binding(){
...
+    else
+        echo "  🐱 Installing Tekton via upstream Operator..." >&2
+        deploy_upstream_tekton
+    fi
+}
+
+deploy_openshift_pipelines() {
+    # Install OpenShift Pipelines Operator via OLM
+    cat <<EOF | kubectl apply -f -
+apiVersion: operators.coreos.com/v1alpha1
```
gbenhaim: We store dependencies in manifests under the dependencies directory. I suggest to be consistent with this and move the subscription. This will also let's kubelint to validate it.

gbenhaim:
Great progress! I left some comments.

deploy-deps.sh
kelchen123: Good catch, added a timeout

deploy-deps.sh
kelchen123: added a timeout

deploy-deps.sh
kelchen123: added

deploy-konflux-on-ocp.sh
kelchen123: originally the CI set GOFLAGS to mod-vendor, causing: `go: sigs.k8s.io/controller-tools/cmd/controller-gen@v0.18.0: cannot query module due to -mod=vendor`

deploy-deps.sh
kelchen123: added

deploy-deps.sh
kelchen123: done

deploy-konflux-on-ocp.sh
kelchen123: done

dependencies/tekton-chains-rbac-ocp/tekton-chains.yaml
amisstea, line 1: Perhaps we should consider restructuring the `tekton-chains-rbac` and `tekton-chains-rbac-ocp` kustomize components so that the latter inherits from the former. That could help reduce the repetition currently needed.

kelchen123: That' a good point, I can do that in a follow-up pr!

PR #5601, opened 2026-02-26
chore(KFLUXVNGD-795): add secret references to segment-bridge

Omeramsc:
Changes the default segment host as well as reading the host and key
secrets, and add logic to append "/batch" to the host url, as expected
from the segment-bridge component.

Depends on #5499

### DIFF
--- a/operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
+++ b/operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
@@ -53,9 +53,12 @@ const (
 )
 
 // SegmentBridgeCleanupGVKs defines which resource types should be cleaned up when they are
-// no longer part of the desired state. All resources managed by this controller are always
-// applied, so no cleanup GVKs are needed (they're always tracked and never become orphans).
-var SegmentBridgeCleanupGVKs = []schema.GroupVersionKind{}
+// no longer part of the desired state. Only optional/conditional resources are listed here.
+// Always-applied resources don't need cleanup (they're always tracked and never become orphans).
+var SegmentBridgeCleanupGVKs = []schema.GroupVersionKind{
+	// Secret is conditional - only created when a Segment write key is configured
+	{Group: "", Version: "v1", Kind: "Secret"},
+}
 
 // SegmentBridgeClusterScopedAllowList restricts which cluster-scoped resources can be deleted
 // during orphan cleanup. All cluster-scoped resources managed by this controller are always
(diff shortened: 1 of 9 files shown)


operator/api/v1alpha1/konfluxsegmentbridge_types.go
```diff
@@ -17,28 +17,52 @@ limitations under the License.
...
 	// +optional
 	SegmentAPIURL string `json:"segmentAPIURL,omitempty"`
+
+	// SegmentKeySecretRef references an existing Secret containing Segment
+	// credentials. Expected keys: SEGMENT_WRITE_KEY (required) and
+	// SEGMENT_HOST (optional, full URL path without "/batch").
+	// Inline fields (SegmentKey, SegmentAPIURL) take precedence over
+	// values read from this Secret.
+	// +optional
+	SegmentKeySecretRef *SegmentSecretRef `json:"segmentKeySecretRef,omitempty"`
```
ifireball: I don't think we need to add this to the CRD. Nothing like this was intended in any of the issues.

Omeramsc: I wasn't sure if we want the key to be visible in plain text in the CR spec. should we keep it as it was? is that not a concern?

ifireball: not a concern, the key is visible in the client side JS code

Omeramsc: removed

operator/api/v1alpha1/konfluxsegmentbridge_types.go
```diff
@@ -49,13 +54,23 @@ func (s *KonfluxSegmentBridgeSpec) GetSegmentKey() string {
...
+	if s != nil && s.SegmentAPIURL != "" {
+		url = s.SegmentAPIURL
 	}
-	return s.SegmentAPIURL
+	return sanitizeSegmentHost(url)
+}
+
+// sanitizeSegmentHost strips trailing slashes and an accidental "/batch" suffix
+// so that callers can safely append "/batch" to the result.
+func sanitizeSegmentHost(url string) string {
```
gbenhaim: Maybe use net.Url? It provides method for processing urls.

operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
```diff
@@ -148,10 +143,21 @@ func (r *KonfluxSegmentBridgeReconciler) applyManifests(ctx context.Context, tc
...
+func reconcileSegmentBridgeSecret(ctx context.Context, tc *tracking.Client, spec *konfluxv1alpha1.KonfluxSegmentBridgeSpec) error {
 	log := logf.FromContext(ctx)
 
+	segmentKey := spec.GetSegmentKey()
+	if segmentKey == "" {
+		log.Info("No Segment write key configured; skipping Secret creation")
+		return nil
+	}
+
+	batchURL := spec.GetSegmentAPIURL() + "/batch"
```
gbenhaim: Same comment as above. If spec.GetSegmentAPIURL() will return net.Url instance joining the parh would be safer.

PR #5689, opened 2026-03-02
feat(RELEASE-2158): Deploy self-hosted Quay to the cluster

querti:
The self-hosted Quay instance will be used in the
release-service-catalog's integration tests, verifying that the pipeline can successfully use self-hosted Quay. The deployment is minimalist, only having resources that are necessary for the basic functions of Quay. Operator was not used because it creates too heavy, production-ready deployment. Despite this, Quay still requires 8GB of memory and 4 CPU cores to run smoothly and not have OOM issues. That's why the deployment is disabled by default and will only run if SKIP_QUAY=false. Cert manager is used to create a certificate that Quay will use as a CA cert. This way, the cert can be trusted through the trusted-ca configmap, and 3rd party tools in the pipeline can interact with the instance.

Assisted-by: Cursor

### DIFF
--- a/dependencies/quay/kustomization.yml
+++ b/dependencies/quay/kustomization.yml
@@ -1,0 +1,16 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - namespace.yaml
+  - postgres.yaml
+  - redis.yaml
+  - certificate.yaml
+  - quay.yaml
+images:
+  - name: quay.io/projectquay/quay
+    newTag: "3.16.2"
+  - name: quay.io/sclorg/postgresql-15-c9s
+    newTag: "20260302"
+  - name: quay.io/sclorg/redis-7-c9s
+    newTag: "20250108"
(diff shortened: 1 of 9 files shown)


dependencies/quay/kustomization.yml
yftacherzog, line 1: I think we should have here `images` block, which will allow us to have renovate track image updates. On the other hand, we don't have a way to verify that any such updates don't regress, as we don't have enough resources to run this in CI.

What do you think is the best approach here?

querti: It's hard to know the risk of regression, but it would break our e2e test so we would find out quickly. How about we keep it here and if we observe the deployment breaking due to renovate updates, we move it back?

yftacherzog: I'm good with that

gbenhaim: > It's hard to know the risk of regression, but it would break our e2e test so we would find out quickly. How about we keep it here and if we observe the deployment breaking due to renovate updates, we move it back?

Which e2e tests will it break?

PR #5701, opened 2026-03-02
chore(KFLUXVNGD-795): inject Segment write-key at build time

Omeramsc:
- Add build-time injection for the Segment write key.
- A new DefaultSegmentWritekey variable in pkg/version is set via the
SEGMENT_WRITE_KEY build-arg in the Containerfile and Makefile.
 The controller's reconcileSegmentBridgeSecret() resolves the key by CR
inline sprc.segmentKey > build-time default > skip (telemetry is
 disabled).
- on the pipeline side, a new read-segment-key inline task in
build-pipeline.yaml mounts the default-segment-key Secret as an optional
volume, and wires it into the buildah BUILD_ARGS for both build-images

### DIFF
--- a/.tekton/build-pipeline.yaml
+++ b/.tekton/build-pipeline.yaml
@@ -133,6 +133,37 @@ spec:
       workspaces:
         - name: basic-auth
           workspace: git-auth
+    # The Segment write key is a client-side analytics token (similar to a
+    # Google Analytics tracking ID). It only permits writing telemetry events
+    # and is baked into the public binary via ldflags, so exposing it through
+    # a Tekton result is acceptable.
+    - name: read-segment-key
+      runAfter:
+        - init
+      taskSpec:
+        results:
+          - name: segment-write-key
+            description: Segment write key read from the default-segment-key secret
+        volumes:
+          - name: segment-key
+            secret:
+              secretName: default-segment-key
+              optional: true
+        steps:
+          - name: read-secret
+            image: registry.access.redhat.com/ubi10/ubi-minimal@sha256:a74a7a92d3069bfac09c6882087771fc7db59fa9d8e16f14f4e012fe7288554c
+            volumeMounts:
+              - name: segment-key
+                mountPath: /var/run/secrets/segment
+                readOnly: true
+            script: |
+              #!/usr/bin/env bash
+              KEY_FILE="/var/run/secrets/segment/SEGMENT_WRITE_KEY"
+              if [ -f "$KEY_FILE" ]; then
+                tr -d '\n' < "$KEY_FILE" > "$(results.segment-write-key.path)"
+              else
+                printf '' > "$(results.segment-write-key.path)"
+              fi
     - name: prefetch-dependencies
       params:
         - name: input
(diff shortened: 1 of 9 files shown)


.tekton/build-pipeline.yaml
```diff
@@ -133,6 +133,33 @@ spec:
...
+              optional: true
+        steps:
+          - name: read-secret
+            image: registry.access.redhat.com/ubi10/ubi-minimal@sha256:a74a7a92d3069bfac09c6882087771fc7db59fa9d8e16f14f4e012fe7288554c
+            volumeMounts:
+              - name: segment-key
+                mountPath: /var/run/secrets/segment
+                readOnly: true
+            script: |
+              #!/usr/bin/env bash
```
gbenhaim, line 160: Worth adding a comment which explains why it's OK to expose the "secret" like that.

.tekton/build-pipeline.yaml
```diff
@@ -416,6 +445,7 @@ spec:
         - name: BUILD_ARGS
           value:
             - $(params.build-args[*])
+            - "SEGMENT_WRITE_KEY=$(tasks.read-segment-key.results.segment-write-key)"
```
gbenhaim: why do we need to pass the key to this task?

operator/pkg/version/version.go
```diff
@@ -17,10 +17,14 @@ limitations under the License.
...
+// Default values are "unknown" / empty and will be replaced during the build process.
 var (
 	// Version is the version of the operator (e.g., "v0.0.1")
 	Version = "unknown"
 	// GitCommit is the git commit SHA used to build the binary
 	GitCommit = "unknown"
+	// DefaultSegmentWriteKey is the Segment write key baked into production builds.
+	// Empty in development; set via -ldflags during container image builds.
+	// Used as a fallback when the KonfluxSegmentBridge CR does not specify a key.
+	DefaultSegmentWriteKey = ""
```
gbenhaim: This should be added in its own package since it's not related to version.

I also suggest to add a function for getting the value of this variable, so in the test you can pass a mock function.

gbenhaim: You'll include the function in the KonfluxSegmentBridgeReconciler struct, so in the test you can inject a mock.

operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
```diff
@@ -142,20 +144,37 @@ func (r *KonfluxSegmentBridgeReconciler) applyManifests(ctx context.Context, tc
...
 	log := logf.FromContext(ctx)
 
 	segmentKey := spec.GetSegmentKey()
+	keySource := "cr"
 	if segmentKey == "" {
-		log.Info("No Segment write key configured; skipping Secret creation")
+		segmentKey = r.GetDefaultSegmentKey()
+		keySource = "build-time-default"
+	}
+	if segmentKey == "" {
```
gbenhaim, line 168: You don't need to add logic for deletion, if you won't call the tc
ApplyOwned on the secret, tc will delete it.

operator/pkg/segment/segment.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package segment
+
+// DefaultWriteKey is the Segment write key baked into production builds.
+// Empty in development; set via -ldflags during container image builds.
+// Used as a fallback when the KonfluxSegmentBridge CR does not specify a key.
+var DefaultWriteKey = ""
```
gbenhaim: Now that you have a getter function, there is no need to export this variable

gbenhaim (approved):

PR #5768, opened 2026-03-04
refactor(KFLUXVNGD-806): Restructure tekton-chains-rbac for inheritance

kelchen123:
Restructure tekton-chains-rbac and tekton-chains-rbac-ocp kustomize components so that OCP inherits from a shared base, reducing repetition.

Assisted by: Cursor


dependencies/tekton-chains-rbac/kustomization.yaml
```diff
@@ -1,8 +1,56 @@
...
 
-images:
-  - name: quay.io/konflux-ci/appstudio-utils
-    digest: sha256:a8d6c26006f22e61f700b547dc589f2a0c6bccbcade7eddc34304d1d9d3f898b
+patches:
+  # Extend job to also create public-key secret in openshift-pipelines namespace
+  - target:
+      kind: Job
+      name: tekton-chains-signing-secret
+    patch: |-
```
amisstea, line 14: Rather than replace the entire script in each overlay (there are now 3 slightly different variations), you might be able to use `env` variables on the job in the base overlay to account for the subtle differences in the script and patch those instead.

kelchen123: done

PR #5835, opened 2026-03-06
Add optional Sigstore integration

gbenhaim:
Introduce integrations/sigstore/ with a self-contained install script that deploys the Sigstore stack (Fulcio, Rekor, CT Log, TUF, Trillian) using the scaffold Helm chart.

The script:
- Uses helm upgrade --install with a pinned chart version
- Applies minimal resource limits via a values.yaml
- Configures Fulcio with the Kubernetes service account issuer
- Enables TUF for root-of-trust distribution
- Patches the Konflux CR with in-cluster service URLs when present

Renovate is configured to track the scaffold chart version in install.sh and open PRs when new versions are released.

Sigstore is not a core dependency — it requires only helm and kubectl and is deployed independently of deploy-deps.sh.

In addition, add a script for verifying the signature of and image and its SBOM.

Assisted-By: Cursor

### DIFF
--- a/integrations/sigstore/verify.sh
+++ b/integrations/sigstore/verify.sh
@@ -1,0 +1,133 @@
+#!/bin/bash
+set -euo pipefail
+
+# Verifies an OCI image built by Konflux:
+#   1. Image signature  (cosign verify)
+#   2. SBOM attestation (cosign verify-attestation --type spdxjson)
+#
+# Because the Sigstore services are not exposed via ingress, this script
+# sets up kubectl port-forwards to Rekor and TUF before running cosign.
+#
+# Prerequisites: cosign, kubectl, curl
+#
+# Usage:
+#   ./integrations/sigstore/verify.sh <image-reference>
+#
+# Example:
+#   INSECURE_REGISTRY=true ./integrations/sigstore/verify.sh \
+#       localhost:5001/test-component:on-pr-abc123
+#
+# Environment variables (optional):
+#   CERTIFICATE_IDENTITY_REGEXP  - signing identity regexp (default: .*)
+#   CERTIFICATE_OIDC_ISSUER      - OIDC issuer (default: auto-detected)
+#   REKOR_PORT  - local port for Rekor  (default: 30800)
+#   TUF_PORT    - local port for TUF    (default: 30801)
+#   INSECURE_REGISTRY - "true" for insecure registries (default: false)
+
+IMAGE_REF="${1:-}"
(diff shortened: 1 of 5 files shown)


integrations/sigstore/verify.sh
```diff
@@ -0,0 +1,132 @@
...
+PORT_FORWARD_PIDS+=($!)
+kubectl port-forward -n tuf-system svc/tuf-server "${TUF_PORT}:80" &>/dev/null &
+PORT_FORWARD_PIDS+=($!)
+
+wait_for_port "${REKOR_PORT}" "Rekor"
+wait_for_port "${TUF_PORT}"   "TUF"
+
+# ---------- TUF init ----------
+
+export TUF_ROOT="$(mktemp -d)"
```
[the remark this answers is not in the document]
gbenhaim: done

gbenhaim:
@avi-biton @yftacherzog can you please review?

operator/konflux-cr.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: konflux
+spec:
+  imageController:
+    enabled: true
+  info:
+    spec:
+      clusterConfig:
+        data:
+          defaultOIDCIssuer: https://kubernetes.default.svc
+          fulcioInternalUrl: https://fulcio.gal.com
```
avi-biton: isn't `https://fulcio.gal.com` a personal domain?

integrations/sigstore/verify.sh
```diff
@@ -0,0 +1,133 @@
...
+        wait "$pid" 2>/dev/null || true
+    done
+}
+trap cleanup EXIT
+
+wait_for_port() {
+    local port="$1" name="$2"
+    echo "Waiting for ${name} on localhost:${port}..." >&2
+    for ((i = 1; i <= 30; i++)); do
+        if curl -s -o /dev/null "http://localhost:${port}/" 2>/dev/null; then
```
avi-biton:
```suggestion
        if curl -s -f -o /dev/null "http://localhost:${port}/" 2>/dev/null; then
```
curl can exit 0 even if we get 404 which will cause the test to pass even if the service returns an error.
Adding -f will fail responses that are not 2xx

operator/konflux-cr.yaml
gbenhaim: good catch. I committed this file by mistake.

integrations/sigstore/verify.sh
gbenhaim: done

PR #5866, opened 2026-03-09
test: maintain e2e tests locally

yftacherzog:
Before, we were running e2e tests implemented in an image built from a separate repo. With this change, we are moving the e2e tests to this repo, and running them locally. This should provide better maintainability and faster feedback loops.

Assisted-by: Cursor


.github/workflows/operator-test-e2e.yaml
```diff
@@ -252,7 +253,17 @@ jobs:
...
+          go test .
+
+      - name: Set release catalog revision
+        if: matrix.test_scope == 'integration+e2e'
+        run: grep '^RELEASE_SERVICE_CATALOG_REVISION=' test/e2e/release-service-catalog-revision >> $GITHUB_ENV
+
+      - name: Set build pipeline bundle (min)
+        if: matrix.test_scope == 'integration+e2e'
+        run: |
+          BUNDLE=$(yq eval-all 'select(.kind == "ConfigMap" and .metadata.name == "build-pipeline-config") | .data["config.yaml"]' operator/pkg/manifests/build-service/manifests.yaml | yq eval '.pipelines[] | select(.name == "docker-build-oci-ta-min") | .bundle' -)
```
gbenhaim, line 267: having this logic here will make it hard to run the e2e test locally.

.github/workflows/operator-test-e2e.yaml
```diff
@@ -252,18 +253,37 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          MY_GITHUB_ORG: ${{ secrets.GH_ORG }}
           QUAY_DOCKERCONFIGJSON: ${{ secrets.QUAY_DOCKERCONFIGJSON }}
           RELEASE_CATALOG_TA_QUAY_TOKEN: ${{ secrets.RELEASE_CATALOG_TA_QUAY_TOKEN }}
+          E2E_APPLICATIONS_NAMESPACE: user-ns2
+          TEST_ENVIRONMENT: upstream
         run: |
-          ./test/e2e/run-e2e.sh
+          cd test/go-tests
+          # -ginkgo.vv forces full spec output (like local TTY); -ginkgo.no-color for CI logs
```
gbenhaim: There is output mode for github actions

.github/workflows/operator-test-e2e.yaml
```diff
@@ -252,18 +253,37 @@ jobs:
         working-directory: ${{ github.workspace }}
         run: |
           cd test/go-tests
-          go test ./...
+          go test .
```
gbenhaim: Which tests are we running here? Isn't it the same as the tests triggered below?

yftacherzog: `go test .` will runs tests in the current package (the one in the current directory). So this is not the same as the ones below, but I think it means it made us not trigger some other tests in subdirs. I changed it now.

.github/workflows/operator-test-e2e.yaml
yftacherzog: I wanted to remove the scripts for running tests completely. I added a template for an env file that can be sourced locally instead (it uses grep instead of yq so we don't introduce another dependency).

I was actually working on getting it to pass both locally and in CI.

PR #5871, opened 2026-03-09
fix(segment-bridge): Writable tmpdir

ifireball:
Mount an empty dir for the segment-bridge job so it can write to `/tmp`.

Assisted-by: Gemini (via Cursor)
Signed-off-by: Barak Korren <bkorren@redhat.com>


gbenhaim:
Thanks @Omeramsc 
So I guess we can close this pr since the required change to the manifests will be done I the segment repo?

PR #5876, opened 2026-03-09
fix(KFLUXVNGD-820): Add fallback to main branch for rehearsal jobs

kelchen123:
Implementing fallback that doesnt use PULL_PULL_SHA as it will not work for non pr jobs.

### DIFF
--- a/deploy-konflux-on-ocp.sh
+++ b/deploy-konflux-on-ocp.sh
@@ -12,8 +12,12 @@
 #   - OpenShift cluster with OperatorHub access
 #
 # Environment variables (optional):
-#   OPERATOR_IMAGE            - Full operator image to use (skips SHA-based construction)
+#   OPERATOR_IMAGE        - Full operator image to use (skips SHA-based construction)
 #   KONFLUX_OPERATOR_REPO - Operator image repository (default: quay.io/redhat-user-workloads/konflux-vanguard-tenant/konflux-operator)
+#
+# Environment variables (set by OpenShift CI/Prow):
+#   REPO_NAME      - Name of the repository being tested (e.g., "konflux-ci", "release")
+#   PULL_PULL_SHA  - Git SHA of the PR head commit being tested
 #
 
 set -o nounset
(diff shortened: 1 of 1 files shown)


deploy-konflux-on-ocp.sh
```diff
@@ -45,13 +45,14 @@ if [[ -n "${OPERATOR_IMAGE:-}" ]]; then
...
+    # Only use PULL_PULL_SHA if this is actually a PR to konflux-ci (not rehearsal jobs)
+    if [[ "${REPO_NAME:-}" == "konflux-ci" && -n "${PULL_PULL_SHA:-}" ]]; then
         COMMIT_SHA="${PULL_PULL_SHA}"
         echo "Using PR head SHA (PULL_PULL_SHA): ${COMMIT_SHA}"
     else
-        COMMIT_SHA=$(git rev-parse HEAD)
-        echo "Using git HEAD SHA: ${COMMIT_SHA}"
+        # Rehearsal, periodic, or local testing - use main branch (always has an image)
+        COMMIT_SHA=$(git ls-remote https://github.com/konflux-ci/konflux-ci.git refs/heads/main | cut -f1)
+        echo "Using main branch SHA: ${COMMIT_SHA}"
```
amisstea: I don't think this will always work. The commit SHA is being used to compute the `on-pr-{sha}` tag below. These tags auto expire after 5 days.

Can we not just use the `quay.io/konflux-ci/konflux-operator:latest` image as a reliable default instead?

deploy-konflux-on-ocp.sh
```diff
@@ -45,13 +45,14 @@ if [[ -n "${OPERATOR_IMAGE:-}" ]]; then
     echo "Using provided OPERATOR_IMAGE: ${OPERATOR_IMAGE}"
 else
     # Construct image tag from commit SHA
-    # In CI (Prow), use PULL_PULL_SHA (PR head) instead of HEAD (merge commit)
-    if [[ -n "${PULL_PULL_SHA:-}" ]]; then
+    # Only use PULL_PULL_SHA if this is actually a PR to konflux-ci (not rehearsal jobs)
+    if [[ "${REPO_NAME:-}" == "konflux-ci" && -n "${PULL_PULL_SHA:-}" ]]; then
```
amisstea, line 52: These env variables aren't documented anywhere in this script. I assume they originate from openshift-CI but that's non-obvious. Please make sure all env variables are documented at the top of the script.

PR #5906, opened 2026-03-10
fix: segment-bridge CronJob runtime configs

Omeramsc:
The CronJob deploys, but could not run. it lacked the Tekton Results API address and could not enable telemtry. this change adds the missing TEKTON_RESULTS_API_ADDR to the controller-managed Secret and wire SEGMENT_WRITE_KEY through deploy-local.sh for local testing.

assisted-by: cursor

### DIFF
--- a/konflux-ci/enterprise-contract/core/kustomization.yaml
+++ b/konflux-ci/enterprise-contract/core/kustomization.yaml
@@ -13,7 +13,7 @@ configMapGenerator:
   - name: ec-defaults
     namespace: enterprise-contract-service
     literals:
-      - verify_ec_task_bundle=quay.io/enterprise-contract/ec-task-bundle@sha256:080172276d25afea2e90aaba8b24734b6243e8e3ff0ca8994c14cece263278e0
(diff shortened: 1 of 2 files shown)


operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
```diff
@@ -51,6 +51,11 @@ const (
 
 	segmentBridgeNamespace  = "segment-bridge"
 	segmentBridgeSecretName = "segment-bridge-config"
+
+	// defaultTektonResultsAPIAddr is the in-cluster gRPC address of the Tekton
+	// Results API. Matches the service created by the Tekton Operator when
+	// TektonConfig has result.disabled=false with targetNamespace=tekton-pipelines.
+	defaultTektonResultsAPIAddr = "tekton-results-api-service.tekton-pipelines.svc.cluster.local:8080"
```
gbenhaim: this would be different when using `openshift-pipelines`

Omeramsc: hmm...  Should we have the controller discover it at reconcilation time by reading the TektonConfig CR?

gbenhaim: We have the ability to discover if we run on openshift (you can see it in the ui controller). You can use it for determining the right endpoint.

PR #5928, opened 2026-03-11
fix: kyverno startup failure

yftacherzog:
Creation of Kyverno policies was some times failing during setup, as the admission controller was not ready in time.

This change replaces the fixed wait time with a retry that verifies the admission controller is ready.

Assisted-by: Cursor


gbenhaim (approved):

yftacherzog:
> We need to check if we can remove kyverno entirely from our dependencies

We're still using it for a few things. We'll need to review this carefully, as we're using it, and there were also some Konflux services that were using it, if I recall correctly.

PR #5951, opened 2026-03-12
deploy-deps.sh: add failure logging to show completed steps

### LINKED ISSUE #5947: deploy-deps.sh: No cleanup when deployment fails partway
When `deploy-deps.sh` fails midway through deployment (e.g., cert-manager succeeds but Tekton times out), there's no cleanup of the resources that were already created.

**Current behavior:**
- Script exits on first failure due to `set -eu`
- Previously deployed components remain in the cluster
- User has to manually figure out what succeeded and clean it up

**Expected behavior:**
- Track what's been deployed
- On failure, offer option to rollback or at least print what was deployed

**Why this matters:**
Running the script again after a partial failure can cause conflicts or unexpected state. It's especially frustrating when debugging on a slow machine where you might hit timeouts more often.

I noticed `scripts/setup-kind-local-cluster.sh` already uses `trap ... EXIT` for cleanup in some cases - a similar pattern could work here.

**Relevant code:**
- `deploy-deps.sh` has `set -eu -o pipefail` (line ~48) but no trap handler
- Multiple components deployed sequentially: cert-manager, tekton, pipelines-as-code, kyverno, registry, etc.

harshakumar25:
When `deploy-deps.sh` fails partway through, users currently have to scroll back through logs to figure out what succeeded. This adds a simple exit trap that prints:

- Which step was running when failure occurred
- Which steps completed successfully before that

Example output on failure:
```
❌ Deployment failed during: Tekton
✅ Completed steps:
   - Cert Manager
   - Trust Manager
   - Cluster Issuer

💡 The script is idempotent - you can re-run it after fixing the issue.
```

Keeps the script maintainable - just tracking step names in the existing `deploy()` function, no changes to individual deployment functions.

Fixes #5947

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -2,6 +2,29 @@
 
 script_path="$(dirname -- "${BASH_SOURCE[0]}")"
 
+# Track deployment progress for failure reporting
+COMPLETED_STEPS=()
+CURRENT_STEP=""
+
+# Print deployment status on failure
+print_failure_status() {
+    local exit_code=$?
+    if [[ $exit_code -ne 0 && -n "$CURRENT_STEP" ]]; then
+        echo "" >&2
+        echo "❌ Deployment failed during: $CURRENT_STEP" >&2
+        if [[ ${#COMPLETED_STEPS[@]} -gt 0 ]]; then
+            echo "✅ Completed steps:" >&2
+            for step in "${COMPLETED_STEPS[@]}"; do
+                echo "   - $step" >&2
+            done
+        fi
+        echo "" >&2
+        echo "💡 The script is idempotent - you can re-run it after fixing the issue." >&2
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


deploy-deps.sh
```diff
@@ -69,26 +92,58 @@ check_req(){
 }
 
 deploy() {
+    CURRENT_STEP="Cert Manager"
     echo "🔐 Deploying Cert Manager..." >&2
     deploy_cert_manager
+    COMPLETED_STEPS+=("Cert Manager")
```
```suggestion
    COMPLETED_STEPS+=("$CURRENT_STEP")
```
Can we do something like this to keep it more maintainable?

yftacherzog (approved):

harshakumar25:
hey @yftacherzog ,
i think everything is good , please merge this 
Thanks !

PR #5952, opened 2026-03-12
feat(ISV-6683): add TPA OIDC secret to managed namespace in backwards compatible format

BorekZnovustvoritel:
Assisted-by: Claude-4.6-Opus-High

* Requires https://github.com/redhat-appstudio/tsf-cli/pull/67 to be merged first
* Tested by running the script on a custom cluster

### DIFF
--- a/operator/hack/setup-release.sh
+++ b/operator/hack/setup-release.sh
@@ -279,10 +279,11 @@ EOF
 
 # Step 10: Copy SSO credentials from tpa-realm-clients secret in tsf namespace
 SSO_SECRET_CREATED=false
-if kubectl get namespace tsf &>/dev/null; then
+SSO_ACCOUNT=release
+SSO_TOKEN=$(kubectl get secret tpa-realm-clients -n tsf \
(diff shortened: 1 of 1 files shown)


operator/hack/setup-release.sh
```diff
@@ -277,7 +277,29 @@ subjects:
...
+    SSO_ACCOUNT=release
+    SSO_TOKEN=$(kubectl get secret tpa-realm-clients -n tsf -o jsonpath="{.data.$SSO_ACCOUNT}")
+    kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Secret
+metadata:
+  name: release-sso-secret
+  namespace: ${MANAGED_NS}
+type: Opaque
+data:
```
gbenhaim, line 296: Maybe use stringData here instead of using base64

gbenhaim: Why did you use echo and not just expending the variable?

BorekZnovustvoritel: My bad, I focused mainly on the data/stringData accuracy and forgot to simplify this.

PR #5953, opened 2026-03-12
deploy-deps.sh: make retry count and sleep configurable

### LINKED ISSUE #5949: deploy-deps.sh: Make retry/sleep values configurable
The `retry()` function in `deploy-deps.sh` uses hardcoded values that can't be adjusted:

```bash
retry() {
    for i in {1..3}; do
        ...
        sleep 3
    done
}
```

There are also various `sleep 5` and `sleep 10` calls scattered through the script with only one comment explaining why (referencing issue #161).

**Why this is a problem:**
- On slower machines or CI with resource constraints, 3 retries with 3s sleep might not be enough
- On fast machines, the sleeps are just wasted time
- The magic numbers make it hard to understand what tolerance the script actually needs

**Suggestion:**
Make these configurable via env vars with sensible defaults:

```bash
RETRY_COUNT="${RETRY_COUNT:-3}"
RETRY_SLEEP="${RETRY_SLEEP:-3}"
```

And maybe add comments for the standalone sleeps explaining what they're waiting for (like the one that already references #161).

Not a huge deal but would make local dev less frustrating when things are timing out.

harshakumar25:
## Summary
Add `RETRY_COUNT` and `RETRY_SLEEP` environment variables to allow customization of retry behavior.

## Changes
- Add env vars with sensible defaults (3 retries, 3s sleep)
- Update `retry()` function to use these values dynamically

## Usage
```bash
# More retries for slow CI
RETRY_COUNT=5 RETRY_SLEEP=10 ./deploy-deps.sh

# Faster for local dev
RETRY_COUNT=2 RETRY_SLEEP=1 ./deploy-deps.sh
```

Fixes #5949


harshakumar25:
hi @rrosatti ,
i have implemeted changes and fixed issue ,if everything looks good i will glade to get lgtm.
Thanks !

yftacherzog (commented):
I'm not sure how useful it would be, but it's clean, and I don't think it's doing any harm, so I'm ok with merging it.
@gbenhaim @avi-biton can you have a look?

yftacherzog (approved):

yftacherzog (approved):

yftacherzog:
> Thanks @yftacherzog for approval , i will glade if it can get merge Thanks !

I realize now that the operator e2e is not tested in CI for this change.
Would you mind fixing something small so the change is verified against the operator e2e tests?

We need to include deploy-deps.sh [here](https://github.com/konflux-ci/konflux-ci/blob/main/.github/workflows/operator-test-e2e.yaml#L87).

yftacherzog:
> > Thanks @yftacherzog for approval , i will glade if it can get merge Thanks !
> 
> I realize now that the operator e2e is not tested in CI for this change. Would you mind fixing something small so the change is verified against the operator e2e tests?
> 
> We need to include deploy-deps.sh [here](https://github.com/konflux-ci/konflux-ci/blob/main/.github/workflows/operator-test-e2e.yaml#L87).

@harshakumar25 I believe this is not addressed yet. We cannot test this PR without it.

PR #5954, opened 2026-03-12
docs: GitHub app installation

yftacherzog:
Make the GitHub App installation steps more coherent. Later commits will walk through the next installation steps.

Assisted-by: Cursor

### DIFF
--- a/operator/docs/content/docs/installation/install-local.md
+++ b/operator/docs/content/docs/installation/install-local.md
@@ -9,57 +9,54 @@ This guide walks you through deploying Konflux locally on macOS or Linux using
 [Kind](https://kind.sigs.k8s.io/). The automated `deploy-local.sh` script handles
 cluster creation, operator deployment, and GitHub integration in a single step.
 
+It relies on sourcing `deploy-local.env` file which provides it with environment
+variables. The following steps guide you through setting up the env file and running
+the script.
+
 ## Prerequisites
 
-Verify that the following tools are installed:
+- Verify that the following tools are installed:
 
-| Tool | Minimum version |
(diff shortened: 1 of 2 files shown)


operator/docs/content/docs/installation/install-local.md
```diff
@@ -9,57 +9,53 @@ This guide walks you through deploying Konflux locally on macOS or Linux using
...
-Verify that the following tools are installed:
+- Verify that the following tools are installed:
 
-| Tool | Minimum version |
-|------|----------------|
-| [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) | v0.26.0 |
-| [podman](https://podman.io/docs/installation) or [docker](https://docs.docker.com/engine/install/) | podman v5.3.1 / docker v27.0.1 |
-| [kubectl](https://kubernetes.io/docs/tasks/tools/) | v1.31.4 |
-| [git](https://git-scm.com/) | v2.46 |
-| [openssl](https://www.openssl.org/) | v3.0.13 |
```
avi-biton, line 22: Why was it removed?

yftacherzog: The pac tutorial moved from openssl to `head -c 30 /dev/random | base64` which does not require openssl. Using it in our docs would reduce a dependency and also make it easier to understand that what is being suggested there is also what we were mentioning in our doc.

I had some recollection that we depend on it in someplace else but did not find it, and now Qodo found it -- depoy-deps still uses it, so I reinstated it in the doc.

avi-biton (approved):

PR #5965, opened 2026-03-12
refactor: use setup-release.sh for E2E release config instead of manual Go setup

gbenhaim:
Replace the hand-rolled release configuration in conformance tests (createReleaseConfig, createE2EQuaySecret) with a shell-out to operator/hack/setup-release.sh. The script uses ImageRepository CRs managed by image-controller to provision Quay push credentials automatically, eliminating the need for QUAY_DOCKERCONFIGJSON, RELEASE_CATALOG_TA_QUAY_TOKEN, and RELEASE_TA_OCI_STORAGE env vars.

Key changes:
- Add --catalog-revision / -R flag to setup-release.sh
- Replace createReleaseConfig/createE2EQuaySecret with runSetupRelease()
- Use default-tenant namespace (operator-created) instead of user-ns2
- Use push-to-external-registry pipeline and default EC policy
- Remove Quay credential env vars from CI workflows and env template

Assisted-By: Cursor
Made-with: Cursor


gbenhaim:
The legacy sanity test is expected to fail since this change expects the cluster to have the `default-tenant` namespace.

yftacherzog:
> The legacy sanity test is expected to fail since this change expects the cluster to have the `default-tenant` namespace.

@gbenhaim What does it mean in terms of how this PR should be handled?

gbenhaim:
replaced by https://github.com/konflux-ci/konflux-ci/pull/6396

Issue #5948, opened 2026-03-12
Track deprecated scripts/workflows scheduled for removal


harshakumar25:
Found several scripts and workflows marked deprecated with "will be removed in a future release" but there's no tracking anywhere for when they'll actually be removed:

**Deprecated items without removal plan:**
- `deploy-konflux.sh` - says "will be removed in a future release"
- `deploy-image-controller.sh` - same message
- `.github/workflows/sanity-test.yml` - "will be removed once all environments migrate"
- `dependencies/kyverno/policy/e2e-reduce-resources.yaml` - "will be removed in the future"

**The problem:**
Without any timeline or tracking issue, these just sit there indefinitely. Contributors might not realize they're working on code that's going away, and the codebase accumulates cruft over time.

**Suggestion:**
Either:
1. Create tracking issues for each (with rough timeline like "Q2 2026" or "after v2.0")
2. Add a `deprecations.md` that lists everything pending removal
3. Use GitHub labels like "pending-removal" to make it searchable

Happy to help with any of these approaches if there's a preference!

harshakumar25:
@yftacherzog Thanks for the guidance! I've updated the issue description to include a checklist with the known blockers for each deprecated item.

I noticed that `e2e-reduce-resources.yaml` is not referenced anywhere anymore and appears safe to remove today, so I went ahead and opened #5976 to clear it out.

PR #5976, opened 2026-03-13
chore: remove deprecated e2e-reduce-resources policy

harshakumar25:
## Description

This PR removes the `e2e-reduce-resources` kyverno policy as part of #5948 tracking.

This policy is explicitly marked as deprecated, and its namespace matcher contains `NON_EXISTING_LABEL-NOT_REQUIRED_ANYMORE`. My search confirms no other resources or tests have dependencies on it, so it is safe to remove immediately.

Related to #5948


yftacherzog:
@harshakumar25 this change seems to be based on top of a PR used for another PR. I think you could do interactive rebase on top of the latest upstream main branch and keep only the relevant commit.

harshakumar25:
sure @yftacherzog ,
i will do this shortly.
Thanks !

harshakumar25:
Hi @yftacherzog, rebased on latest upstream main and force-pushed. This PR now has only the single relevant commit for #5976 (deprecated policy removal). Please re-review.
Thanks !

yftacherzog (changes requested):
This change is indeed something we need to do, but we need to make sure it's not being referenced also downstream. E.g. need to make sure Fedora is not referencing it.

yftacherzog (approved):

PR #5981, opened 2026-03-15
fix(deploy-local): update template

avi-biton:
Updating the precedence of the variables in the deploy-local.sh.
The script now follows the precedence of the variables.
1 - Injected env vars (CLI)
2 - Values in this file (deploy-local.env)
3 - Script defaults

### DIFF
--- a/scripts/deploy-local.env.template
+++ b/scripts/deploy-local.env.template
@@ -7,6 +7,26 @@
 # in your environment (useful for CI/automated deployments).
 #
 # IMPORTANT: deploy-local.env is in .gitignore - never commit secrets!
+#
+# ==============================================================================
+# Variable Precedence (high to low)
+# ==============================================================================
+#
+#   1. Injected env vars (CLI)  -  highest priority, always wins
+#        OPERATOR_INSTALL_METHOD=none ./scripts/deploy-local.sh
+#
+#   2. Values in this file (deploy-local.env)
+#        Sourced by the script; only takes effect when the variable was NOT
+#        already provided on the command line.
+#
+#   3. Script defaults  -  lowest priority
+#        Fallback values defined inside deploy-local.sh with ${VAR:-default}.
+#        Applied only when the variable is unset or empty after steps 1 and 2.
+#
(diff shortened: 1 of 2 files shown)


scripts/deploy-local.env.template
```diff
@@ -22,7 +22,14 @@
...
 #   git checkout v1.0.0 && OPERATOR_INSTALL_METHOD=local ./scripts/deploy-local.sh
-OPERATOR_INSTALL_METHOD=release
+#
+# NOTE: This variable is intentionally left commented out. The script defaults to
+# 'release' when it is unset (see deploy-local.sh). If you set it explicitly here,
+# it will override any value passed on the command line (e.g.
+# OPERATOR_INSTALL_METHOD=none ./scripts/deploy-local.sh), because the env file is
+# sourced inside the script and overwrites CLI-passed environment variables.
+# Uncomment and set only if you want this file to always enforce a specific method.
+# OPERATOR_INSTALL_METHOD=release
```
yftacherzog: What's special about this variable? This is the same for any other variable in this template. No?

I think we had things mentioned around the precedence of how env vars are provided. If we do, we need to make sure that we act upon it.

avi-biton: I think that in all our scripts, the precedence should be:
1 - Injecting Env var should come up first
2 - env vars in anv files should be second 
3 - vars in Script

Meaning that if we run the script with overriding vars like:

- `OPERATOR_INSTALL_METHOD=none ./scripts/deploy-local.sh` -> `OPERATOR_INSTALL_METHOD` should be none - no matter what the env file value is.
- If we run it without overriding values, the value should be taken from the env file.
- And finally we have the defaults in the script

yftacherzog: So the env file could have this assignment with default value that we use in some scripts. No?
That would be better than commenting a line in and out.

avi-biton: Yes, I will take more general approach than handling each variable.
This PR was opened because I hit it in the installation docs

yftacherzog (approved):

PR #5997, opened 2026-03-16
chore: onboard to codecov

lipka28:
Uploads test results to codecov. This is part of a wider onboarding of all the konflux-ci components to upload to codecov.

### DIFF
--- a/.github/workflows/operator-test.yaml
+++ b/.github/workflows/operator-test.yaml
@@ -1,6 +1,8 @@
 name: Operator Tests
 
 on:
+  push:
+    branches: [main]
   pull_request:
   merge_group:
     types: [checks_requested]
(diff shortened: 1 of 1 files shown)


yftacherzog (dismissed):

.github/workflows/operator-test.yaml
```diff
@@ -1,6 +1,8 @@
 name: Operator Tests
 
 on:
+  push:
+    branches: [main]
```
yftacherzog, line 5: What's the reason for the additional trigger?

lipka28: The reason for trigger on push is that codecove needs it to estabilish a baseline of the main branch to be later able to calculate acurate covarage dif on PRs. Without it the data would not be accurate.

yftacherzog (approved):

PR #6031, opened 2026-03-17
feat: enable e2e conformance tests on OpenShift deployments

amisstea:
Support running the e2e conformance test suite against the operator when deployed on OpenShift.

Operator PAC_WEBHOOK_URL auto-detection:

On upstream/Kind clusters, Pipelines-as-Code runs in the "pipelines-as-code" namespace on port 8180. On OpenShift, PaC is deployed by the OpenShift Pipelines operator into the "openshift-pipelines" namespace on port 8080. The build-service controller uses PAC_WEBHOOK_URL to POST to PaC's incoming webhook endpoint when retriggering builds, so the wrong URL causes retrigger failures (DNS lookup fails). The operator now auto-detects OpenShift via ClusterInfo and sets the correct PAC_WEBHOOK_URL default. User- provided env overrides in the KonfluxBuildService CR still take precedence via strategic merge.

Smee client OpenShift compatibility:

The smee-client deployment forwards GitHub webhooks from smee.io to the in-cluster PaC controller. On OpenShift, PaC lives in a different namespace and port than upstream, so the health-check sidecar's DOWNSTREAM_SERVICE_URL must be patched to target
openshift-pipelines:8080. The smee-client manifest also had hardcoded runAsUser/fsGroup values that conflict with OpenShift's restricted SCC, so those are set to null to let OpenShift assign them.

Script refactoring:

deploy-secrets.sh and resolve-konflux-cr.sh were extracted from deploy-local.sh so they can be shared with deploy-konflux-on-ocp.sh, which needs the same secret creation and CR resolution logic when deploying on OpenShift.

Assisted-by: Claude claude-opus-4-6

### DIFF
--- a/dependencies/smee/smee-client.yaml
+++ b/dependencies/smee/smee-client.yaml
@@ -23,7 +23,7 @@ spec:
         - name: shared-health
           emptyDir: {}
       securityContext:
-        fsGroup: 65532
+        fsGroup: null
       containers:
         - image: ghcr.io/chmouel/gosmee:latest
           imagePullPolicy: Always
@@ -38,7 +38,7 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
             runAsNonRoot: true
-            runAsUser: 65532
+            runAsUser: null
           resources:
             limits:
               cpu: 100m
@@ -85,7 +85,7 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
             runAsNonRoot: true
-            runAsUser: 65532
+            runAsUser: null
           resources:
             limits:
               cpu: 100m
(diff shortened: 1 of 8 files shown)


dependencies/smee/smee-client.yaml
```diff
@@ -23,7 +23,7 @@ spec:
         - name: shared-health
           emptyDir: {}
       securityContext:
-        fsGroup: 65532
+        fsGroup: null
```
gbenhaim, line 26: Can we just remove it?

gbenhaim (approved):

dependencies/smee/smee-client.yaml
amisstea: Probably. I'll open a followup PR to remove it.

PR #6039, opened 2026-03-18
feat: add custom CA bundle support for image-controller

gbenhaim:
Enable image-controller to communicate with self-hosted Quay registries that use custom CA certificates. Users configure this through the top-level Konflux CR, which propagates the spec to the KonfluxImageController sub-CR.

Assisted-By: Cursor


operator/internal/controller/imagecontroller/konfluximagecontroller_controller.go
```diff
@@ -156,6 +172,44 @@ func (r *KonfluxImageControllerReconciler) applyManifests(ctx context.Context, t
 	return nil
 }
 
+// applyImageControllerDeploymentCustomizations applies user-defined customizations to ImageController deployments.
+func applyImageControllerDeploymentCustomizations(deployment *appsv1.Deployment, spec konfluxv1alpha1.KonfluxImageControllerSpec) error {
+	switch deployment.Name {
```
yftacherzog, line 177: Are we expected to ever have additional cases here? switch statement feels a bit strange here. No?

gbenhaim: I don't know what will happen in the future. This is the pattern we use in other reconcilers, so I prefer to keep like this.

operator/pkg/manifests/image-controller/manifests.yaml
```diff
@@ -972,7 +972,7 @@ spec:
         - --health-probe-bind-address=:8081
         command:
         - /manager
-        image: quay.io/konflux-ci/image-controller:6e77d754a97567b71fd53cc6dc02f98e305bd2e0
+        image: quay.io/konflux-ci/image-controller:75bb86fc29324d794f76d14454e0b2c56326a8bc
```
yftacherzog: `75bb86fc29324d794f76d14454e0b2c56326a8bc` does not exist. Maybe release failure.

yftacherzog (approved):

operator/pkg/manifests/image-controller/manifests.yaml
gbenhaim: I use another commit which includes the change we need.

gbenhaim:
Verified locally that image-controller was able to connect to a self hosted quay using self-signed cert.

PR #6044, opened 2026-03-18
fix(operator): always create segment-bridge-config Secret

Omeramsc:
The operator previously skipped creating the segment-bridge-config Secret when no Segment write key was resolved, causing the CronJob to fall back to the Dockerfile default of localhost:50051 for TEKTON_RESULTS_API_ADDR and crash. This only affects local development — production builds always have the key baked in via ldflags. The Secret is now always created with the correct in-cluster Tekton Results API address, with SEGMENT_WRITE_KEY and SEGMENT_BATCH_API set to empty strings when no key is available. The companion segment-bridge scripts change handles the empty key gracefully by skipping the upload step.

Made-with: Cursor

### DIFF
--- a/operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
+++ b/operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
@@ -63,11 +63,9 @@ const (
 
 // SegmentBridgeCleanupGVKs defines which resource types should be cleaned up when they are
 // no longer part of the desired state. Only optional/conditional resources are listed here.
-// Always-applied resources don't need cleanup (they're always tracked and never become orphans).
-var SegmentBridgeCleanupGVKs = []schema.GroupVersionKind{
-	// Secret is conditional - only created when a Segment write key is configured
-	{Group: "", Version: "v1", Kind: "Secret"},
-}
+// All segment-bridge resources (including the Secret) are always applied, so no cleanup
+// candidates exist.
+var SegmentBridgeCleanupGVKs []schema.GroupVersionKind
 
 // SegmentBridgeClusterScopedAllowList restricts which cluster-scoped resources can be deleted
 // during orphan cleanup. All cluster-scoped resources managed by this controller are always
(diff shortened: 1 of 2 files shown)


operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
```diff
@@ -169,24 +167,34 @@ func (r *KonfluxSegmentBridgeReconciler) tektonResultsAPIAddr() string {
...
+	data := map[string]string{
+		"TEKTON_RESULTS_API_ADDR": r.tektonResultsAPIAddr(),
+		"SEGMENT_WRITE_KEY":       segmentKey,
+		"SEGMENT_BATCH_API":       "",
 	}
 
-	batchURL, err := url.JoinPath(spec.GetSegmentAPIURL(), "batch")
-	if err != nil {
-		return fmt.Errorf("invalid segment API URL: %w", err)
+	if segmentKey != "" {
```
gbenhaim: Why calculating the batch api path depends on the existence of the key?

operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller_test.go
```diff
@@ -110,7 +110,7 @@ var _ = Describe("KonfluxSegmentBridge Controller", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
-		It("should skip Secret creation when no write key is configured (CR empty, no build-time default)", func() {
+		It("should create Secret with only TEKTON_RESULTS_API_ADDR when no write key is configured", func() {
```
gbenhaim: I dont think it's accurate. The key field will be in the secret, it's value will be empty.

operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller.go
Omeramsc: I thought no need to compute the URL if a key is missing, but we can drop it for the sake of uniformity.

operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller_test.go
Omeramsc: changed

PR #6048, opened 2026-03-18
Update conforma policies

lcarva:
This change updates the `default` conforma policy to a more strict version that is still easily achievable when onboarding to Konflux. In addtion to include the more comprehensive `@redhat` collection, it also uses a custom set of rule data that fulfills the more generic use cases.

This change also removes all the other ECPs as they are not currently used.

### DIFF
--- a/test/go-tests/tests/conformance/setup.go
+++ b/test/go-tests/tests/conformance/setup.go
@@ -7,8 +7,8 @@ import (
 	"os"
 	"time"
 
-	buildcontrollers "github.com/konflux-ci/build-service/controllers"
 	ecp "github.com/conforma/crds/api/v1alpha1"
+	buildcontrollers "github.com/konflux-ci/build-service/controllers"
 	tektonutils "github.com/konflux-ci/release-service/tekton/utils"
 
 	"github.com/konflux-ci/konflux-ci/test/go-tests/pkg/clients/kube"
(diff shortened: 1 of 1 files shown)


konflux-ci/enterprise-contract/policies/enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_all.yaml
yftacherzog, line 1: the `konflux-ci` dir is used for the non-operator deployment.
There are still some users for this, specifically the Fedora cluster.
We need to be extra careful making sure nothing is actually using this there.

Can we limit the change only to operator?

lcarva: Of course! Happy to do so if it's ok to deviate those ECPs.

lcarva: Done!

gbenhaim (approved):

operator/upstream-kustomizations/enterprise-contract/policies/kustomization.yaml
```diff
@@ -1,11 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_all.yaml
   - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_default.yaml
-  - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_redhat-no-hermetic.yaml
-  - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_redhat-trusted-tasks.yaml
-  - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_redhat.yaml
-  - enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_slsa3.yaml
```
yftacherzog, line 9: We remove the references inside the kustomization file, but we don't remove the files that we stopped referencing. I think we should remove them as well, as this is dead code now.

avi-biton: If by that you are referring to remove them from a deployed operator after upgrade, then we can update the `EnterpriseContractCleanupGVKs` to clean those resources
https://github.com/konflux-ci/konflux-ci/blob/main/operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller.go#L56

We can update it to something like:
```
var EnterpriseContractCleanupGVKs = []schema.GroupVersionKind{
    {
        Group:   "appstudio.redhat.com",
        Version: "v1alpha1",
        Kind:    "EnterpriseContractPolicy",
    },
}
```

yftacherzog: I'm saying that this dir: https://github.com/lcarva/konflux-ci/tree/34c86c8d196d0a8354711e4c300ce91b80c9ebc8/operator/upstream-kustomizations/enterprise-contract/policies now contains yamls that are not referenced by the kustomization.yaml in that file.

avi-biton: They were originally removed, we probably got them back in one of the rebases

lcarva: Yeah my bad... let me fix that.

lcarva: Fixed it.

PR #6122, opened 2026-03-22
fix: bump image and adds list on namespaces without resourceNames

Omeramsc:
bump the image version of segment-bridge, and Adds a `list` verb on namespaces without `resourceNames restriction, so
`fetch-namespace-records.sh can query renant namespaces by label.

both were indentified as "gaps" during e2e deployment.

### DIFF
--- a/konflux-ci/ui/core/kustomization.yaml
+++ b/konflux-ci/ui/core/kustomization.yaml
@@ -9,6 +9,6 @@ images:
   - name: quay.io/konflux-ci/workspace-manager
     digest: sha256:d72968158bec10442236d4f84a904beeff5a022f3e032ecdd2d421651c01f8d3
   - name: quay.io/konflux-ci/konflux-ui
-    digest: sha256:6fdd17c9c8815dc8dec0caf80b63808eec2066ca0f1869938f4860f41f56bc88
+    digest: sha256:ccd4a980791307b0fe6b177e64eaccf0574a91b13f5685323c6456c33e7b749c
 
 namespace: konflux-ui
--- a/operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml
+++ b/operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml
@@ -7,7 +7,7 @@ resources:
 images:
 - name: quay.io/konflux-ci/konflux-ui
   newName: quay.io/konflux-ci/konflux-ui
-  digest: sha256:6fdd17c9c8815dc8dec0caf80b63808eec2066ca0f1869938f4860f41f56bc88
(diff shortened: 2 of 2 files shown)


scripts/deploy-local.sh
```diff
@@ -238,13 +238,12 @@ echo "========================================="
 
 # In 'none' mode, namespaces don't exist yet (operator isn't running),
 # so create them directly and skip waiting for pods that don't exist yet.
-DEPLOY_SECRETS_ENV=()
 if [ "${INSTALL_METHOD}" = "none" ]; then
```
gbenhaim, line 241: why this change is needed?

Omeramsc: When running the e2e I got the following error: `./scripts/deploy-local.sh: line 246: DEPLOY_SECRETS_ENV[@]: unbound variable`

it crash for every install method that is not none (release, build, local, ...).  making it currently broken for most users on macOS (like me).


we can use another approach, keeping the array, but checking its length before extending. WDYT?

yftacherzog: It was [added just a few days ago](https://github.com/konflux-ci/konflux-ci/pull/6031/changes#diff-ebb77f825fa5a1c2efa2e85e8ae2cbaf434bf67945d20cdf415fa7e9d49bbbb6R232) for specific reasons, so we should not just remove it.

yftacherzog: AI is saying:
```
# Option 1: Use parameter expansion with default
env "${DEPLOY_SECRETS_ENV[@]:-}" "${SCRIPT_DIR}/deploy-secrets.sh"

# Option 2: Only expand if array has elements
if [ ${#DEPLOY_SECRETS_ENV[@]} -gt 0 ]; then
    env "${DEPLOY_SECRETS_ENV[@]}" "${SCRIPT_DIR}/deploy-secrets.sh"
else
    env "${SCRIPT_DIR}/deploy-secrets.sh"
fi

# Option 3: Quote the expansion differently
env $(printf '%s\n' "${DEPLOY_SECRETS_ENV[@]}") "${SCRIPT_DIR}/deploy-secrets.sh"
```

@amisstea suppose that we fix this for mac, would triggering the e2e tests for openshift CI would be a good-enough verification for this?

amisstea: The Openshift-CI jobs don't use this script but rather `deploy-konflux-on-ocp.sh` so it won't matter if you adjust this.

yftacherzog: Do you recall what was the reason for adding that line?

amisstea: It was just a side effect of extracting the secret setup into `deploy-secrets.sh`. My AI agent probably didn't account for users on macOS. I see no harm in changing how these env variables are passed.

PR #6141, opened 2026-03-23
feat(KFLUXVNGD-845): add OpenShift CI trigger script for release gating

kelchen123:
Adds a script to trigger openshift ci e2e tests via gangway API, passing the operator image derived from the rc tag's git sha, to gate release promotions on passing tests.

Assisted By: Cursor

### DIFF
--- a/.github/scripts/trigger-openshift-ci-tests.sh
+++ b/.github/scripts/trigger-openshift-ci-tests.sh
@@ -5,14 +5,14 @@ set -euo pipefail
 # Triggers multiple Prow jobs via the Gangway REST API, polls for completion,
 # and returns success only if all jobs pass. Used to gate release promotions.
 #
-# The script derives the operator image tag from the git SHA of the RC tag.
-# Images are tagged with release-sha-<short-sha> by Konflux builds.
+# The script derives the operator image tag from the git SHA of the provided ref.
+# Images are tagged with their full commit SHA by Konflux builds.
 #
 # Usage:
-#   trigger-openshift-ci-tests.sh <rc_tag>
+#   trigger-openshift-ci-tests.sh <git_ref>
 #
 # Arguments:
-#   rc_tag - Release candidate tag (e.g., v0.1.5-rc.2)
+#   git_ref - Any git reference (tag, branch, SHA) e.g., v0.1.5-rc.2
 #
 # Environment:
 #   OPENSHIFT_CI_TOKEN - Gangway API token (required)
@@ -62,32 +62,33 @@ trap cleanup SIGINT SIGTERM SIGHUP
 
 # Validate arguments
 if [ $# -ne 1 ]; then
-  echo "Error: RC tag argument required"
-  echo "Usage: $0 <rc_tag>"
+  echo "Error: git ref argument required"
+  echo "Usage: $0 <git_ref>"
   echo "Example: $0 v0.1.5-rc.2"
   exit 1
 fi
 
-RC_TAG="$1"
+GIT_REF="$1"
 
-# Verify required environment variables
+# Verify required environment variables (disable tracing to avoid leaking token)
+{ set +x; } 2>/dev/null
 if [ -z "${OPENSHIFT_CI_TOKEN:-}" ]; then
   echo "Error: OPENSHIFT_CI_TOKEN environment variable is not set"
   exit 1
 fi
 
-# Derive operator image tag from the git SHA of the RC tag
-echo "Resolving git SHA for ${RC_TAG}..."
-COMMIT_SHA=$(git rev-parse "${RC_TAG}^{commit}" 2>/dev/null) || {
-  echo "Error: Failed to resolve git SHA for tag ${RC_TAG}"
-  echo "Ensure the tag exists and the repository has been fetched with tags"
+# Derive operator image tag from the git SHA of the provided ref
+echo "Resolving git SHA for ${GIT_REF}..."
+COMMIT_SHA=$(git rev-parse "${GIT_REF}^{commit}" 2>/dev/null) || {
+  echo "Error: Failed to resolve git SHA for ref ${GIT_REF}"
+  echo "Ensure the ref exists and the repository has been fetched"
   exit 1
 }
 
-SHORT_SHA="${COMMIT_SHA:0:7}"
-OPERATOR_IMAGE="${OPERATOR_REPO}:release-sha-${SHORT_SHA}"
+OPERATOR_IMAGE="${OPERATOR_REPO}:${COMMIT_SHA}"
 echo "Commit SHA: ${COMMIT_SHA}"
 echo "Operator image: ${OPERATOR_IMAGE}"
+echo "(KONFLUX_REF will be derived from image tag by step scripts)"
 echo ""
 
 # Trigger all jobs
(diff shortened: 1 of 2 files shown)


.github/scripts/trigger-openshift-ci-tests.sh
```diff
@@ -0,0 +1,183 @@
...
+# Exit codes:
+#   0 - Job completed successfully (SUCCESS)
+#   1 - Job failed, was aborted, or encountered an error
+#
+# Example:
+#   export OPENSHIFT_CI_TOKEN="your_gangway_token"
+#   trigger-openshift-ci-tests.sh v0.1.5-rc.2
+
+GANGWAY_URL="https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com/v1/executions"
+JOB_NAME="periodic-ci-konflux-ci-konflux-ci-main-ocp420-konflux-e2e-v420"
```
amisstea: We should also kick off the ARM job. Try to make it dynamic with a list of jobs so it's easy to add and remove them in the future.

.github/scripts/trigger-openshift-ci-tests.sh
```diff
@@ -0,0 +1,224 @@
...
+
+# Trigger OpenShift CI Tests Script
+# Triggers multiple Prow jobs via the Gangway REST API, polls for completion,
+# and returns success only if all jobs pass. Used to gate release promotions.
+#
+# The script derives the operator image tag from the git SHA of the RC tag.
+# Images are tagged with release-sha-<short-sha> by Konflux builds.
+#
+# Usage:
+#   trigger-openshift-ci-tests.sh <rc_tag>
```
yftacherzog: Can we just call it git_ref?
At first I thought, why mention rc, but now I think there's nothing specific here for tags at all.
It can be any ref and the code doesn't need to change

.github/scripts/trigger-openshift-ci-tests.sh
```diff
@@ -0,0 +1,224 @@
...
+  echo "Error: Failed to resolve git SHA for tag ${RC_TAG}"
+  echo "Ensure the tag exists and the repository has been fetched with tags"
+  exit 1
+}
+
+SHORT_SHA="${COMMIT_SHA:0:7}"
+OPERATOR_IMAGE="${OPERATOR_REPO}:release-sha-${SHORT_SHA}"
+echo "Commit SHA: ${COMMIT_SHA}"
+echo "Operator image: ${OPERATOR_IMAGE}"
+echo ""
```
yftacherzog: We can skip it and use the long sha directly.
We added the `release-sh-${SHORT_SHA}` tag in this project as I had some plans for it, but I think there's higher certainty that the long sha tag will stay there longer (it's used for more or less everything built with Konflux) and it also doesn't require additional logic.

.github/scripts/trigger-openshift-ci-tests.sh
```diff
@@ -0,0 +1,224 @@
...
+  echo "Error: RC tag argument required"
+  echo "Usage: $0 <rc_tag>"
+  echo "Example: $0 v0.1.5-rc.2"
+  exit 1
+fi
+
+RC_TAG="$1"
+
+# Verify required environment variables
+if [ -z "${OPENSHIFT_CI_TOKEN:-}" ]; then
```
gbenhaim, line 75: can you please explicitly disable tracing around this check (set +x)?

.github/scripts/trigger-openshift-ci-tests.sh
```diff
@@ -0,0 +1,224 @@
...
+echo ""
+
+# Trigger all jobs
+echo "Triggering ${#JOBS[@]} job(s)..."
+echo "========================================"
+
+for job_name in "${JOBS[@]}"; do
+  echo "Triggering: ${job_name}"
+  
+  REQUEST_PAYLOAD=$(jq -n \
```
gbenhaim, line 101: how does it know which revision of the konflux-ci repo it should check out?

yftacherzog: I think we're going to need to provide `KONFLUX_BRANCH` here as well. No?
Could the test on openshift CI handle tag/sha? Branch is not going to be relevant here.

kelchen123: The periodic jobs check out main by default. So the tests come from main while the image is from the RC tag; meaning there's potential for failure if tests changed significantly between when the RC was tagged and when the promotion script is called, but I don't know how likely this is. If it is not common for tests to be changed, and if test changes are usually compatible with recent release candidates, then I think it should be okay as we could just occasionally override those cases. But let me know if you think we need the exact ref checkout.

yftacherzog: We need the tests and the code to come from the same ref. 

Is it only the tests that are used by the ref? I suppose the ref is also needed for the crds and cr. No?

gbenhaim: test, manifest, and image should all use the same git commit.

kelchen123: implemented

.github/workflows/promote-release.yaml
```diff
@@ -34,8 +40,28 @@ jobs:
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
 
+      - name: Run OpenShift CI e2e tests
+        env:
+          OPENSHIFT_CI_TOKEN: ${{ secrets.OPENSHIFT_CI_TOKEN }}
+        run: |
+          .github/scripts/trigger-openshift-ci-tests.sh "${{ inputs.release_candidate_tag }}"
```
amisstea: To avoid the potential for arbitrary shell injection, this would be safer:
```suggestion
        env:
          OPENSHIFT_CI_TOKEN: ${{ secrets.OPENSHIFT_CI_TOKEN }}
          RC_TAG: ${{ inputs.release_candidate_tag }}
        run: |
          .github/scripts/trigger-openshift-ci-tests.sh "$RC_TAG"
```

.github/workflows/promote-release.yaml
```diff
@@ -34,8 +40,28 @@ jobs:
...
+
       - name: Promote release candidate to release tag
+        if: ${{ !inputs.dry_run }}
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           .github/scripts/promote-release.sh "${{ inputs.release_candidate_tag }}"
```
amisstea: Same shell injection concern here.

.github/workflows/promote-release.yaml
```diff
@@ -34,8 +40,28 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_RUN_ID: ${{ github.run_id }}
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
+          WORKFLOW_NAME: Promote Release
+        run: |
+          .github/scripts/create-or-update-issue.sh \
+            "Release Promotion Failure" \
+            "E2E tests or promotion failed for ${{ inputs.release_candidate_tag }}. See workflow run for details."
```
amisstea: Another one here.

gbenhaim (approved):

PR #6157, opened 2026-03-24
test(KFLUXVNGD-842): e2e on tekton

yftacherzog:
Adding Tekton tasks and pipeline for deploying the operator, optionally overriding upstream manifests, and running e2e tests.
    
This can later be used by Konflux teams to verify their changes within Konflux.

Assisted-by: Cursor

### DIFF
--- a/.tekton/konflux-operator-e2e-pull-request.yaml
+++ b/.tekton/konflux-operator-e2e-pull-request.yaml
@@ -1,0 +1,56 @@
+apiVersion: tekton.dev/v1
+kind: PipelineRun
+metadata:
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/konflux-ci/konflux-ci?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/cancel-in-progress: "true"
+    pipelinesascode.tekton.dev/max-keep-runs: "2"
+    pipelinesascode.tekton.dev/on-cel-expression: |
+      event == "pull_request" && target_branch == "main" ||
+      event == "push" && target_branch.startsWith("gh-readonly-queue/main/")
+  creationTimestamp: null
+  labels:
+    appstudio.openshift.io/application: konflux-operator
+    appstudio.openshift.io/component: konflux-operator
+    pipelines.appstudio.openshift.io/type: test
+  name: konflux-operator-e2e-on-pull-request
+  namespace: konflux-vanguard-tenant
+spec:
+  params:
+    - name: git-url
+      value: '{{source_url}}'
+    - name: revision
+      value: '{{revision}}'
+    - name: overrides-yaml
+      # Optional; See example in .tekton/pipelines/operator-e2e/pipeline.yaml.
+      value: ""
+    - name: konflux-ready-timeout
+      value: 15m
+    # OCI artifact prefix for kind-aws provision/deprovision (tag = PipelineRun name). Credentials secret must allow push/pull to this repo.
+    - name: oci-container-repo
+      value: quay.io/konflux-vanguard/konflux-operator-e2e-artifacts
+    # Secret in this namespace: registry creds for oci-container-repo above (not a pipeline default).
+    - name: oci-container-repo-credentials-secret
+      value: konflux-test-infra
+    - name: aws-credentials-secret
+      value: konflux-mapt-us-east-1
+    - name: deprovision-aws-credentials-secret
+      value: konflux-mapt-us-east-1
+    - name: release-ta-oci-storage
+      value: ""
+  pipelineRef:
+    resolver: git
+    params:
+      - name: url
+        value: '{{source_url}}'
+      - name: revision
+        value: '{{revision}}'
+      - name: pathInRepo
+        value: .tekton/pipelines/operator-e2e/pipeline.yaml
+  taskRunTemplate:
+    # kind-aws provision creates/patches Secret kfg-<PipelineRun name>; konflux-integration-runner has secrets create/patch (build-pipeline-* SAs do not).
+    serviceAccountName: konflux-integration-runner
+status: {}
--- a/.tekton/pipelines/operator-e2e/README.md
+++ b/.tekton/pipelines/operator-e2e/README.md
@@ -1,0 +1,113 @@
+# `operator-e2e-pipeline`
+
+Tekton pipeline for operator E2E that provisions Kind on AWS, deploys Konflux, runs E2E tests, and deprovisions.
+
+## Scope
+
+- Operator install uses Tekton `none` mode: the deploy task runs out-of-cluster `bin/manager` (not an operator image).
+- Deploy and tests are split across `deploy-konflux` and `konflux-e2e-tests`.
+- There are **no Pipeline workspaces**: each Task clones `konflux-ci` into an `emptyDir` (see the Task YAML). This is so the pipeline can be triggered both by Pipelines as Code and as IntegrationTestScenario.
+
+### Operator process vs. test phase
+
+In `none` mode, `bin/manager` is started as a **background process inside the `deploy-konflux` Task pod** (see `scripts/operator-e2e/tekton-deploy-operator-and-wait.sh`). When that Task finishes, the pod exits and **that operator process terminates** — it is not left running on the Kind cluster.
+
+The **`konflux-e2e-tests` Task** runs in a **separate pod** with kubeconfig only. There is **no `bin/manager` reconciliation loop** during integration or conformance tests. The cluster still runs the workloads the operator applied (build-service, integration-service, PaC, and so on); those controllers keep reconciling their own resources.
+
+The current conformance suite is written for that model: it exercises deployed services and GitHub flows, not “delete a Deployment and expect the Konflux operator to recreate it.” A test that assumed operator-level reconciliation during the test phase would not behave like a long-running operator install.
+
+## Inputs (params)
+
+- `git-url` (default: `https://github.com/konflux-ci/konflux-ci.git`): repository URL used for clone + git-resolved local tasks.
+- `revision` (required): git ref from `git-url` to test.
+- `overrides-yaml` (default: empty): optional inline overrides consumed by deploy task.
+- `konflux-ready-timeout` (default: `30m`): readiness timeout for Konflux CR.
+- `oci-container-repo` (required): OCI registry/repo prefix for kind-aws provision/deprovision artifacts (logs/state); no tag suffix—the pipeline appends `:$(context.pipelineRun.name)` for provision/deprovision. The deploy task also pushes **post-prep** `operator/pkg/manifests` to the **same repo** with tag `$(context.pipelineRun.name).pkg-manifests` so it does not replace the provision artifact.
+- `oci-container-repo-credentials-secret` (required): name of a Secret with registry credentials for `oci-container-repo` (kind-aws `oci-credentials`). This repo’s PAC PipelineRun uses `konflux-test-infra`.
+- `aws-credentials-secret` (default: `konflux-mapt-us-east-1`): AWS credentials for provision task.
+- `deprovision-aws-credentials-secret` (default: `konflux-mapt-us-east-1`): AWS credentials for deprovision task.
+- `release-ta-oci-storage` (default: empty): optional OCI ref for conformance trusted-artifacts flow.
+- `integration-go-test-extra-args` (default: empty): optional space-separated extra flags appended to integration `go test . ./pkg/...` (e.g. `-run=TestFoo -count=1`).
+- `conformance-go-test-extra-args` (default: empty): optional space-separated extra flags appended to conformance `go test` after the fixed Ginkgo options (e.g. `-ginkgo.focus=Subsuite`), same idea as `./test/e2e/run-e2e.sh` forwarding `"$@"`.
+- `catalog-url` (default: `https://github.com/konflux-ci/tekton-integration-catalog.git`): integration catalog repository.
+- `catalog-revision` (default: pinned commit SHA): `tekton-integration-catalog` ref for catalog tasks; override to move to a different commit.
+
+### Examples: `integration-go-test-extra-args` / `conformance-go-test-extra-args`
+
+Params are a **single string**; the Task passes them into the shell **without extra quoting**, so **spaces separate flags** (same as typing multiple words after `go test` locally). Do not wrap the whole value in inner quotes unless you intend one literal argument.
+
+**IntegrationTestScenario** — add under `spec.params` next to your other params:
+
+```yaml
+    - name: integration-go-test-extra-args
+      value: "-run=TestKonfluxIntegration -count=1"
+    - name: conformance-go-test-extra-args
+      value: "-ginkgo.skip=Flaky -ginkgo.v=false"
+```
+
+**Pipelines as Code** — in the PipelineRun template, set param values (adjust to your PAC variable syntax):
+
+```yaml
+    - name: integration-go-test-extra-args
+      value: ""
+    - name: conformance-go-test-extra-args
+      value: "-ginkgo.skip=Flaky"
+```
+
+**Direct `tkn` / YAML PipelineRun** — same shape: each param is one scalar string; use `=` style Ginkgo flags to avoid embedded spaces when possible.
+
+## Expected Secret shapes
+
+The pipeline passes **Secret names** as parameters. The **keys and formats** below match what **kind-aws-provision 0.2** and **kind-aws-deprovision 0.1** expect; open those task files at the same Git commit as `catalog-revision` in `pipeline.yaml` (for example [provision 0.2](https://github.com/konflux-ci/tekton-integration-catalog/blob/489cd0a413f52fd3fac90f38694f8fe51871be4a/tasks/mapt-oci/kind-aws-spot/provision/0.2/kind-aws-provision.yaml) and [deprovision 0.1](https://github.com/konflux-ci/tekton-integration-catalog/blob/489cd0a413f52fd3fac90f38694f8fe51871be4a/tasks/mapt-oci/kind-aws-spot/deprovision/0.1/kind-aws-deprovision.yaml) at the default pin). Re-verify whenever you bump `catalog-revision`.
+
+### `oci-container-repo-credentials-secret` (registry auth for `oci-container-repo`)
+
+- **Type:** `Opaque` is typical.
+- **Required data key:** `oci-storage-dockerconfigjson` — value must be the **contents of a `.dockerconfigjson`** file (same JSON `docker`/`podman` use). In manifests, use `stringData` for the raw JSON body, or put base64-encoded content under `data`.
+- The key name must be exactly `oci-storage-dockerconfigjson` (used by the catalog `secure-push-oci` step action).
+
+Example (replace placeholders; prefer creating via `kubectl create secret generic` or your GitOps tool rather than committing raw credentials):
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: my-oci-push-secret
+type: Opaque
+stringData:
+  # Body must be valid .dockerconfigjson; auth is base64("username:password").
+  oci-storage-dockerconfigjson: '{"auths":{"quay.io":{"auth":"<base64(username:password)>"}}}'
+```
+
+### `aws-credentials-secret` and `deprovision-aws-credentials-secret`
+
+Both reference the **same shape** of Secret unless your deprovision catalog task differs (this pipeline uses the same catalog family for both).
+
+- **Type:** `Opaque`
+- **Required data keys** (values are plain strings in `stringData`, or base64 in `data`):
+  - `access-key` — AWS access key ID
+  - `secret-key` — AWS secret access key
+  - `region` — AWS region name (e.g. `us-east-1`)
+  - `bucket` — S3 bucket name used by Mapt for this flow
+
+Example:
+
+```yaml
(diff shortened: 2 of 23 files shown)


.tekton/konflux-operator-e2e-pull-request.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - name: url
+        value: '{{source_url}}'
+      - name: revision
+        value: '{{revision}}'
+      - name: pathInRepo
+        value: .tekton/pipelines/operator-e2e/pipeline.yaml
+  taskRunTemplate:
+    # kind-aws provision creates/patches Secret kfg-<PipelineRun name>; konflux-integration-runner has secrets create/patch (build-pipeline-* SAs do not).
+    serviceAccountName: konflux-integration-runner
+  workspaces:
```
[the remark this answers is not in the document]
yftacherzog: PRs opened from forks need Tekton pipelines to be triggered with comment `/retest` to trigger this pipeline. As far as I understand it, only authorized users will be able to use this comment to trigger CI.

I think this issue extends to repos that will use the pipelines/tasks. Might be that ultimately, we need to release the pipeline/tasks as OCI artifacts.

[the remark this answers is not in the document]
yftacherzog: BTW, I'm sure if this is very different than the issue we had with pull_request_target workflows. I.e. we accept the fact that PRs code can expose secrets, so we only allow authorized users to trigger CI.

.tekton/pipelines/operator-e2e/README.md
```diff
@@ -0,0 +1,34 @@
+# `operator-e2e-pipeline`
```
gbenhaim, line 1: I wonder if we should include it in our docs website under the installation methods. @yftacherzog WDYT?

.tekton/pipelines/operator-e2e/README.md
```diff
@@ -0,0 +1,34 @@
...
+- Tekton flow is `none`-mode oriented for operator install (out-of-cluster `bin/manager` in deploy task).
+- Deploy and tests are split across `deploy-konflux` and `konflux-e2e-tests`.
+
+## Inputs (params)
+
+- `git-url` (default: `https://github.com/konflux-ci/konflux-ci.git`): repository URL used for clone + git-resolved local tasks.
+- `revision` (required): git ref from `git-url` to test.
+- `overrides-yaml` (default: empty): optional inline overrides consumed by deploy task.
+- `konflux-ready-timeout` (default: `30m`): readiness timeout for Konflux CR.
+- `oci-container-repo` (default: `quay.io/example/konflux-operator-e2e-artifacts`): OCI prefix used by provision/deprovision tasks.
```
gbenhaim: not sure I understand the purpose of this repo

.tekton/pipelines/operator-e2e/README.md
```diff
@@ -0,0 +1,34 @@
...
+- `konflux-test-infra-secret` (default: `konflux-test-infra`): secret for OCI credentials.
```
gbenhaim: not sure I understand the purpose of this secret

scripts/operator-e2e/apply-overrides-from-yaml.sh
```diff
@@ -0,0 +1,76 @@
...
+mkdir -p "${REPO_ROOT}/.tmp"
+if [[ -n "${OVERRIDES_YAML_PATH:-}" ]]; then
+  if [[ ! -f "${OVERRIDES_YAML_PATH}" ]]; then
+    echo "error: OVERRIDES_YAML_PATH is not a readable file: ${OVERRIDES_YAML_PATH}" >&2
+    exit 1
+  fi
+  cp "${OVERRIDES_YAML_PATH}" "${REPO_ROOT}/.tmp/overrides.yaml"
+else
+  : "${OVERRIDES_YAML:?set OVERRIDES_YAML or OVERRIDES_YAML_PATH}"
+  echo "${OVERRIDES_YAML}" > "${REPO_ROOT}/.tmp/overrides.yaml"
```
gbenhaim: 1. echo can misinterpret YAML content — apply-overrides-from-yaml.sh line 33
Severity: Concern

echo "${OVERRIDES_YAML}" will interpret leading -n, -e, or -E as options on some implementations (e.g. dash, some busybox). Since OVERRIDES_YAML is user-supplied YAML that starts with - name:, the leading - could trigger this.


apply-overrides-from-yaml.sh
Lines 33-33
  echo "${OVERRIDES_YAML}" > "${REPO_ROOT}/.tmp/overrides.yaml"
Suggested fix:


printf '%s\n' "${OVERRIDES_YAML}" > "${REPO_ROOT}/.tmp/overrides.yaml"

.tekton/pipelines/operator-e2e/pipeline.yaml
```diff
@@ -0,0 +1,208 @@
...
+            '[{\"containerPort\":30012,\"hostPort\":8180,\"protocol\":\"TCP\"}]'
+        - name: cpus
+          value: "16"
+        - name: memory
+          value: "32"
+        - name: spot
+          value: "false"
+        - name: harden-control-plane
+          value: "true"
+    - name: clone-repository
```
gbenhaim: any reason not to use the git-clone task konflux provides?

.tekton/konflux-operator-e2e-pull-request.yaml
```diff
@@ -0,0 +1,78 @@
...
+metadata:
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/konflux-ci/konflux-ci?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/cancel-in-progress: "true"
+    pipelinesascode.tekton.dev/max-keep-runs: "2"
+    pipelinesascode.tekton.dev/on-cel-expression: |
+      event == "pull_request" && target_branch == "main" ||
```
avi-biton, line 12: Do we want to run it for each PR in main?
We can reduce it only to operator related PRs and also  for example filter out documentation PRs

.tekton/konflux-operator-e2e-pull-request.yaml
```diff
@@ -0,0 +1,78 @@
...
+      # value: |
+      #   - name: segment-bridge
+      #     git:
+      #       - sourceRepo: konflux-ci/segment-bridge
+      #         remote:
+      #           repo: https://github.com/konflux-ci/segment-bridge
+      #           ref: {{revision}}
+      #     images:
+      #       - orig: quay.io/konflux-ci/segment-bridge
+      #         replacement: quay.io/redhat-user-workloads/konflux-vanguard-tenant/segment-bridge/segment-bridge:on-pr-{{revision}}
```
avi-biton: The `revision` parameter is taken from `konflux-ci` sha.
It does not necessary means that we have the same sha in other repositories (like `segment-bridge` in this example)

yftacherzog: It's coming from PAC (https://pipelinesascode.com/docs/concepts/#8-variable-substitution)

avi-biton: So IIUC, this example is valid for use in components repository where the revision will be taken from their sha and will be replaced correctly, however, if we will try to use overrides in this repository it will fail

yftacherzog: We can use it everywhere, but the content of the overrides will be different. When running in our repo, I don't think there's a practical use case for the `revision` argument, but for our upstream services, I think it will be the main instrument for testing PR code together with latest from konflux-ci/konflux-ci.

.tekton/konflux-operator-e2e-pull-request.yaml
gbenhaim: I thinl that it can be beneficial to start with running it for each pr so we can understand how stable it's.

.tekton/pipelines/operator-e2e/pipeline.yaml
yftacherzog: I'll look into that.

scripts/operator-e2e/apply-overrides-from-yaml.sh
yftacherzog: I'm testing a go-based approach. I'll close this one for now.

.tekton/pipelines/operator-e2e/README.md
yftacherzog: Might be

.tekton/pipelines/operator-e2e/pipeline.yaml
yftacherzog: Changed

.tekton/pipelines/operator-e2e/README.md
yftacherzog: It's used by the VM provisioning process to store logs. I'm removing the default value as it makes no sense to have it.

.tekton/pipelines/operator-e2e/README.md
yftacherzog: I renamed some of the things, changed the default values and enhanced the pipeline readme to explain how the secret should look like.

yftacherzog:
@gbenhaim @avi-biton @flacatus 
The pipeline is now pushing the manifests that are used for deploying Konflux to Quay, so we have more confidence in what is being deployed when manifests are overridden by the pipeline. For example, below, is the result of a run in which I overridden segment and ui, and also included a broken ref to an image to see that Konflux is failing deployment.

This is the artifact: quay.io/konflux-vanguard/konflux-operator-e2e-artifacts:konflux-operator-e2e-on-pull-request-62qnh.pkg-manifests

This is the overrides inputs:

```yaml
    - name: overrides-yaml
      value: |
        - name: segment-bridge
          git:
            - sourceRepo: konflux-ci/segment-bridge
              remote:
                repo: https://github.com/konflux-ci/segment-bridge
                ref: f9b0012a9b90842c502dc88419000a6eabb9b276
          images:
            - orig: quay.io/konflux-ci/segment-bridge
              replacement: quay.io/redhat-user-workloads/konflux-vanguard-tenant/segment-bridge/segment-bridge:on-pr-f9b0012a9b90842c502dc88419000a6eabb9b276
        - name: ui
          images:
            - orig: quay.io/konflux-ci/konflux-ui
              replacement: quay.io/redhat-user-workloads/konflux-ui-tenant/konflux-ui/konflux-ui:on-pr-e8b4f4148e19ba5a456bc8786d9b7d81a48fd1ccZZZ
```

operator/pkg/overrides/overrides.go
```diff
@@ -0,0 +1,660 @@
...
+				if !ok {
+					continue
+				}
+				if imageNameFromMap(im) != ov.Orig {
+					continue
+				}
+				im["newName"] = newName
+				im["newTag"] = newTag
+				delete(im, "digest")
+				changed = true
```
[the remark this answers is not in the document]
yftacherzog: This should now be fixed. can you reevaluate?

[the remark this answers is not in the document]
yftacherzog: /ask "This should now be fixed. can you reevaluate?"

operator/cmd/overrides/main.go
```diff
@@ -0,0 +1,79 @@
+package main
```
gbenhaim, line 11: is it possible to add module docs and describe the purpose of this command? it would be useful for AI as well.

gbenhaim:
_Comment from Gal's Cursor_

It would be useful to document somewhere (pipeline README or `scripts/operator-e2e/README.md`) that the operator process (`bin/manager`) **terminates when the deploy-konflux Task completes** — it runs as a background process inside that Task's pod, so when the pod exits, the operator is gone.

The `konflux-e2e-tests` Task then runs in a separate pod **without the operator running**. The tests rely on the services the operator deployed (build-service, integration-service, PaC, etc.) which continue running independently in the Kind cluster, but there is no operator reconciliation loop active during the test phase.

This is fine for the current conformance suite since it exercises the deployed services rather than operator reconciliation. But it's a non-obvious aspect of the architecture that future contributors should be aware of — e.g., if a test were to delete a Deployment expecting the operator to recreate it, that wouldn't work.

A short note in the pipeline README's "Scope" section (which already mentions `none` mode) would be a natural place for this.

gbenhaim:
_Comment from Gal's Cursor_

### Suggestion: additional test coverage for `operator/pkg/overrides`

The existing 14 tests cover the happy paths well, but a few areas could use more coverage:

**1. Validation edge cases (`validateOverrides`)**
Currently only one rejection case is tested. Worth adding negative tests for:
- Both `Remote` and `LocalPath` set on the same `GitRule`
- Missing `SourceRepo`
- Invalid `SourceRepo` format (not `org/repo`)
- Invalid image references
- Duplicate component names

**2. Input parsing helpers — no direct tests today**
- `parseImageReference` — handles `name:tag@digest` parsing; only `splitImageReference` is tested
- `normalizeOrgRepo` / `githubURLParts` — parse GitHub URLs and org/repo strings, edge cases include trailing slashes, `.git` suffixes, HTTPS vs SSH
- `firstMatchingRule` — matches git rules to kustomization resource URLs

These are all parsing external/user input, so they're high-value targets for table-driven tests.

**3. `Runner.Apply()` with git rules**
`TestRunnerApplyWithImageOverrideOnly` only exercises the image override path. There's no integration-level test that goes through git rule application → `kustomize build` → manifest rebuild. (Understandable if `kustomize` availability is a concern — could be gated with a build tag or test skip.)

Not a blocker, but would increase confidence in the override engine — especially since the `overrides-yaml` parameter is user-facing input from PipelineRun params.

scripts/operator-e2e/prepare-conformance-env.sh
```diff
@@ -0,0 +1,22 @@
...
+append_env() {
+  local line="$1"
+  if [[ -n "${GITHUB_ENV:-}" ]]; then
+    echo "$line" >> "$GITHUB_ENV"
+  else
+    echo "export $line"
+  fi
+}
+
+line="$(grep 'RELEASE_SERVICE_CATALOG_REVISION=' test/e2e/release-service-catalog-revision | sed 's/^export //')"
```
gbenhaim: we don't have this file anymore. the revision is baked into the setup-release.sh script.

.tekton/tasks/konflux-e2e-tests/task.yaml
```diff
@@ -0,0 +1,135 @@
...
+          valueFrom:
+            secretKeyRef:
+              name: konflux-operator-e2e-credentials
+              key: GH_ORG
+        - name: GH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: konflux-operator-e2e-credentials
+              key: GH_TOKEN
+        - name: QUAY_DOCKERCONFIGJSON
```
gbenhaim: QUAY_DOCKERCONFIGJSON and RELEASE_CATALOG_TA_QUAY_TOKEN are not needed anymore, since we use image-controller for creating the repositories required for the release process. it's done as part of setup-release.sh.

gbenhaim:
## Conformance test setup is out of sync with `main`

The PR's conformance test code diverges from the current `main` branch in several significant ways. It appears to be based on an older version of the tests that predates the `setup-release.sh` consolidation.

### 1. Release setup mechanism: `setup-release.sh` ConfigMap (main) vs. inline Go code (PR)

**On `main`**, the conformance test calls `runSetupRelease()` which downloads `setup-release.sh` from the `konflux-cli/setup-release` ConfigMap (deployed by the operator) and executes it. This single script handles creating the managed namespace, ImageRepositories, EnterpriseContractPolicy, ReleasePlanAdmission, etc.

**In this PR**, the conformance test calls `createReleaseConfig()` — a completely different Go function that manually creates all release resources via Kubernetes client calls. This bypasses the operator-shipped script and re-implements its logic in Go.

### 2. Catalog revision source: deleted file (PR) vs. `setup-release.sh` default (main)

**On `main`**: The catalog revision is baked into `setup-release.sh` (line 87: `CATALOG_REVISION="48a31f6..."`) and tracked by Renovate there. No external file is needed.

**In this PR**: `prepare-conformance-env.sh` (line 18) reads `RELEASE_SERVICE_CATALOG_REVISION` from `test/e2e/release-service-catalog-revision`:
```bash
line="$(grep 'RELEASE_SERVICE_CATALOG_REVISION=' test/e2e/release-service-catalog-revision | sed 's/^export //')"
```
And `GetReleaseServiceCatalogRevision()` in Go also falls back to that same file.

**That file no longer exists on `main`.** Both paths will fail at runtime.

### 3. Release pipeline path: different pipelines

- **`main`**: `setup-release.sh` uses `pipelines/managed/push-to-external-registry/push-to-external-registry.yaml`
- **PR**: `createReleaseConfig()` uses `pipelines/managed/e2e/e2e.yaml`

### Suggested fix

The PR should rebase on current `main` and use the existing `runSetupRelease()` + `patchECPForE2E()` pattern from `main`'s `setup.go`, which delegates to the operator-shipped `setup-release.sh` ConfigMap. This would:
- Eliminate the need for `prepare-conformance-env.sh` to resolve the catalog revision (it's already in the script)
- Use the same release pipeline as the rest of CI
- Avoid re-implementing release resource creation in Go

---
*From Gal's Cursor*

gbenhaim:
## Question: custom test suite support

The existing `test/e2e/run-e2e.sh` forwards extra arguments (`"$@"`) to `go test`, allowing callers to pass `-ginkgo.focus`, `-ginkgo.skip`, etc. The new Tekton pipeline hardcodes the test invocation without exposing any such parameter.

Could you double-check whether the legacy upstream task being replaced here allowed users to provide their own test suite or test filters? If so, it would be good to preserve that capability (e.g., via an optional pipeline param).

---
*From Gal's Cursor*

gbenhaim:
## Suggestion: make the release service catalog revision customizable

Currently the pipeline hardcodes / auto-extracts the `RELEASE_SERVICE_CATALOG_REVISION` from the deployed manifests. It would be valuable to expose this as an optional pipeline parameter (with the current behavior as the default).

This would allow the `konflux-ci/release-service-catalog` repo to reuse this same Tekton pipeline in its own CI — pointing the revision at the PR branch under test instead of the pinned production SHA. Without this, that repo would need a separate test harness or fork the pipeline definition.

---
*From Gal's Cursor*

yftacherzog:
> Could you double-check whether the legacy upstream task being replaced here allowed users to provide their own test suite or test filters? If so, it would be good to preserve that capability (e.g., via an optional pipeline param).
> 
> _From Gal's Cursor_

The legacy task only deployed Konflux, so for the task running the e2e tests, we can decide what to include.

operator/cmd/overrides/main.go
yftacherzog: Done

scripts/operator-e2e/prepare-conformance-env.sh
yftacherzog: I rebased on top of the latest changes and aligned with the current approach.

.tekton/tasks/konflux-e2e-tests/task.yaml
yftacherzog: done

yftacherzog:
> ## Suggestion: make the release service catalog revision customizable
> Currently the pipeline hardcodes / auto-extracts the `RELEASE_SERVICE_CATALOG_REVISION` from the deployed manifests. It would be valuable to expose this as an optional pipeline parameter (with the current behavior as the default).
> 
> This would allow the `konflux-ci/release-service-catalog` repo to reuse this same Tekton pipeline in its own CI — pointing the revision at the PR branch under test instead of the pinned production SHA. Without this, that repo would need a separate test harness or fork the pipeline definition.
> 
> _From Gal's Cursor_

I'm tempted to leave this for later.

yftacherzog:
> #### High
> * **[Correctness]** `.tekton/pipelines/operator-e2e/pipeline.yaml:416,480` — ITS (`deploy-konflux-its` and `konflux-e2e-tests-its`) task resolver `revision` is hardcoded to `upstream-ci-action` (the PR branch) instead of `main`. Once the branch is deleted after merge, the ITS flow will fail to resolve task YAML.
>   Remediation: Change both `value: upstream-ci-action` lines to `value: main`.

this is temporary so we can test the PR changes.

yftacherzog:
>[Possible Issue](https://github.com/konflux-ci/konflux-ci/pull/6157/files#diff-6d96153337934071a24af2fda56a1265eeb8579270f97bec1c23502375336c3aR257-R337)
applyGitRulesToKustomization updates images[].newTag for all images whenever exactly one remote ref is used in rewritten resources. This can unintentionally retag unrelated images in the same kustomization file. Consider scoping the newTag update to images that correspond to the rewritten git resource/repo, or avoid mutating images automatically unless an explicit image override exists.

Added a test case and applied a fix.

yftacherzog:
@gbenhaim I addressed your comments except for one which I prefer we keep for later. I also went over the bots reviews and addressed the points which I think are important at the moment.

The pipeline is referencing the PR branch as the source of the tasks to allow it to be tested. This will need to change to point to `main` before we merge this PR. Users could pin the task reference by digest.

gbenhaim:
Lgtm

gbenhaim:
What is this konfku-ci-review bot? Can I configured to reuse the same comment when updating the pr?

yftacherzog:
> What is this konfku-ci-review bot? Can I configured to reuse the same comment when updating the pr?

That's the review from the Fullsend process. We have it more complete on the caching repo.
@ralphbean @kelchen123 @amisstea can comment on what's possible here in terms of configs.

yftacherzog:
> Lgtm

@gbenhaim I changed the pipeline to point to main now. It's expected to fail now until we have the tekton resources on that branch.

gbenhaim (approved):

flacatus (approved):

PR #6158, opened 2026-03-24
feat(KFLUXVNGD-852): add clusterId to info.json

Omeramsc:
Populate the clusterId field in the konflux-public-info ConfigMap so both UI and backend share a stable cluster identifier for Segment event hashing. On OpenShift the value comes from ClusterVersion.spec.clusterID; on vanilla K8s it falls back to the kube-system namespace UID.

Made-with: Cursor

### DIFF
--- a/operator/internal/controller/info/konfluxinfo_controller.go
+++ b/operator/internal/controller/info/konfluxinfo_controller.go
@@ -322,16 +322,18 @@ func (r *KonfluxInfoReconciler) ensureNamespaceExists(ctx context.Context, tc *t
 
 // generateInfoJSON generates info.json content from PublicInfo.
 // Provides defaults if fields are missing. k8sVersion is the current cluster Kubernetes version (non-cached).
-func (r *KonfluxInfoReconciler) generateInfoJSON(config *konfluxv1alpha1.PublicInfo, k8sVersion, openShiftVersion string) ([]byte, error) {
-	info := r.applyInfoDefaults(config, k8sVersion, openShiftVersion)
+// clusterId is the stable cluster identifier (OpenShift ClusterVersion UID or kube-system namespace UID).
+func (r *KonfluxInfoReconciler) generateInfoJSON(config *konfluxv1alpha1.PublicInfo, k8sVersion, openShiftVersion, clusterId string) ([]byte, error) {
+	info := r.applyInfoDefaults(config, k8sVersion, openShiftVersion, clusterId)
 	return json.MarshalIndent(info, "", "    ")
 }
 
 // applyInfoDefaults applies default values to PublicInfo if not specified.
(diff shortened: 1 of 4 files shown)


operator/internal/controller/info/konfluxinfo_controller.go
```diff
@@ -391,19 +394,29 @@ func (r *KonfluxInfoReconciler) reconcileInfoConfigMap(ctx context.Context, tc *
 			if err != nil {
 				return fmt.Errorf("failed to get OpenShift version: %w", err)
 			}
+			clusterId, err = clusterinfo.GetOpenShiftClusterID(ctx, r.Client)
+			if err != nil {
+				log.V(1).Info("Could not get OpenShift cluster ID, falling back to kube-system UID", "error", err)
```
gbenhaim: I don't think this fallback is valid. Imagine that the api call for getting the cluster version fails because some temporary network issue, it will cause the cluster ID to change. I think that in this case we should just return an errors and not make any update to the configmap.

operator/internal/controller/info/konfluxinfo_controller.go
```diff
@@ -391,19 +394,29 @@ func (r *KonfluxInfoReconciler) reconcileInfoConfigMap(ctx context.Context, tc *
...
+			}
+		}
+		if clusterId == "" {
+			var err error
+			clusterId, err = clusterinfo.GetKubeSystemUID(ctx, r.Client)
+			if err != nil {
+				log.V(1).Info("Could not get kube-system UID for cluster ID", "error", err)
```
gbenhaim: we shouldn't ignore this error IMO. Is there a case where we intentionally can't get this namespace?

operator/internal/controller/info/konfluxinfo_controller.go
gbenhaim, line 147: the controller now need permission to get the `kube-system` namespace.

Omeramsc: The operator's role.yaml already grants unrestricted get on namespaces

operator/internal/controller/info/konfluxinfo_controller.go
Omeramsc: good catch, fixed

operator/internal/controller/info/konfluxinfo_controller.go
Omeramsc: Right, if this namespace does not exist, something has probably gone wrong... will return an error

operator/internal/controller/info/konfluxinfo_controller.go
gbenhaim: each reconciler should declare the permissions it needs. This is useful for documentation and also for not removing permissions by mistake.

PR #6193, opened 2026-03-26
chore: add agent skill for debugging e2e test failures

gbenhaim:
Add an Agent Skill that guides AI assistants through debugging failed Konflux e2e test runs from GitHub Actions. The skill covers downloading logs/artifacts, analyzing JUnit reports and Ginkgo output, inspecting cluster state, tracing PaC webhook delivery chains, and classifying root causes.

Agent Skills (https://claude.com/blog/skills) are a portable format for packaging domain expertise that AI coding assistants can load on demand. They are supported across different tools including Cursor and Claude Code. Skills are loaded automatically when the AI detects a relevant task, or can be invoked explicitly using a slash command.

Files added:
- SKILL.md: main skill with step-by-step debugging workflow
- reference.md: log structure, upstream repo mapping, failure patterns
- pac-webhook-debugging.md: PaC webhook chain tracing guide
- scripts/download-logs.sh: helper to fetch artifacts from GH Actions

Assisted-By: Cursor

### DIFF
--- a/.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
+++ b/.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+#
+# Download and extract e2e test artifacts from a GitHub Actions run.
+#
+# Usage: download-logs.sh <run-id> [arch]
+#   run-id  — GitHub Actions run ID
+#   arch    — amd64 (default) or arm64
+#
+# Output directory: /tmp/e2e-debug-<run-id>/
+
+set -euo pipefail
+
+REPO="konflux-ci/konflux-ci"
+
+run_id="${1:?Usage: download-logs.sh <run-id> [arch]}"
+arch="${2:-amd64}"
+workdir="/tmp/e2e-debug-${run_id}"
+
+mkdir -p "$workdir"
+echo "Downloading artifacts for run ${run_id} (${arch}) into ${workdir}..." >&2
+
+gh run download "$run_id" \
+  --repo "$REPO" \
+  --name "logs-${arch}" \
+  --dir "${workdir}/logs" 2>/dev/null || {
+    echo "Warning: could not download logs-${arch} artifact. It may have expired (90-day retention)." >&2
+  }
+
+echo "Downloading raw job log for the failing job..." >&2
+failing_job_id=$(
+  gh api "repos/${REPO}/actions/runs/${run_id}/jobs" \
+    --jq ".jobs[] | select(.conclusion==\"failure\" and (.name | test(\"${arch}\"; \"i\"))) | .id" \
+  | head -1
+) || {
+  echo "Warning: could not query jobs for run ${run_id}." >&2
+  failing_job_id=""
+}
+
+if [[ -n "$failing_job_id" ]]; then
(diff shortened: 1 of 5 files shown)


.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
```diff
@@ -0,0 +1,72 @@
...
+
+mkdir -p "$workdir"
+echo "Downloading artifacts for run ${run_id} (${arch}) into ${workdir}..." >&2
+
+gh run download "$run_id" \
+  --repo "$REPO" \
+  --name "logs-${arch}" \
+  --dir "${workdir}/logs" 2>/dev/null || {
+    echo "Warning: could not download logs-${arch} artifact. It may have expired (90-day retention)." >&2
+  }
```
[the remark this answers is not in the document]
avi-biton: I think this one is actually a bug

.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo "Downloading raw job log for the failing job..." >&2
+failing_job_id=$(
+  gh api "repos/${REPO}/actions/runs/${run_id}/jobs" \
```
avi-biton, line 31: We might want to add a failure handling like in the 
```
gh run download ...
... || {}
```

gbenhaim: I think this this something we can let the agent tweak during runtime.

.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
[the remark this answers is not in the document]
gbenhaim: Gal's Cursor here. I downloaded an actual artifact from a recent run (run `24230837163`) and verified the structure. The artifact does **not** contain a `logs/` prefix — files like `operator-logs.log`, `artifacts/`, etc. are at the root of the zip. So `--dir "${workdir}/logs"` correctly places them at `${workdir}/logs/operator-logs.log`, `${workdir}/logs/artifacts/...`, which matches the paths in SKILL.md.

`upload-artifact@v7` with `path: logs` strips the specified directory and stores contents with paths relative to it. And `gh run download --name logs-amd64` extracts directly into the `--dir` target without creating a subdirectory. No nesting bug here.

.cursor/skills/debug-e2e-tests/scripts/download-logs.sh
gbenhaim: Gal's Cursor here. Added error handling for both `gh api` calls that fetch the `failing_job_id` — they now use the same `|| { echo "Warning: ..." ; failing_job_id="" }` pattern as the artifact download. This way the script always completes (prints the directory listing and outputs the workdir) even when the API call fails, instead of crashing due to `set -euo pipefail`.

gbenhaim:
Gal's Cursor here. Good catch — added `.cursor/` to the exclusion list in the TOC check workflow (`check-toc.yaml`). The find command now skips both `./operator/docs/*` and `./.cursor/*`.

gbenhaim:
@avi-biton @yftacherzog this one is ready.

yftacherzog (approved):

avi-biton (approved):

PR #6299, opened 2026-04-07
integrationservice: handle KonfluxUI read errors without clearing CONSOLE_URL

harshakumar25:
## Summary
- treat only NotFound when reading the KonfluxUI singleton as optional
- return reconcile error for non-NotFound KonfluxUI read failures
- keep CONSOLE_URL stable by avoiding the degraded apply path on transient read failures
- add focused unit tests for URL present, NotFound, and non-NotFound error cases

## Why
When KonfluxUI read errors were all handled like NotFound, transient API/RBAC/transport failures could lead to an empty console URL being written into CONSOLE_URL during reconcile.

## Testing
- go test ./internal/controller/integrationservice -run TestGetConsoleURLFromKonfluxUI -count=1
- go test ./internal/controller/integrationservice -run TestBuildControllerManagerOverlay_ConsoleURL -count=1

Fixes #6295

### DIFF
--- a/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
+++ b/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


harshakumar25:
@dirgim 
returning an error here causes backoff requeues that hit the same stressed API, which is counter-productive. Since the controller already watches Konflux-ui events via SetupWithManager, recovery happens naturally when the API stabilises, so there's no need for an explicit requeue. Updated to log.Error + return nil which skips the manifest apply (protecting CONSOLE_URL) without adding any load
Thanks !

harshakumar25:
hi @yftacherzog ,
everything is perfect can you pleasse approve/merge this 
thanks !

yftacherzog:
@harshakumar25 there's no need to merge main into this branch. It also breaks the reporting of any running tests.

yftacherzog:
> * **[style]** `operator/internal/controller/integrationservice/konfluxintegrationservice_controller_consoleurl_test.go` — The new test file uses `testing.T` + gomega `NewWithT` subtests, while the existing `_controller_test.go` uses Ginkgo/Gomega suite style. This is reasonable for focused unit tests vs. integration suite tests, but worth noting as a minor style difference.

@harshakumar25 Can you look into that? Also, what's the reason for introducing a new test file rather than adding to existing one that handles the modified file?

For the issue being fixed, I can't say I understand the context completely. I'll try reading into it more when I have the chance. I also requested a review from @gbenhaim to get his perspective.

PR #6324, opened 2026-04-10
feat: add webhookURLs field and fix PAC_WEBHOOK_URL for OpenShift

gbenhaim:
Add a `webhookURLs` map to KonfluxBuildServiceSpec that lets users map repository URL prefixes to external webhook endpoints. The operator creates a hashed ConfigMap from this mapping and mounts it into the build-service-controller-manager pod via a volume already declared in the manifest (with optional: true).

On non-OpenShift clusters (e.g. Kind), PAC_WEBHOOK_URL is set to the in-cluster PaC service only when no webhookURLs are configured, so it does not override the webhook config JSON. On OpenShift, the build-service auto-discovers the PaC Route, so PAC_WEBHOOK_URL is never injected. User-provided env overrides from the CR spec always take precedence.

Set PAC_WEBHOOK_INSECURE_SSL=1 in the e2e sample CR so the build-service
skips TLS verification for PaC webhook URLs in OpenShift test environments
where the PaC route uses a self-signed certificate.

Includes documentation for the new feature and unit/integration tests.

Ref: RHTAP-6484

Assisted-By: Cursor


operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -200,51 +227,121 @@ func (r *KonfluxBuildServiceReconciler) applyManifests(ctx context.Context, tc *
...
+		webhookConfigNamespace,
+		webhookConfigDataKey,
+		webhookConfigLabel,
+		FieldManager,
 	)
+
+	result, err := hcm.Apply(ctx, string(jsonData), owner)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
gbenhaim: Thanks for flagging this — Gal's Cursor here 👋

This concern is theoretically valid but practically a non-issue for the following reasons:

1. **This is not new behavior.** The `hashedconfigmap` package is already used by other controllers in this repo (e.g., the UI controller for the Dex ConfigMap) with the same cleanup semantics. If this were a real operational problem, it would have surfaced there already.

2. **Kubernetes volume mounts use a local copy.** Once a pod starts and the kubelet mounts the ConfigMap volume, the data is cached locally on the node (via tmpfs/projected volumes). Deleting the ConfigMap object from the API server does **not** immediately remove the mounted file from a running pod — the old pod keeps running with the old file until it is terminated.

3. **The rolling update window is short.** The build-service deployment typically runs 1 replica. The old pod only stays alive for the brief termination grace period after the new pod becomes ready, making the risk window negligible.

No action needed here. If we ever wanted to keep N≥2 most recent hashed ConfigMaps as an extra safety margin, that would be a change to the shared `hashedconfigmap` package and outside the scope of this PR.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -200,51 +227,121 @@ func (r *KonfluxBuildServiceReconciler) applyManifests(ctx context.Context, tc *
...
 }
 
 // buildBuildControllerManagerOverlay builds the pod overlay for the controller-manager deployment.
-// On OpenShift, the PaC webhook URL is automatically set to the OpenShift Pipelines service
-// endpoint. User-provided env vars from the CR spec take precedence via strategic merge.
-func buildBuildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec, clusterInfo *clusterinfo.Info) *customization.PodOverlay {
+// On non-OpenShift without webhookURLs, PAC_WEBHOOK_URL is set to the default PaC service
+// endpoint so that Kind works out of the box. When webhookURLs are configured, PAC_WEBHOOK_URL
+// is not set because it would override the webhook config JSON.
+// On OpenShift, PAC_WEBHOOK_URL is never set — the build-service auto-discovers the PaC Route.
```
yftacherzog, line 247: Was there anything new that allowed this auto-discovery?

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -200,51 +227,121 @@ func (r *KonfluxBuildServiceReconciler) applyManifests(ctx context.Context, tc *
...
-		),
+	podOpts = append(podOpts,
+		customization.WithContainerBuilder(buildManagerContainerName, containerOpts...)(customization.DeploymentContext{Replicas: spec.BuildControllerManager.Replicas}),
+	)
+	return customization.NewPodOverlay(podOpts...)
+}
+
+// ensureNamespaceExists applies the build-service Namespace from the embedded manifests.
+// This must run before any resources are created in the namespace (e.g., ConfigMaps).
+func (r *KonfluxBuildServiceReconciler) ensureNamespaceExists(ctx context.Context, tc *tracking.Client) error {
```
yftacherzog, line 291: Maybe we should create an issue to unify the implementation of this function. We have something very similar in 2 other reconcilers.

yftacherzog (approved):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
gbenhaim: Gal's Cursor here 👋

The auto-discovery is not new — it's existing behavior in the build-service controller. When `PAC_WEBHOOK_URL` is not set, the build-service already discovers the PaC Route on OpenShift automatically.

Previously, the operator always set `PAC_WEBHOOK_URL` (to the OpenShift Pipelines service endpoint), which actually overrode this built-in auto-discovery. This PR stops setting it on OpenShift, letting the existing discovery mechanism work as intended.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
gbenhaim: Gal's Cursor here 👋

Agreed — created #6339 to track extracting `ensureNamespaceExists` into `internal/common/`.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
yftacherzog: Do we know it to work properly? Running e2e tests on openshift CI will confirm?

gbenhaim: Yes it will! I'll trigger the openshift tests.
I also verified it on my openshift and kind cluster.

gbenhaim:
Gal's Cursor here 👋

Reviewed the 3 suggested tests:

**Test 1 (idempotency — re-applying doesn't duplicate volumes):** Added. This is a useful test that verifies applying customizations twice to the same deployment with different hashed ConfigMap names doesn't create duplicate `webhook-config` volumes. The check works because `WithConfigMapVolumeUpdate` / `applyConfigMapVolumeUpdates` finds the existing volume by name and updates its ConfigMap reference in-place, never appending.

**Test 2 (empty configMapName forces Optional=true):** Skipped. When `configMapName` is empty, our code doesn't touch the volume at all — it leaves the manifest's default as-is. `optional: true` is baked into the manifest itself and we trust it. This test would fail with the current implementation because we intentionally don't actively enforce the Optional flag.

**Test 3 (webhookURLs configured removes pre-existing PAC_WEBHOOK_URL):** Skipped. Our code doesn't *remove* existing env vars — it simply doesn't *add* `PAC_WEBHOOK_URL` when `webhookURLs` are configured. The scenario this test covers (PAC_WEBHOOK_URL baked into the base manifest) is no longer realistic since we removed it from the manifest in this PR. This test would also fail with the current implementation.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
gbenhaim: @yftacherzog good catch with running the test on ocp. I needed to disable tls verification for the pac route since it used self signed cert. We do the same in the ci of infra-deployments.

yftacherzog: Humans FTW!

PR #6374, opened 2026-04-14
feat(operator): add KonfluxCLI reconciler for the CLI component

gbenhaim:
Deliver CLI tools alongside the operator so that users can download the scripts directly from the cluster. This ensures the scripts are always compatible with the operator version and the other deployed services.

Implements a new reconciler that deploys CLI component resources (namespace and configmaps) following existing operator patterns: Server-Side Apply with ownership tracking, orphan cleanup, and status condition management. The component has no configuration flags and always deploys its content. Includes envtest tests.

Assisted-By: Cursor


gbenhaim:
The followup for this PR will be to use a specific commit of the release service catalog in the `setup-release.sh` script and update it using renovate.

yftacherzog:
/ask "the new reconciler will generate configmaps that should be readable by users. With the content of this PR, which users will have what access to those new configmaps?"

yftacherzog:
@gbenhaim are we ok with this? https://github.com/konflux-ci/konflux-ci/pull/6374#issuecomment-4249963442

gbenhaim:
we should make the configmaps with the script readable by all authenticated users.

yftacherzog (approved):

gbenhaim:
> @gbenhaim are we ok with this? [#6374 (comment)](https://github.com/konflux-ci/konflux-ci/pull/6374#issuecomment-4249963442)

fixed

PR #6390, opened 2026-04-15
Remove legacy flow

gbenhaim:
Remove the legacy manifests, Github workflow and konflux deploy script. All of those were replaced by operator based installation.


yftacherzog:
```
Lines changed: 2 additions & 3246 deletions
```

and 

> ### Great, no issues found!

Nice! :)

test/resources/demo-users/user/sample-components/managed-ns1/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ns.yaml
-  - ../../../../../../konflux-ci/enterprise-contract/policies
+  - ../../../../../../operator/upstream-kustomizations/enterprise-contract/policies
```
yftacherzog, line 5: Good that we did this cleanup!

yftacherzog (approved):
lgtm. I think we should create an issue to evaluate the conditions for running e2e now. If we need to run operator e2e test for almost everything now, it might be better that we simplify the workflows to always run them instead of the more complicated logic we have at the moment both to when to run and also for how to report success if we did not run them.

avi-biton (approved):
lgtm

gbenhaim:
> lgtm. I think we should create an issue to evaluate the conditions for running e2e now. If we need to run operator e2e test for almost everything now, it might be better that we simplify the workflows to always run them instead of the more complicated logic we have at the moment both to when to run and also for how to report success if we did not run them.

https://github.com/konflux-ci/konflux-ci/issues/6393#issue-4270298137

PR #6436, opened 2026-04-19
fix: exclude Clair from EC policy

yftacherzog:
The EC policy had clair excluded from the list of required tasks, but cve.cve_results_found was still part of the policy, so as our minimal onboarding pipeline does not include Clair, EC would fail.

This change is addressing that gap by excluding cve.cve_results_found from the default EC policy.

### DIFF
--- a/operator/upstream-kustomizations/enterprise-contract/policies/enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_default.yaml
+++ b/operator/upstream-kustomizations/enterprise-contract/policies/enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_default.yaml
@@ -30,6 +30,10 @@ spec:
         # https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/known_rpm_repositories.yml
         # However, it is not possible to include just that file in the rule data.
         - rpm_repos
+        # The minimal build pipeline we use does not include Clair scan for finding CVEs.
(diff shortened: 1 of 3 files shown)


operator/upstream-kustomizations/enterprise-contract/policies/enterprise-contract-service_appstudio.redhat.com_v1alpha1_enterprisecontractpolicy_default.yaml
```diff
@@ -30,6 +30,10 @@ spec:
         # https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/known_rpm_repositories.yml
         # However, it is not possible to include just that file in the rule data.
         - rpm_repos
+        # The minimal build pipeline we use does not include Clair scan for finding CVEs.
+        # The Clair scan task is also expected to be excluded from the the list of required
+        # tasks defined in `data` section.
+        - cve.cve_results_found
```
gbenhaim, line 36: we still have the tpa scan which scans for cves. will it disable it as well?

yftacherzog: The error message associated with that failure, as well as all docs I could find mention clair scan specifically in this context, rather than something more generic for looking for CVEs.

gbenhaim:
I don't think this PR is needed anymore

PR #6455, opened 2026-04-20
fix(KONFLUX-13134): setup-release.sh fails on macOS

yftacherzog:
The script uses some commands which are not available in Bash on macOS. Modified the implementation to use commands which should be available on both Linux and macOS.

Assisted-by: Cursor


gbenhaim:
@yftacherzog can you please add a comment in the script that it should run on macos as well (which has only bash v3), so the next time an agent will edit the file it will know not to introduce breaking changes?

yftacherzog:
> @yftacherzog can you please add a comment in the script that it should run on macos as well (which has only bash v3), so the next time an agent will edit the file it will know not to introduce breaking changes?

done

gbenhaim (approved):

PR #6485, opened 2026-04-23
add new minimal pipeline to build-pipelines-config in legacy branch

rcerven:


rcerven:
@amisstea can you review

I have no idea why those CI tasks are failing

this is fix for broken e2e tests

amisstea (approved):

amisstea:
Those CI jobs can't be run from a fork but I'm not sure why we have the `/allow` instructions. They don't seem to work right now. Anyway, they're not critical to merge this.

PR #6561, opened 2026-04-28
fix: Zot authentication and docker compatibility

yftacherzog:
While testing using the internal registry, I noticed a couple of issues. First, Zot was not configured to expose Docker Registry HTTP API V2 schema2, which caused the build pipeline to fail.

Next, we originally provided dummy empty credentials to the tenant namespaces that allowed build and release pipelines to work with a registry that does not require authentication. It seems as the implementation of some of those tasks have changed and this hack no longer works. I was not able to come up with a different hack to make all tasks work without authentication.

I opted to configure authentication for the Zot registry, having the operator generate credentials on the fly, and enhance the default tenant reconciler so it can take references to secrets that should be copied to tenant namespaces. This way, we can have the operator generate credentials for the internal registry, and have the default tenant include them automatically.

Assisted-by: Cursor


yftacherzog:
@gbenhaim have a look when you get the chance.

operator/api/v1alpha1/konflux_types.go
```diff
@@ -161,6 +161,11 @@ type DefaultTenantConfig struct {
 	// Defaults to true if not specified.
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// SecretCopies configures additional secrets to copy into the default tenant namespace.
+	// These are merged with any system-managed secret copies populated by the top-level reconciler.
+	// +optional
+	SecretCopies []SecretCopySpec `json:"secretCopies,omitempty"`
```
gbenhaim: why do we need to expose this in the CR?

gbenhaim: also, we usually put here the entire spec of the sub cr.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -150,6 +179,103 @@ func (r *KonfluxDefaultTenantReconciler) applyManifests(ctx context.Context, tc
...
+			if apierrors.IsNotFound(err) {
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("get source secret %s/%s: %w", ref.SourceNamespace, ref.SourceSecretName, err)
+		}
+
+		target := &corev1.Secret{}
+		target.SetName(ref.TargetSecretName)
+		target.SetNamespace(DefaultTenantNamespace)
+		_, err := tc.CreateOrUpdate(ctx, target, func() error {
```
gbenhaim: why can't we use apply?

gbenhaim (commented):
## Review feedback

Nice work on the overall approach — the three-tier design (InternalRegistry generates credentials → Konflux wiring decides what to copy → DefaultTenant performs the copy) is clean and well-tested.

One suggestion on the `mapSourceSecretToDefaultTenant` implementation:

### Consider using the informer cache instead of the custom in-memory cache

The current approach maintains a custom `configuredSources` map with a `sync.RWMutex` that's populated during `Reconcile()` and read from the map function. This works, but it's non-standard for controller-runtime and adds complexity (mutex, cache invalidation reasoning).

The idiomatic approach for this pattern is to read the singleton CR directly from the informer cache inside the map function:

```go
func (r *KonfluxDefaultTenantReconciler) mapSourceSecretToDefaultTenant(
    ctx context.Context, obj client.Object,
) []ctrl.Request {
    dt := &konfluxv1alpha1.KonfluxDefaultTenant{}
    if err := r.Get(ctx, client.ObjectKey{Name: CRName}, dt); err != nil {
        return nil
    }
    for _, ref := range dt.Spec.SecretCopies {
        if ref.SourceNamespace == obj.GetNamespace() && ref.SourceSecretName == obj.GetName() {
            return []ctrl.Request{{NamespacedName: client.ObjectKey{Name: CRName}}}
        }
    }
    return nil
}
```

This works because:
- The controller already watches `KonfluxDefaultTenant` via `For()`, so the informer cache always has it
- `r.Get()` reads from the **local informer cache** (not the API server) — it's a cheap in-memory lookup
- No mutex needed, no custom cache, no invalidation questions
- Always consistent with what the informer knows

This would let you remove the `sourceSecretsMu`, `configuredSources` field, `setConfiguredSourceSecrets()`, `isConfiguredSourceSecret()`, and `sourceSecretKey()` — simplifying the controller considerably.

---

*Posted by Gal's Cursor*

operator/api/v1alpha1/konflux_types.go
yftacherzog: The alternative was to couple the registry implementation and the default tenant implementation so the registry "knows" which secrets are needed. I preferred having it as input so the top layer has the knowledge of what needs to be copied where rather than the underlying reoncilers.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: Not sure I understand the difference here. Do you think apply will be better?

yftacherzog:
@gbenhaim I applied some changes based on [Cursor's comment](https://github.com/konflux-ci/konflux-ci/pull/6561#pullrequestreview-4219927995).

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
gbenhaim: in general server side apply is better since it skips the "get, mutate, update" steps and just do update.

yftacherzog: ok. modified

avi-biton:
Thanks for the fix — treating the two secrets as an atomic pair (rotate both or neither) is the right approach and handles the common cases correctly.

There's still a narrow desync window worth considering. Suppose `htpasswd` gets deleted while `registryAuthSecret` still holds valid data (e.g. manual deletion, or orphan cleanup removing just one). The controller correctly sets `rotateCredentials = true`, generates a new password, and persists the new `htpasswd` — but then the `CreateOrUpdate` for `registryAuthSecret` issues a **Patch** (since the object already exists). If that Patch fails transiently (timeout, API server hiccup, etc.), the function returns an error, and on the next reconcile `hasSecretData` returns `true` for **both** secrets — so `rotateCredentials = false` and the mismatch is never self-healed without manual intervention.

Two ways to fix this:

1. **Detect desync explicitly**: after reading both secrets, call `bcrypt.CompareHashAndPassword` against the password extracted from the dockerconfig. If they don't match, force rotation regardless of whether both have data.

2. **Persist the plaintext password**: store it as an additional key in one of the secrets (e.g. `htpasswd.Data["password"]`). The reconciler can then always re-derive the dockerconfig deterministically from the stored password, sidestepping the atomicity concern entirely.

Option 1 is a smaller change; option 2 removes the whole class of atomicity bugs.

---
*Posted by Avi's Cursor*

gbenhaim (approved):
LGTM. The revision addressed the previous feedback well — the custom mutex-guarded cache is replaced with an idiomatic informer-cache read in the map function, and the watch scope is properly narrowed with a namespace+name predicate. Clean and simple.

*Approved by Gal's Cursor*

yftacherzog:
> Thanks for the fix — treating the two secrets as an atomic pair (rotate both or neither) is the right approach and handles the common cases correctly.
> 
> There's still a narrow desync window worth considering. Suppose `htpasswd` gets deleted while `registryAuthSecret` still holds valid data (e.g. manual deletion, or orphan cleanup removing just one). The controller correctly sets `rotateCredentials = true`, generates a new password, and persists the new `htpasswd` — but then the `CreateOrUpdate` for `registryAuthSecret` issues a **Patch** (since the object already exists). If that Patch fails transiently (timeout, API server hiccup, etc.), the function returns an error, and on the next reconcile `hasSecretData` returns `true` for **both** secrets — so `rotateCredentials = false` and the mismatch is never self-healed without manual intervention.
> 
> Two ways to fix this:
> 
> 1. **Detect desync explicitly**: after reading both secrets, call `bcrypt.CompareHashAndPassword` against the password extracted from the dockerconfig. If they don't match, force rotation regardless of whether both have data.
> 2. **Persist the plaintext password**: store it as an additional key in one of the secrets (e.g. `htpasswd.Data["password"]`). The reconciler can then always re-derive the dockerconfig deterministically from the stored password, sidestepping the atomicity concern entirely.
> 
> Option 1 is a smaller change; option 2 removes the whole class of atomicity bugs.
> 
> _Posted by Avi's Cursor_

I'm tempted to avoid fixing this for now. It looks like something that would not be obvious to troubleshoot suppose it happens, but I'm not too convinced it's likely to happen without manual action around those secrets. This feature is not expected to be used in production, so it's just dev env which we're aiming for.

I'm going to avoid fixing it for now.

avi-biton (approved):

PR #6585, opened 2026-04-29
chore: update build and conforma

yftacherzog:
With this, the build pipeline will drop the use of Clair scan

Assisted-by: Cursor


operator/pkg/manifests/build-service/manifests.yaml
```diff
@@ -382,7 +382,6 @@ metadata:
   name: build-service-controller-manager
   namespace: build-service
 spec:
-  progressDeadlineSeconds: 2147483647
```
gbenhaim, line 385: why it was removed? same question about the removals below

gbenhaim (approved):

PR #6586, opened 2026-04-29
fix: regenerate build manifest

yftacherzog:
While patching a failing automated PR for build-service it seems as we only propagated some of the changes in the upstream kusotmizations but not all. This commit regenerates the build manifest with the latest changes.

Assisted-by: Cursor


yftacherzog:
The manifests was generated incorrectly. Closing.

PR #6608, opened 2026-04-30
fix: bump docker-build-oci-ta-min bundle digest to latest one

tisutisu:
bump docker-build-oci-ta-min bundle digest to latest one which contains [the fix](https://github.com/konflux-ci/build-definitions/pull/3486)

### DIFF
--- a/konflux-ci/build-service/core/build-pipeline-config.yaml
+++ b/konflux-ci/build-service/core/build-pipeline-config.yaml
@@ -18,5 +18,5 @@ data:
     - name: tekton-bundle-builder-oci-ta
       bundle: quay.io/konflux-ci/tekton-catalog/pipeline-tekton-bundle-builder-oci-ta@sha256:f26d5fea5a12de33b9ca5454992afb4707120a5cf9ed076a6477c2cee55f141d
     - name: docker-build-oci-ta-min
-      bundle: quay.io/konflux-ci/tekton-catalog/pipeline-docker-build-oci-ta-min@sha256:5596db15d773ea5892b0c6a3d701a19757fed875d20cc862b3bcfb9fcf32523b
+      bundle: quay.io/konflux-ci/tekton-catalog/pipeline-docker-build-oci-ta-min@sha256:151372f879bc7e7bc2870e1134ed6d0c3aed6bea7ab40bfbf3903a855c0e0696
       description: minimal version of the docker-build-oci-ta pipeline, which requires less compute resources. In addition, it doesn't contain tasks which require user provided secrets.


yftacherzog:
@tisutisu where do we see those failures? I don't think we're seeing them in our e2e tests. Looking at the upstream change, it adds 768Mi to the resource requests. That might be quite a lot for potential upstream users.

cc @gbenhaim

tisutisu:
> @tisutisu where do we see those failures? I don't think we're seeing them in our e2e tests. Looking at the upstream change, it adds 768Mi to the resource requests. That might be quite a lot for potential upstream users.
> 
> cc @gbenhaim

we have observed `OOMKilled` failures while running CI on [this PR](https://github.com/konflux-ci/build-service/pull/591), artifacts for the failure run can be fetched using `oras pull quay.io/konflux-test-storage/konflux-team/build-service:konflux-e2e-build-service-9582t` , from the downloaded artifacts find the alerts in `konflux-artifacts/events.json`
we are building container for this [sample repo](https://github.com/redhat-appstudio-qe/sample-multi-component/tree/main/python-component) , which I think is light weight. I think minimal build pipeline should support building of light weight components like this.

gbenhaim:
@tisutisu I suggest that the e2e test will use one of the following components (whic are used in the CI of this repo and in TSF). They work with the current memory requests and limits.

 https://github.com/konflux-ci/testrepo
 https://github.com/konflux-ci/sample-component-golang

gbenhaim:
> > whic are used in th
> 
> @gbenhaim we moved all our build related e2e-tests to use minimal pipeline with [this change](https://github.com/konflux-ci/e2e-tests/commit/0b234f3a09186969ff3be5f5f46f0d0d33894e0f), its failing in one of the build e2e-tests in our build-service repo CI.

I understand. I suggest to change the test to use up to date sample component which fits the resource the min pipeline has, and not increasing the resources of the pipeline.

mmorhun:
@gbenhaim I agree with you that it's better to align tests than increase the resources, however, we need to have e2e tests that cover certain scenarios. Currently, it usually fails on python one, but we cannot just replace it with a golang sample because it tests python specific things, I think.
@tisutisu does the test under question uses python specific parts?

mmorhun:
We've found better solution, closing this PR as not needed.

gbenhaim:
@tisutisu did you revert the resource request change to the min pipeline?

tisutisu:
@gbenhaim yes we have reverted the change with [this PR](https://github.com/konflux-ci/build-definitions/pull/3488)

PR #6646, opened 2026-05-03
fix(KONFLUX-13599): CI fails for go linter

yftacherzog:
It seems as the linter version was bumped as we're not pinning it.

Assisted-by: Cursor


yftacherzog:
> [correctness] operator/.golangci-lint-version — Version jumps from v2.8.0 to v2.11.4. This is a significant minor-version bump. The PR is specifically about fixing CI by pinning, so this is expected, but worth noting in case the new linter version surfaces new findings in existing code.

In reality, we're downgrading the linter version that was running in CI.
I'm not sure if whatever the makfile defines was used very widely.

avi-biton (approved):

PR #6654, opened 2026-05-03
test: don't hang waiting for cleanup

yftacherzog:
When running e2e tests locally, the cleanup stage tends to hang. This commit makes it give up eventually, so the tests eventually finish even if cleanup hangs, not affecting the test results.

Assisted-by: Cursor

### DIFF
--- a/test/go-tests/tests/conformance/conformance.go
+++ b/test/go-tests/tests/conformance/conformance.go
@@ -31,6 +31,17 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// conformanceCleanupBudget returns the max wall time to wait for AfterAll cleanup.
+// Set E2E_CLEANUP_TIMEOUT to a Go duration (e.g. 90s, 1m30s); invalid or empty values use the default.
+func conformanceCleanupBudget() time.Duration {
+	if s := strings.TrimSpace(os.Getenv("E2E_CLEANUP_TIMEOUT")); s != "" {
+		if d, err := time.ParseDuration(s); err == nil && d > 0 {
+			return d
+		}
+	}
+	return conformanceCleanupMaxDuration
+}
+
 var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamKonfluxTestLabel), func() {
 	defer ginkgo.GinkgoRecover()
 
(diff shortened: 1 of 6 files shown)


test/go-tests/tests/conformance/conformance.go
```diff
@@ -87,16 +98,26 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
-				})
-				cleanupWithRetry("delete base branch", func() error {
-					return fw.AsKubeAdmin.CommonController.GitHub.DeleteRef(componentRepositoryName, componentNewBaseBranch)
-				})
-				cleanupWithRetry("cleanup webhooks", func() error {
-					return build.CleanupWebhooks(fw, componentRepositoryName)
-				})
+				budget := conformanceCleanupBudget()
+				done := make(chan struct{})
+				go func() {
```
gbenhaim: I think that the way to achieve it is to pass a context with deadline to each of the cleanup calls.

gbenhaim (approved):

test/go-tests/tests/conformance/setup.go
```diff
@@ -212,14 +211,41 @@ func dumpDiagnostics(hub *framework.ControllerHub, componentName, appName, names
...
-		klog.Warningf("conformance cleanup: %s failed after retries: %v", description, err)
+// cleanupWithRetry runs fn until it returns nil, ctx is done, or a 30s per-step retry budget elapses.
+// fn should use ctx for outbound calls so they honor the overall cleanup deadline.
+func cleanupWithRetry(ctx context.Context, description string, fn func() error) {
+	const maxStep = 30 * time.Second
+	const poll = 5 * time.Second
+	stepStart := time.Now()
+	var lastErr error
+	for {
+		if err := ctx.Err(); err != nil {
```
gbenhaim, line 222: /ask I think that this condition is useless since we already handle cancellation in 	case <-ctx.Done():
line 245

gbenhaim: /ask I think that this condition is useless since we already handle cancellation in case <-ctx.Done():

gbenhaim: @yftacherzog how do I call the /ask command in Qodo

yftacherzog: @gbenhaim are you ok with merging this as is, or want some changes?

PR #6676, opened 2026-05-04
UI proxy remove long lived token

filariow:
We want to migrate from Long-Lived Tokens to Short-Lived Tokens.

Unfortunately, NGINX is not automatically reloading its configuration, so we need to introduce some logic to rebuild the configuration and hot reload it.

This PR introduces a mechanism to regenerate and reload the configuration with a short-lived token provided by Kubernetes. A time based reload approach was preferred to an `inotify`-based one for its simplicity.

"The kubelet proactively requests rotation for the token if it is older than 80% of its total time-to-live (TTL), or if the token is older than 24 hours" \[1\]. 
With the proposed the default values, `120 seconds` will be available to the reload mechanism to process the new value and, in the worst case, the reload mechanism will require `36 seconds`:
* `generate-loop`: 10 seconds sleep , ~8 seconds retry producing the token -> `18 seconds`
* `hot-reload`: 10 seconds sleep, ~8 seconds retry hot reloading. -> `18 seconds`

References
* \[1\]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#launch-a-pod-using-service-account-token-projection

### DIFF
--- a/operator/pkg/manifests/ui/manifests.yaml
+++ b/operator/pkg/manifests/ui/manifests.yaml
@@ -263,6 +263,114 @@ metadata:
 ---
 apiVersion: v1
 data:
+  proxy-nginx-generate-loop-probe.sh: |
+    #!/bin/bash
+
+    set -euo pipefail
+
+    AUTH_CONF_FILE=/mnt/nginx-generated-config/auth.conf
+    AUTH_CONF_NEW_FILE=/mnt/nginx-generated-config/auth.conf.new
+
+    { [ -f "${AUTH_CONF_NEW_FILE}" ] && [ $(( $(date +%s) - $(stat -c %Y "${AUTH_CONF_NEW_FILE}") )) -lt 60 ]; } || \
+      { [ -f "${AUTH_CONF_FILE}" ] && [ $(( $(date +%s) - $(stat -c %Y "${AUTH_CONF_FILE}") )) -lt 60 ]; }
+  proxy-nginx-generate-loop.sh: |
+    #!/bin/bash
+
+    set -euo pipefail
+
+    RETRY_INTERVAL=10
+    TOKEN_FILEPATH=/var/run/secrets/konflux-ci.dev/serviceaccount/token
+    AUTH_CONF_FILE=/mnt/nginx-generated-config/auth.conf.new
+    AUTH_CONF_TMP_FILE=/mnt/nginx-generated-config/auth.conf.new.tmp
+    AUTH_CONF_TEMPLATE_FILE=/mnt/nginx-templates/auth.conf
+
+    produceToken() (
+      # Copy the auth.conf template and replace the bearer token
+      token=$(cat "${TOKEN_FILEPATH}")
+
+      # Produce a tmp file
+      sed "s/__BEARER_TOKEN__/${token}/" "${AUTH_CONF_TEMPLATE_FILE}" > "${AUTH_CONF_TMP_FILE}"
+      chmod 640 "${AUTH_CONF_TMP_FILE}"
+
+      # Rename (atomic) the file to avoid sync issues
+      mv "${AUTH_CONF_TMP_FILE}" "${AUTH_CONF_FILE}"
+    )
+
+    produceTokenWithRetry() (
+      produceToken || \
+        { sleep 3; produceToken; } || \
+        { sleep 5; produceToken; }
+    )
+
+    while produceTokenWithRetry; do sleep "${RETRY_INTERVAL}"; done
+
+    echo "loop broke, crashing"
+    exit 1
+kind: ConfigMap
+metadata:
+  name: proxy-nginx-generate-loop
+  namespace: konflux-ui
+---
+apiVersion: v1
+data:
+  proxy-nginx-run.sh: |
+    #!/bin/bash
+
+    set -euo pipefail
+
+    NGINX_AUTH_CONF_FILE='/mnt/nginx-generated-config/auth.conf'
+    NGINX_NEW_AUTH_CONF_FILE='/mnt/nginx-generated-config/auth.conf.new'
+
+    NGINX_CONF_FILE='/etc/nginx/nginx.conf'
+    RETRY_INTERVAL=10
+
+    # test configuration
+    nginx -g "daemon off;" -c "${NGINX_CONF_FILE}" -t
+
+    # run the hot-reload loop in background
+    (
+      # wait for nginx to start before first reload attempt
+      while [ ! -f /run/nginx.pid ]; do sleep 1; done
+
+      # retries hot reloading the nginx configuration multiple
+      # times with increasing timeouts before returning the error
+      reloadWithRetry() {
+        if [ -f "${NGINX_NEW_AUTH_CONF_FILE}" ] && \
+          ! cmp -s "${NGINX_NEW_AUTH_CONF_FILE}" "${NGINX_AUTH_CONF_FILE}"; then
+          # Move (atomic) the new configuration and reload it in NGINX
+          mv "${NGINX_NEW_AUTH_CONF_FILE}" "${NGINX_AUTH_CONF_FILE}" && \
+            {
+              nginx -s reload || \
+                { sleep 3; nginx -s reload; } || \
+                { sleep 5; nginx -s reload; }
+            }
+        fi
+      }
+
+      # hot reload infinite loop
+      while reloadWithRetry; do sleep "${RETRY_INTERVAL}"; done
+
+      # not supposed to be terminated
+      echo "loop broke, crashing"
+      exit 1
+    ) &
+
+    # run the nginx server in background
+    (
+      nginx -g "daemon off;" -c "${NGINX_CONF_FILE}"
+      echo "nginx crashed"
+      exit 1
+    ) &
+
+    # wait for any of the background tasks to crash
+    wait -n
+kind: ConfigMap
+metadata:
+  name: proxy-nginx-run
+  namespace: konflux-ui
+---
+apiVersion: v1
+data:
   auth.conf: |
     # Auth configuration with impersonate headers
     auth_request_set $user  $upstream_http_x_auth_request_email;
@@ -282,15 +390,6 @@ kind: ConfigMap
 metadata:
   name: proxy-nginx-templates-dt292585th
   namespace: konflux-ui
----
-apiVersion: v1
-kind: Secret
-metadata:
-  annotations:
-    kubernetes.io/service-account.name: proxy
-  name: proxy
-  namespace: konflux-ui
-type: kubernetes.io/service-account-token
 ---
 apiVersion: v1
 kind: Service
@@ -427,12 +526,45 @@ spec:
         app: proxy
     spec:
       containers:
-      - command:
-        - nginx
-        - -g
-        - daemon off;
-        - -c
-        - /etc/nginx/nginx.conf
+      - args:
+        - /var/run/scripts/konflux-ci.dev/proxy-nginx-generate-loop.sh
+        command:
+        - bash
+        image: quay.io/konflux-ci/konflux-ui@sha256:db1ed966d9d0d89a37b31cdec23684e653cd11e53387c549e6d381f4ba7811e1
+        livenessProbe:
+          exec:
+            args:
+            - /var/run/scripts/konflux-ci.dev/proxy-nginx-generate-loop-probe.sh
+            command:
+            - bash
+          failureThreshold: 3
+          periodSeconds: 30
+        name: generate-nginx-configs-loop
+        resources:
+          limits:
+            cpu: 50m
+            memory: 128Mi
+          requests:
+            cpu: 10m
+            memory: 64Mi
+        securityContext:
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+        volumeMounts:
+        - mountPath: /var/run/scripts/konflux-ci.dev/
+          name: proxy-generate-loop-script
+          readOnly: true
+        - mountPath: /var/run/secrets/konflux-ci.dev/serviceaccount
+          name: kube-api-token
+          readOnly: true
+        - mountPath: /mnt/nginx-generated-config
+          name: nginx-generated-config
+        - mountPath: /mnt/nginx-templates
+          name: nginx-templates
+      - args:
+        - /var/run/scripts/konflux-ci.dev/proxy-nginx-run.sh
+        command:
+        - bash
         image: registry.access.redhat.com/ubi9/nginx-124@sha256:6cfc252017036e3a6b5eaff546ac72d1a2fe8d3aec1d3a4305f40cded79331e0
         livenessProbe:
           failureThreshold: 3
@@ -473,6 +605,9 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
         volumeMounts:
+        - mountPath: /var/run/scripts/konflux-ci.dev/
+          name: proxy-nginx-run-script
+          readOnly: true
         - mountPath: /etc/nginx/nginx.conf
           name: proxy
           readOnly: true
@@ -548,7 +683,7 @@ spec:
           set -e
 
           # Copy the auth.conf template and replace the bearer token
-          token=$(cat /mnt/api-token/token)
+          token=$(cat /var/run/secrets/konflux-ci.dev/serviceaccount/token)
           sed "s/__BEARER_TOKEN__/$token/" /mnt/nginx-templates/auth.conf > /mnt/nginx-generated-config/auth.conf
           chmod 640 /mnt/nginx-generated-config/auth.conf
 
@@ -578,12 +713,13 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
         volumeMounts:
+        - mountPath: /var/run/secrets/konflux-ci.dev/serviceaccount
+          name: kube-api-token
+          readOnly: true
         - mountPath: /mnt/nginx-generated-config
           name: nginx-generated-config
         - mountPath: /mnt/nginx-templates
           name: nginx-templates
-        - mountPath: /mnt/api-token
-          name: api-token
         - mountPath: /mnt/nginx-generated-location-config
           name: nginx-generated-location-config
       serviceAccountName: proxy
@@ -595,6 +731,27 @@ spec:
         topologyKey: topology.kubernetes.io/zone
         whenUnsatisfiable: ScheduleAnyway
       volumes:
+      - configMap:
+          defaultMode: 488
+          items:
+          - key: proxy-nginx-generate-loop.sh
+            path: proxy-nginx-generate-loop.sh
+          name: proxy-nginx-generate-loop
+        name: proxy-generate-loop-script
+      - configMap:
+          defaultMode: 488
+          items:
+          - key: proxy-nginx-run.sh
+            path: proxy-nginx-run.sh
+          name: proxy-nginx-run
+        name: proxy-nginx-run-script
+      - name: kube-api-token
+        projected:
+          defaultMode: 420
+          sources:
+          - serviceAccountToken:
+              expirationSeconds: 600
+              path: token
       - configMap:
           defaultMode: 420
           items:
@@ -619,9 +776,6 @@ spec:
         name: nginx-generated-config
       - emptyDir: {}
         name: nginx-generated-location-config
-      - name: api-token
-        secret:
-          secretName: proxy
       - emptyDir: {}
         name: static-content
       - name: segment-bridge-config
--- a/operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
+++ b/operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+set -euo pipefail
+
+RETRY_INTERVAL=10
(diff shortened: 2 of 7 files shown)


operator/upstream-kustomizations/ui/core/proxy/proxy.yaml
```diff
@@ -98,14 +99,39 @@ spec:
             cpu: 10m
             memory: 64Mi
       containers:
+      - name: generate-nginx-configs-loop
+        image: quay.io/konflux-ci/konflux-ui
+        command:
+          - sh
```
gbenhaim: **Suggestion**: Both this container and the `nginx` container below use `sh` as the command, but the scripts have `#!/bin/bash` shebangs and `proxy-nginx-run.sh` uses `wait -n` (a bash 4.3+ extension, not POSIX). On UBI9 `/bin/sh` is bash so it works today, but it's fragile — a base image change could break it silently.

Consider using `bash` instead:

```yaml
command:
  - bash
```

This makes the intent explicit and matches the shebangs.

operator/upstream-kustomizations/ui/core/proxy/proxy.yaml
```diff
@@ -98,14 +99,39 @@ spec:
...
+        securityContext:
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+        resources:
+          limits:
+            cpu: 50m
+            memory: 128Mi
+          requests:
+            cpu: 10m
+            memory: 64Mi
```
gbenhaim, line 136: **Minor**: The `generate-nginx-configs-loop` container has no liveness or readiness probe. If this container hangs, the token won't be refreshed and nginx will continue serving with an expired bearer token — silently.

A simple file-age-based exec probe could detect this:

```yaml
livenessProbe:
  exec:
    command:
    - sh
    - -c
    - "[ $(( $(date +%s) - $(stat -c %Y /mnt/nginx-generated-config/auth.conf) )) -lt 60 ]"
  periodSeconds: 30
  failureThreshold: 3
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e
+
+NGINX_CONF_FILE='/etc/nginx/nginx.conf'
+RETRY_INTERVAL=10
+
+# test configuration
+nginx -g "daemon off;" -c "${NGINX_CONF_FILE}" -t
```
gbenhaim, line 12: **Nit**: The `-t` flag only tests config and exits — the `-g "daemon off;"` directive is harmless but unnecessary for a config test. Could simplify to:

```bash
nginx -c "${NGINX_CONF_FILE}" -t
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# run the hot-reload loop in background
+(
```
gbenhaim, line 15: **Minor — race condition**: The hot-reload loop starts immediately in parallel with nginx. On the first iteration, `nginx -s reload` may fire before the nginx master process has its PID file ready. The retry logic (3s + 5s) should cover typical startup, but if all 3 attempts fail the loop breaks and the container crashes unnecessarily.

Consider adding a readiness wait before entering the reload loop:

```bash
# wait for nginx to start before first reload attempt
while [ ! -f /run/nginx.pid ]; do sleep 1; done
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -e
```
gbenhaim: **Nit**: Per project conventions, shell scripts should use `set -euo pipefail`. While there are no pipes or undefined-variable risks in the current script, using the full form would be more consistent. Same applies to `proxy-nginx-run.sh`.

gbenhaim (commented):

yftacherzog (commented):
/ask "are the new scripts compatible with bash on macos?"

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
```diff
@@ -0,0 +1,37 @@
...
+  # retries hot reloading the nginx configuration multiple
+  # times with increasing timeouts before returning the error
+  reloadWithRetry() {
+    nginx -s reload || \
+      { sleep 3; nginx -s reload; } || \
+      { sleep 5; nginx -s reload; }
+  }
+
+  # hot reload infinite loop
+  while reloadWithRetry; do sleep "${RETRY_INTERVAL}"; done
```
gbenhaim, line 35: **Suggestion — skip unnecessary reloads**: Currently the reload loop fires `nginx -s reload` every 10 seconds unconditionally. On a high-traffic cluster, each reload spawns new workers and gracefully drains old ones — that's ~8,640 unnecessary reloads per day when the token only rotates ~180 times.

Consider having the reload loop keep a copy of the last-reloaded config and only reload when it actually changes, using a simple `cmp`:

```bash
last_conf="/tmp/auth.conf.last"
while true; do
  if ! cmp -s /mnt/nginx-generated-config/auth.conf "$last_conf" 2>/dev/null; then
    reloadWithRetry
    cp /mnt/nginx-generated-config/auth.conf "$last_conf"
  fi
  sleep "${RETRY_INTERVAL}"
done
```

This keeps the reload loop self-contained (no dependency on the generate-loop's write behavior), and the generate-loop can keep writing unconditionally so its mtime stays fresh for liveness probes.

_This comment was posted by **Gal's Cursor**._

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
filariow: if the `daemon` directive is also present in the nginx conf file, it will fail with a `directive duplicate` error:

```
sh-5.1$ nginx -t -g 'daemon on;'
nginx: [emerg] "daemon" directive is duplicate in /etc/nginx/nginx.conf:5
nginx: configuration file /etc/nginx/nginx.conf test failed
```

operator/pkg/manifests/ui/manifests.yaml
```diff
@@ -262,6 +262,91 @@ metadata:
   namespace: konflux-ui
 ---
 apiVersion: v1
+data:
```
yftacherzog, line 265: /ask "are the new scripts compatible with bash on macos?"

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
yftacherzog, line 1: /ask "is the new script compatible with bash on macos?"

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
yftacherzog, line 1: /ask "is the new script compatible with bash on macos?"

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
yftacherzog:
```bash
#!/bin/bash

set -e

RETRY_INTERVAL=10
TOKEN_FILEPATH=/var/run/secrets/konflux-ci.dev/serviceaccount/token
AUTH_CONF_FILE=/mnt/nginx-generated-config/auth.conf
AUTH_CONF_TMP_FILE=/mnt/nginx-generated-config/auth.conf.tmp
AUTH_CONF_TEMPLATE_FILE=/mnt/nginx-templates/auth.conf

produceToken() (
  # Copy the auth.conf template and replace the bearer token
  token=$(cat "${TOKEN_FILEPATH}")

  # Produce a tmp file
  sed "s/__BEARER_TOKEN__/${token}/" "${AUTH_CONF_TEMPLATE_FILE}" > "${AUTH_CONF_TMP_FILE}"
  chmod 640 "${AUTH_CONF_TMP_FILE}"

  # Rename (atomic) the file to avoid sync issues
  mv "${AUTH_CONF_TMP_FILE}" "${AUTH_CONF_FILE}"
)

produceTokenWithRetry() (
  produceToken || \
    { sleep 3; produceToken; } || \
    { sleep 5; produceToken; }
)

while produceTokenWithRetry; do sleep "${RETRY_INTERVAL}"; done

echo "loop broke, crashing"
exit 1
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
yftacherzog:
```bash
#!/bin/bash

set -e

NGINX_CONF_FILE='/etc/nginx/nginx.conf'
RETRY_INTERVAL=10

# test configuration
nginx -g "daemon off;" -c "${NGINX_CONF_FILE}" -t

# run the hot-reload loop in background
(
  # retries hot reloading the nginx configuration multiple
  # times with increasing timeouts before returning the error
  reloadWithRetry() {
    nginx -s reload || \
      { sleep 3; nginx -s reload; } || \
      { sleep 5; nginx -s reload; }
  }

  # hot reload infinite loop
  while reloadWithRetry; do sleep "${RETRY_INTERVAL}"; done

  # not supposed to be terminated
  echo "loop broke, crashing"
  exit 1
) &

# run the nginx server in background
(
  nginx -g "daemon off;" -c "${NGINX_CONF_FILE}"
  echo "nginx crashed"
  exit 1
) &

# wait for any of the background tasks to crash
wait -n
```

yftacherzog: /ask Is the script below compatible with bash on macos?
```
#!/bin/bash

set -e

NGINX_CONF_FILE='/etc/nginx/nginx.conf'
RETRY_INTERVAL=10

# test configuration
nginx -g "daemon off;" -c "${NGINX_CONF_FILE}" -t

# run the hot-reload loop in background
(
  # retries hot reloading the nginx configuration multiple
  # times with increasing timeouts before returning the error
  reloadWithRetry() {
    nginx -s reload || \
      { sleep 3; nginx -s reload; } || \
      { sleep 5; nginx -s reload; }
  }

  # hot reload infinite loop
  while reloadWithRetry; do sleep "${RETRY_INTERVAL}"; done

  # not supposed to be terminated
  echo "loop broke, crashing"
  exit 1
) &

# run the nginx server in background
(
  nginx -g "daemon off;" -c "${NGINX_CONF_FILE}"
  echo "nginx crashed"
  exit 1
) &

# wait for any of the background tasks to crash
wait -n
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
yftacherzog: /ask is the script below compatible with with bash on macos?
```
#!/bin/bash

set -e

RETRY_INTERVAL=10
TOKEN_FILEPATH=/var/run/secrets/konflux-ci.dev/serviceaccount/token
AUTH_CONF_FILE=/mnt/nginx-generated-config/auth.conf
AUTH_CONF_TMP_FILE=/mnt/nginx-generated-config/auth.conf.tmp
AUTH_CONF_TEMPLATE_FILE=/mnt/nginx-templates/auth.conf

produceToken() (
  # Copy the auth.conf template and replace the bearer token
  token=$(cat "${TOKEN_FILEPATH}")

  # Produce a tmp file
  sed "s/__BEARER_TOKEN__/${token}/" "${AUTH_CONF_TEMPLATE_FILE}" > "${AUTH_CONF_TMP_FILE}"
  chmod 640 "${AUTH_CONF_TMP_FILE}"

  # Rename (atomic) the file to avoid sync issues
  mv "${AUTH_CONF_TMP_FILE}" "${AUTH_CONF_FILE}"
)

produceTokenWithRetry() (
  produceToken || \
    { sleep 3; produceToken; } || \
    { sleep 5; produceToken; }
)

while produceTokenWithRetry; do sleep "${RETRY_INTERVAL}"; done

echo "loop broke, crashing"
exit 1
```

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
yftacherzog: @filariow can you look into that comment above? We need mac users to be able to run this.

gbenhaim: > The shebang `#!/bin/bash` will _run_ on macOS, because macOS ships `/bin/bash` by default. However, there are two compatibility concerns:
> 
> * **Bash version**: macOS typically ships an older Bash (usually **3.2** on many versions). If this script uses newer Bash features (e.g., `mapfile`, associative arrays `declare -A`, `[[ … =~ … ]]` nuances, `globstar`, etc.), it may **break on macOS**.
> * **Path flexibility**: `#!/bin/bash` hard-codes the path. If someone uses a newer Bash via Homebrew (e.g., `/opt/homebrew/bin/bash`), this won’t pick it up.
> 
> Recommendation:
> 
> * Prefer `#!/usr/bin/env bash` if you want to use the user’s `bash` on PATH (often Homebrew Bash on macOS).
> * Or, if you must keep `#!/bin/bash`, ensure the script is limited to Bash 3.2-compatible syntax.

The script is going to run in a pod

yftacherzog: This is not one of those scripts that users will fetch to their terminal?

filariow: no, they are both executed in containers in the `proxy`'s pod. They'll be stored in a ConfigMap and mounted in the containers.

operator/upstream-kustomizations/ui/core/proxy/proxy.yaml
filariow: done in a3bcdfba2150d48e7025e7d36c155e9e0a8916ca

operator/upstream-kustomizations/ui/core/proxy/proxy.yaml
filariow: done in f837229195ec2422d7d842ae59789e4ed2f9f4aa

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
filariow: done in dfb67a7b84cf58f3503986ac07917002c4824f5b

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-generate-loop.sh
filariow: done in 2f94c855a3eca55f7be98b8967bfda60d9d5c024

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
filariow: implemented something similar in 42707cd29c7a97feec35a44148faf35f7a3c0415.

The generate loop produces the `auth.conf.new` file. Anytime the reload loop finds the file it renames it as `auth.conf` and hot reload the NGINX configuration. If no `auth.conf.new` is found by the reload loop, the hot reload procedure is skipped.

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
yftacherzog: alright

yftacherzog:
@filariow can you squash the commits?

operator/pkg/manifests/ui/manifests.yaml
```diff
@@ -526,7 +660,7 @@ spec:
         - -R
         - /opt/app-root/src/.
         - /mnt/static-content/
-        image: quay.io/konflux-ci/konflux-ui@sha256:5a6df7f522f3686ab1d332f4196d8b8a8391f5cbce5ceb35978c3318659a9322
+        image: quay.io/konflux-ci/konflux-ui@sha256:db1ed966d9d0d89a37b31cdec23684e653cd11e53387c549e6d381f4ba7811e1
```
filariow, line 664: this change was introduced by https://github.com/konflux-ci/konflux-ci/commit/32204c437c92bf845d10fc554ded8af472ca72a1. @gbenhaim @yftacherzog do we have a CI to ensure `operator/upstream-kustomizations` and `operator/pkg/manifest` don't get misaligned? can we regenerate the manifests directly in the bot's PRs?

Also, I'm not getting what's happening in the PR creating the commit above (#6684). It looks like the proposed/approved tag was `df5ed77174266d1b34be95103ec321d8fc8ba3df52906aa16d8f849ccd679a4d` (cf https://github.com/konflux-ci/konflux-ci/pull/6684/changes), but the merged one was `db1ed966d9d0d89a37b31cdec23684e653cd11e53387c549e6d381f4ba7811e1`. The only thing I can think of is that the branch was updated while in the merge queue. WDYT?

yftacherzog: > do we have a CI to ensure operator/upstream-kustomizations and operator/pkg/manifest don't get misaligned? 

We do. It's running daily and can also be triggered manually.

> an we regenerate the manifests directly in the bot's PRs?

If you mean renovate/mintmaker then probably not. Or at least it's not trivial. For repos that also host kustomizations, we align the update in git with the image produced for it, and the automation that generates the manifests knows how to do it, but for UI, we only track images then it's not as trivial. Instead we let renovate update the upstream manifests and then the daily job rebuilds the operator manifests.

So, if the manifest is good after the renovate update, the misalignment will be solved in about a day. If it causes regressions then the skew could get bigger.

yftacherzog: > Also, I'm not getting what's happening in the PR creating the commit above (https://github.com/konflux-ci/konflux-ci/pull/6684). It looks like the proposed/approved tag was df5ed77174266d1b34be95103ec321d8fc8ba3df52906aa16d8f849ccd679a4d (cf https://github.com/konflux-ci/konflux-ci/pull/6684/changes), but the merged one was db1ed966d9d0d89a37b31cdec23684e653cd11e53387c549e6d381f4ba7811e1. The only thing I can think of is that the branch was updated while in the merge queue. WDYT?

That is indeed strange. Not sure what's going on here. Where do we need it?

yftacherzog: I think there were some github incidents lately, maybe also related to merge queues. Might be something related to that. It is strange...

filariow: ok, got it. thanks for the explanation 😄

gbenhaim (commented):

operator/upstream-kustomizations/ui/core/proxy/proxy-nginx-run.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# run the hot-reload loop in background
+(
+  # wait for nginx to start before first reload attempt
+  while [ ! -f /run/nginx.pid ]; do sleep 1; done
+
+  # retries hot reloading the nginx configuration multiple
+  # times with increasing timeouts before returning the error
+  reloadWithRetry() {
+    if [ -f "${NGINX_NEW_AUTH_CONF_FILE}" ]; then
```
gbenhaim: **Suggestion — reload optimization is incomplete**: The file-signaling approach is a great idea, but it doesn't actually skip unnecessary reloads yet. The generate-loop writes `auth.conf.new` unconditionally every 10 seconds, so `[ -f auth.conf.new ]` is always true and nginx still reloads every ~10s.

To complete the optimization, add a `cmp` check before moving/reloading:

```bash
reloadWithRetry() {
    if [ -f "${NGINX_NEW_AUTH_CONF_FILE}" ]; then
      if ! cmp -s "${NGINX_NEW_AUTH_CONF_FILE}" "${NGINX_AUTH_CONF_FILE}" 2>/dev/null; then
        # Move (atomic) the new configuration and reload it in NGINX
        mv "${NGINX_NEW_AUTH_CONF_FILE}" "${NGINX_AUTH_CONF_FILE}" && \
          {
            nginx -s reload || \
              { sleep 3; nginx -s reload; } || \
              { sleep 5; nginx -s reload; }
          }
      else
        rm -f "${NGINX_NEW_AUTH_CONF_FILE}"
      fi
    fi
  }
```

This way:
- Generate-loop keeps writing `.new` unconditionally (liveness probe stays happy since the file mtime stays fresh)
- Reload-loop compares `.new` against current `auth.conf` — only reloads when content actually changed (~every 8 min on token rotation)
- `rm -f` cleans up the identical `.new` file so it doesn't accumulate

_This comment was posted by **Gal's Cursor**._

filariow: right, I forgot to cmp the files

filariow: fixed

gbenhaim:
Replaced by https://github.com/konflux-ci/konflux-ci/pull/6780

PR #6728, opened 2026-05-07
feat(KFLUXVNGD-966): add typed pipeline timeout fields to KonfluxIntegrationService

avi-biton:
Promote PIPELINE_TIMEOUT, TASKS_TIMEOUT, and FINALLY_TIMEOUT from
generic env var examples to first-class typed fields on
KonfluxIntegrationServiceSpec.

Previously, users could only set these via the unvalidated
integrationControllerManager.manager.env list, which provided no
schema validation and required knowledge of the exact env var names.

Changes:
- Add PipelineTimeout, TasksTimeout, FinallyTimeout string fields to
  KonfluxIntegrationServiceSpec, each validated against a duration
  pattern (e.g. "6h", "4h30m").
- Map the fields to their corresponding env vars (PIPELINE_TIMEOUT,
  TASKS_TIMEOUT, FINALLY_TIMEOUT) on the controller-manager container
  via withTimeoutField(), applied after FromContainerSpec so CRD fields
  take precedence over any same-named entry in manager.env.
- When a field is omitted, no env var is injected and the upstream
  integration-service default applies. Values supplied via manager.env
  still pass through when the typed fields are not set.
- Update the KonfluxIntegrationService sample YAML to show the new
  fields alongside the integrationControllerManager customization block.
- Regenerate CRDs and deepcopy to reflect the new API fields.
- Add TestBuildControllerManagerOverlay_PipelineTimeouts covering:
  injection, CRD-wins-over-manager.env precedence, empty fields
  (no injection), and manager.env pass-through.

Assisted-by: Cursor

### DIFF
--- a/operator/config/crd/bases/konflux.konflux-ci.dev_konfluxes.yaml
+++ b/operator/config/crd/bases/konflux.konflux-ci.dev_konfluxes.yaml
@@ -655,6 +655,15 @@ spec:
                   spec:
                     description: Spec configures the integration-service component.
                     properties:
+                      finallyTimeout:
+                        description: |-
+                          FinallyTimeout is the timeout for finally tasks (e.g. "2h", "30m").
+                          Maps to the FINALLY_TIMEOUT env var on the controller-manager container.
+                          Takes precedence over any FINALLY_TIMEOUT entry in integrationControllerManager.manager.env.
+                          When omitted, the upstream integration-service default applies.
+                        minLength: 2
+                        pattern: ^([0-9]+h)?([0-9]+m)?([0-9]+s)?$
+                        type: string
                       integrationControllerManager:
                         description: IntegrationControllerManager defines customizations
                           for the controller-manager deployment.
@@ -896,6 +905,24 @@ spec:
                             minimum: 1
                             type: integer
                         type: object
+                      pipelineTimeout:
+                        description: |-
+                          PipelineTimeout is the overall pipeline run timeout (e.g. "6h", "1h30m", "90m").
+                          Maps to the PIPELINE_TIMEOUT env var on the controller-manager container.
+                          Takes precedence over any PIPELINE_TIMEOUT entry in integrationControllerManager.manager.env.
+                          When omitted, the upstream integration-service default applies.
+                        minLength: 2
+                        pattern: ^([0-9]+h)?([0-9]+m)?([0-9]+s)?$
+                        type: string
+                      tasksTimeout:
+                        description: |-
+                          TasksTimeout is the timeout for tasks within a pipeline run (e.g. "4h", "90m").
+                          Maps to the TASKS_TIMEOUT env var on the controller-manager container.
+                          Takes precedence over any TASKS_TIMEOUT entry in integrationControllerManager.manager.env.
+                          When omitted, the upstream integration-service default applies.
+                        minLength: 2
+                        pattern: ^([0-9]+h)?([0-9]+m)?([0-9]+s)?$
+                        type: string
                     type: object
                 type: object
               internalRegistry:
--- a/operator/config/samples/konflux_v1alpha1_konfluxintegrationservice.yaml
+++ b/operator/config/samples/konflux_v1alpha1_konfluxintegrationservice.yaml
@@ -8,6 +8,12 @@ metadata:
     app.kubernetes.io/managed-by: kustomize
   name: konflux-integration-service
 spec:
+  # Pipeline run timeouts. Each field maps to the corresponding env var on the
+  # controller-manager container. Omit to use the upstream integration-service defaults.
+  pipelineTimeout: "6h"
+  tasksTimeout: "4h"
+  finallyTimeout: "2h"
+
   integrationControllerManager:
     replicas: 3
     manager:
@@ -18,14 +24,16 @@ spec:
         limits:
           cpu: 512m
           memory: 1Gi
-      env:
-        - name: EXAMPLE_CUSTOM_CONFIG
-          valueFrom:
-            configMapKeyRef:
-              name: integration-service-config
-              key: custom-config
-        - name: EXAMPLE_SECRET_VALUE
-          valueFrom:
-            secretKeyRef:
-              name: integration-service-secret
-              key: secret-key
+      # For env vars beyond the typed timeout fields above, use manager.env directly.
+      # Supports plain values and valueFrom references (ConfigMap, Secret).
+      # env:
+      #   - name: EXAMPLE_CUSTOM_CONFIG
+      #     valueFrom:
+      #       configMapKeyRef:
+      #         name: integration-service-config
+      #         key: custom-config
+      #   - name: EXAMPLE_SECRET_VALUE
+      #     valueFrom:
+      #       secretKeyRef:
+      #         name: integration-service-secret
+      #         key: secret-key
--- a/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
+++ b/operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
@@ -59,6 +59,11 @@ const (
 
 	// Container names
 	managerContainerName = "manager"
+
+	// Env var names for pipeline run timeout configuration.
+	envPipelineTimeout = "PIPELINE_TIMEOUT"
+	envTasksTimeout    = "TASKS_TIMEOUT"
+	envFinallyTimeout  = "FINALLY_TIMEOUT"
 )
 
 // IntegrationServiceCleanupGVKs defines which resource types should be cleaned up when they are
@@ -211,7 +216,7 @@ func applyIntegrationServiceDeploymentCustomizations(deployment *appsv1.Deployme
 		if spec.IntegrationControllerManager != nil {
 			deployment.Spec.Replicas = &spec.IntegrationControllerManager.Replicas
 		}
-		if err := buildControllerManagerOverlay(spec.IntegrationControllerManager, consoleURL).ApplyToDeployment(deployment); err != nil {
+		if err := buildControllerManagerOverlay(spec.IntegrationControllerManager, consoleURL, spec).ApplyToDeployment(deployment); err != nil {
 			return err
 		}
 	}
@@ -219,16 +224,16 @@ func applyIntegrationServiceDeploymentCustomizations(deployment *appsv1.Deployme
 }
 
 // buildControllerManagerOverlay builds the pod overlay for the controller-manager deployment.
-func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec, consoleURL string) *customization.PodOverlay {
-	// Build console URL template for pipeline run links in the UI.
-	// Format: https://<host>/ns/{{ .Namespace }}/pipelinerun/{{ .PipelineRunName }}
+// Typed timeout fields (PipelineTimeout, TasksTimeout, FinallyTimeout) are applied last and
+// take precedence over any env entry with the same name in integrationControllerManager.manager.env.
+// When not set in the CRD, the upstream integration-service defaults apply.
+func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploymentSpec, consoleURL string, integrationSpec konfluxv1alpha1.KonfluxIntegrationServiceSpec) *customization.PodOverlay {
 	consoleURLTemplate := ""
 	if consoleURL != "" {
 		consoleURLTemplate = fmt.Sprintf("%s/ns/{{ .Namespace }}/pipelinerun/{{ .PipelineRunName }}",
 			strings.TrimSuffix(consoleURL, "/"))
 	}
 
-	// Determine replicas and manager spec (default replicas to 1 if no spec)
 	replicas := int32(1)
 	var managerSpec *konfluxv1alpha1.ContainerSpec
 	if spec != nil {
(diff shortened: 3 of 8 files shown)


operator/config/samples/konflux-e2e.yaml
```diff
@@ -60,6 +60,13 @@ spec:
             limits:
               cpu: 30m
               memory: 128Mi
+      # snapshotGarbageCollector.resources: lower the GC container requests for resource-constrained
+      # clusters (e.g. local Kind).
+      snapshotGarbageCollector:
+        resources:
+          request:
+            cpu: 50m
+            memory: 64Mi
```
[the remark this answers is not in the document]
avi-biton: Fixed

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -220,33 +228,49 @@ func applyIntegrationServiceDeploymentCustomizations(deployment *appsv1.Deployme
...
 }
 
+func applySnapshotGCCustomizations(cj *batchv1.CronJob, spec *konfluxv1alpha1.ContainerSpec) {
+	containers := cj.Spec.JobTemplate.Spec.Template.Spec.Containers
+	for i := range containers {
+		if containers[i].Name != snapshotGCContainerName {
+			continue
+		}
+		customization.FromContainerSpec(spec)(&containers[i], customization.DeploymentContext{})
+	}
```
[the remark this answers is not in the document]
avi-biton: Fixed

operator/config/samples/konflux-e2e.yaml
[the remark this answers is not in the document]
avi-biton: /ask this issue was addressed in this PR, is it still an issue?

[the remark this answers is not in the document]
avi-biton: /ask look at commit https://github.com/konflux-ci/konflux-ci/commit/8ce3a97f34edc9b193b4d1d7ab5d3406c179031f. the fix is there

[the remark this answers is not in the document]
avi-biton: /ask In our block code we have: 
```
      snapshotGarbageCollector:
        resources:
          requests:
            cpu: 50m
            memory: 64Mi
```

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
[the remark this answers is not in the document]
avi-biton: /ask This issue was addressed, take a look at the latest commit and check it

operator/config/samples/konflux-e2e.yaml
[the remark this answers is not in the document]
avi-biton: /ask Is this this the commit sha you refer to? https://github.com/konflux-ci/konflux-ci/pull/6728/changes/8ce3a97f34edc9b193b4d1d7ab5d3406c179031f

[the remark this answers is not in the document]
avi-biton: /ask Thats the reference to the updated file: https://github.com/konflux-ci/konflux-ci/pull/6728/changes/8ce3a97f34edc9b193b4d1d7ab5d3406c179031f#diff-61bed93dbb4f835b356c996935e13d8e223076efe03169e3b26bf84c553664b9R67

avi-biton:
> [Nil Panic](https://github.com/konflux-ci/konflux-ci/pull/6728/files#diff-d0b6239196f48e8128517c03b341f09ade5066982ccb496f4a8b801668ac582eR258-R271)
applySnapshotGCCustomizations unconditionally calls customization.FromContainerSpec(spec) even when spec is nil. If FromContainerSpec does not explicitly handle a nil *ContainerSpec, this will panic at runtime (and the reconciler will fail) when users omit snapshotGarbageCollector. Consider short-circuiting early when spec == nil to guarantee safety regardless of helper implementation.
```
func applySnapshotGCCustomizations(cj *batchv1.CronJob, spec *konfluxv1alpha1.ContainerSpec) error {
	containers := cj.Spec.JobTemplate.Spec.Template.Spec.Containers
	for i := range containers {
		if containers[i].Name != snapshotGCContainerName {
			continue
		}
		overlay := &corev1.Container{Name: containers[i].Name}
		customization.FromContainerSpec(spec)(overlay, customization.DeploymentContext{})
		if err := customization.StrategicMerge(&containers[i], overlay); err != nil {
			return err
		}
	}
	return nil
}
```
Its not a valid concern. FromContainerSpec already nil-guards on line 51 of pkg/customization/container.go — a nil spec returns a no-op ContainerOption immediately. This is also covered by the existing test case applySnapshotGCCustomizations(cj, nil) in konfluxintegrationservice_customization_test.go:717.

avi-biton:
>[Name Mismatch](https://github.com/konflux-ci/konflux-ci/pull/6728/files#diff-d0b6239196f48e8128517c03b341f09ade5066982ccb496f4a8b801668ac582eR59-R64)
snapshotGCContainerName is set to test-gc, which looks like a test-only name. If the embedded integration manifests use a different container name for the snapshot GC CronJob, the customization will silently do nothing (and tests may pass if they also key off this constant). Validate that this constant matches the real manifest container name and is stable across upstream updates.
```
	controllerManagerDeploymentName = "integration-service-controller-manager"

	// Container names
	managerContainerName    = "manager"
	snapshotGCContainerName = "test-gc"
)
```

The constant is correct. The upstream CronJob container is literally named `test-gc` in the embedded manifests (pkg/manifests/integration/manifests.yaml:1628). Despite the name looking test-like, this is the real upstream container name and the constant accurately reflects it. If upstream renames it, the snapshotGCContainerName constant would need updating — the same way any other container name constant would.

avi-biton:
> [Silent No-op](https://github.com/konflux-ci/konflux-ci/pull/6728/files#diff-0ac299a413d74cfcd73cb007008be1cbff3aa04401657f5c8bde8f5933457cd1R218-R229)
ApplyContainerOpt silently returns when the named container is not found. This can mask manifest drift/renames (e.g., container name changes) and cause leader-election overrides to be skipped without any signal. Consider returning an error (or at least logging) when the container is missing, especially since this is used to enforce leader election behavior.
```
// ApplyContainerOpt applies a ContainerOption directly to a named container in a Deployment.
// It is a post-merge helper: use it after ApplyToDeployment to mutate the already-merged
// container in place, avoiding the atomic Args replacement that StrategicMerge performs on
// slice fields without a patchMergeKey.
func ApplyContainerOpt(deployment *appsv1.Deployment, containerName string, opt ContainerOption) {
	for i := range deployment.Spec.Template.Spec.Containers {
		if deployment.Spec.Template.Spec.Containers[i].Name == containerName {
			opt(&deployment.Spec.Template.Spec.Containers[i], DeploymentContext{})
			return
		}
	}
}
```
In this specific case the risk is mitigated: the container name manager is stable kubebuilder scaffolding (identical across all three controller managers), and the customization tests assert the leader-elect flag is applied after seeding a deployment with that exact name — so a rename would be caught immediately by CI. 
I added a warning log when the container is not found to make silent no-ops visible in production, without making the function return an error that would interrupt an otherwise healthy reconcile.
I also added tests to it

avi-biton:
@yftacherzog @gbenhaim I addressed the comments in https://github.com/konflux-ci/konflux-ci/pull/6728#issuecomment-4396685464
For some reason the bot is not updating the sha of its review

gbenhaim:
It looks like this pr contains 3 unrelated changes:
1. Add abstraction to leader election
2. Support configuring resource requirements for the snapshot gc.
3. Support the configuration / document additional environment variables the integration controller can accept.

Is it possible to split this pr into 3?

avi-biton:
> It looks like this pr contains 3 unrelated changes:
> 
> 1. Add abstraction to leader election
> 2. Support configuring resource requirements for the snapshot gc.
> 3. Support the configuration / document additional environment variables the integration controller can accept.
> 
> Is it possible to split this pr into 3?

1. The abstraction to leader election derived from infra-deployments integration component:
https://github.com/redhat-appstudio/infra-deployments/blob/main/components/integration/development/manager_resources_patch.yaml#L10
It was needed to bridge this gap and allow users to explicitly set the leader election value.
Since it was added as shared resource, it also applies to build and release.
2.  The SnapshotGC update was done because we used a patch to always reduce the resources for the crojob and we were not able to get the original values from upstream (which are used in infra-deployments)
3. I wanted to add the documentation as a part of each CRDs PR with its settings.

If you still think we need to split it, let me know.

gbenhaim:
> > It looks like this pr contains 3 unrelated changes:
> > 
> > 1. Add abstraction to leader election
> > 2. Support configuring resource requirements for the snapshot gc.
> > 3. Support the configuration / document additional environment variables the integration controller can accept.
> > 
> > Is it possible to split this pr into 3?
> 
> 1. The abstraction to leader election derived from infra-deployments integration component:
>    https://github.com/redhat-appstudio/infra-deployments/blob/main/components/integration/development/manager_resources_patch.yaml#L10
>    It was needed to bridge this gap and allow users to explicitly set the leader election value.
>    Since it was added as shared resource, it also applies to build and release.
> 2. The SnapshotGC update was done because we used a patch to always reduce the resources for the crojob and we were not able to get the original values from upstream (which are used in infra-deployments)
> 3. I wanted to add the documentation as a part of each CRDs PR with its settings.
> 
> If you still think we need to split it, let me know.

I understand why those changes are needed, but at the same time I think it would be safer to introduce them in separate prs.

avi-biton:
Splitting it to several PRs.
This is the first one

operator/config/samples/konflux_v1alpha1_konfluxintegrationservice.yaml
```diff
@@ -18,14 +18,36 @@ spec:
...
+      #   CONSOLE_NAME         — display name shown in the Konflux UI
+      #   CONSOLE_URL          — base URL of the Konflux UI (auto-set from KonfluxUI ingress; override only if needed)
+      #   CONSOLE_URL_TASKLOG  — URL template for pipeline run task log deep-links
+      #   PIPELINE_TIMEOUT     — overall pipeline run timeout, e.g. "6h"
+      #   TASKS_TIMEOUT        — timeout for tasks within a pipeline run, e.g. "4h"
+      #   FINALLY_TIMEOUT      — timeout for finally tasks, e.g. "2h"
+      #   PAC_SECRET           — name of the PipelinesAsCode secret in the namespace
+      #   GITHUBAPPLICATION_ID — secret key holding the GitHub App ID
+      #   GITHUBPRIVATE_KEY    — secret key holding the GitHub private key
+      #   INTEGRATION_NS       — namespace where the integration service runs
```
yftacherzog: Can we provide a link to where those env vars are documented or implemented in the integration-service repo? 

As far as our code goes, these env vars are arbitrary. I'm not sure we want to document them in our repo.

avi-biton: I will try to find a reference to these ENV Vars and add a link to it.
I hope such documentation exists

gbenhaim: Another question is which of those env vars we want to really expose to the user. For example, handling the condole URLs is done by the operator, users shouldn't  configure it.

avi-biton: @yftacherzog There is no documentation for these ENV vars in `integration-service`
The keys are taken from `infra-deployments` patches and from the integration-service source code

avi-biton: > Another question is which of those env vars we want to really expose to the user. For example, handling the condole URLs is done by the operator, users shouldn't configure it.


These are exposed and being used in `infra-deployments`

avi-biton:
@gbenhaim @yftacherzog 
I have updated this PR to promote PIPELINE_TIMEOUT, TASKS_TIMEOUT, and FINALLY_TIMEOUT from
generic env var examples to typed fields on KonfluxIntegrationServiceSpec.

avi-biton:
> **[Persistent review](https://github.com/konflux-ci/konflux-ci/pull/6728#issuecomment-4396854170)** updated to latest commit [51c308e](https://github.com/konflux-ci/konflux-ci/commit/51c308ea59e5ca1bda128ed8eae58f3a7a7db17e)

All issues were addressed

operator/config/samples/konflux_v1alpha1_konfluxintegrationservice.yaml
avi-biton: All the env vars and the examples were removed.
Instead we added 3 of these fields to the CRD

yftacherzog (approved):

operator/config/crd/bases/konflux.konflux-ci.dev_konfluxes.yaml
```diff
@@ -896,6 +905,24 @@ spec:
...
+                        type: string
+                      tasksTimeout:
+                        description: |-
+                          TasksTimeout is the timeout for tasks within a pipeline run (e.g. "4h", "90m").
+                          Maps to the TASKS_TIMEOUT env var on the controller-manager container.
+                          Takes precedence over any TASKS_TIMEOUT entry in integrationControllerManager.manager.env.
+                          When omitted, the upstream integration-service default applies.
+                        minLength: 2
+                        pattern: ^([0-9]+h)?([0-9]+m)?([0-9]+s)?$
+                        type: string
```
gbenhaim, line 925: /ask is there a type for time duration?

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -211,24 +216,33 @@ func applyIntegrationServiceDeploymentCustomizations(deployment *appsv1.Deployme
...
+		if err := buildControllerManagerOverlay(spec.IntegrationControllerManager, consoleURL, spec).ApplyToDeployment(deployment); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+// withTimeoutField returns a ContainerOption that sets the given env var to value.
+// It is a no-op when value is empty, leaving the upstream manifest default in place.
+func withTimeoutField(name, value string) customization.ContainerOption {
```
gbenhaim: I think that the name of the function can be changed (there is nothing specific about timeout in it) and it can be moved to the `customization` package (it would be useful for other controller as well).

gbenhaim (commented):
small comment, other than that lgtm

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Done

PR #6776, opened 2026-05-11
feat(KFLUXVNGD-966): add LeaderElect field to ControllerManagerDeploymentSpec

avi-biton:
Add a LeaderElect *bool field to ControllerManagerDeploymentSpec, giving operators explicit control over the --leader-elect flag on all three controller managers (integration, build, release).

Motivation:
infra-deployments sets leader election explicitly per environment. The operator had no CRD field to expose this — it was hardcoded to auto-enable for replicas > 1 with no way to override. The new LeaderElect *bool field closes this gap.

Changes:
- api: add LeaderElect *bool to ControllerManagerDeploymentSpec with precedence over the replica-count auto-logic.
- pkg/customization: replace WithLeaderElection() with:
  - ComputeLeaderElect() — derives the desired value from the spec (CR field wins; auto-enable for replicas > 1; nil lets the embedded manifest default win for single-replica deployments).
  - WithLeaderElectionControl() — filters any existing --leader-elect arg before appending the new value, avoiding duplicates.
  - ApplyContainerOpt() — post-merge helper that mutates a named container directly on the already-merged Deployment, bypassing StrategicMerge slice-replace entirely.
- controllers: apply leader election as a post-merge step via ApplyContainerOpt in all three service controllers.
- tests: add TestComputeLeaderElect, TestApplyContainerOpt, and per-service TestBuild*_LeaderElection cases verifying that base args (e.g. --health-probe-bind-address) are preserved after the override.
- Regenerate CRDs and deepcopy for the new field.

Assisted-by: Cursor


avi-biton:
Closing this PR as it no longer needed, we are already covering the options in our existing logic

PR #6781, opened 2026-05-11
feat(KFLUXVNGD-966): add typed snapshotGC retention and resource fields

avi-biton:
Add typed CRD fields for configuring the snapshot garbage collector (GC)
retention policy and container resources, removing the need for a
hard-coded kustomize patch.

Previously, the snapshot GC CronJob container had resource requests
overridden via a static kustomize patch (snapshot-gc-resource-patch.yaml)
that was always applied, making it impossible to use the upstream resource
defaults for production clusters. Retention counts could only be set
through unvalidated entries in snapshotGarbageCollector.env.

Changes:
- Add PRSnapshotsToKeep, NonPRSnapshotsToKeep, MinSnapshotsToKeepPerComponent
  string fields to KonfluxIntegrationServiceSpec, each validated with
  +kubebuilder:validation:Pattern=^[0-9]+$. These map directly to the
  PR_SNAPSHOTS_TO_KEEP, NON_PR_SNAPSHOTS_TO_KEEP, and
  MIN_SNAPSHOTS_TO_KEEP_PER_COMPONENT env vars on the GC container via
  customization.WithOptionalEnvOverride. CRD fields take precedence over
  same-named entries in snapshotGarbageCollector.env; when omitted, upstream
  defaults apply. The upstream binary reads all three via os.LookupEnv after
  flag.Parse() (confirmed at SHA 97f400b4); MIN_SNAPSHOTS_TO_KEEP_PER_COMPONENT
  is only configurable via env var (no corresponding flag exists upstream).
- Add SnapshotGarbageCollector *ContainerSpec field to allow overriding
  the GC container resource requests/limits and injecting arbitrary env
  vars (including valueFrom references).
- Remove snapshot-gc-resource-patch.yaml. Resources can now be set via
  snapshotGarbageCollector.resources. Users who need low resource limits
  (e.g. local Kind) can set them in the CR; production clusters can omit
  the field to use the upstream defaults.
- Add buildSnapshotGCOverlay() using BuildPodOverlay + WithContainerBuilder,
  following the established PodOverlay pattern used by all other reconcilers.
  applySnapshotGCCustomizations() retains an explicit container-not-found
  guard with an error (ApplyToPodTemplateSpec silently ignores unmatched
  containers, so the check must be explicit).
- Add package-level constants for the CronJob name, container name, and
  the three GC env var names.
- Add Owns(&batchv1.CronJob{}) to the integration controller so CronJob
  modifications and deletions trigger reconciliation.
- Add TestApplySnapshotGCCustomizations_ContainerNotFound covering the
  explicit error/no-op paths when the target container is missing.
- Add TestApplySnapshotGCCustomizations covering ContainerSpec injection,
  resource overrides, and requests-only preserving upstream limits.
- Add TestApplySnapshotGCCustomizations_TypedFields covering: env var
  injection, typed fields overriding snapshotGarbageCollector.env entries,
  empty fields are no-ops, "0" is not treated as empty, and env pass-through
  when typed fields are unset.
- Add envtest integration test asserting retention env vars land on the
  CronJob after a full reconciliation loop.
- Update sample YAMLs (konflux_v1alpha1_konfluxintegrationservice.yaml,
  konflux_v1alpha1_konflux.yaml, konflux-e2e.yaml) to demonstrate the
  new fields; Kind/e2e samples include lowered resource requests.
- Regenerate CRDs and deepcopy.

Notable behavior change:
Removing snapshot-gc-resource-patch.yaml changes the GC container resource
requests from cpu: 50m / memory: 64Mi to cpu: 1000m / memory: 2000Mi
(the upstream defaults, matching limits). Existing clusters upgrading the
operator will see a ~20x CPU and ~31x memory request increase on this
container. Production clusters generally have sufficient capacity and this
is the intended outcome. Local/resource-constrained clusters should add
snapshotGarbageCollector.resources to their CR — the sample YAMLs
(konflux_v1alpha1_konflux.yaml, konflux-e2e.yaml) already include these
lower values as a reference.

Assisted-by: Cursor


operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -247,6 +261,36 @@ func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploy
...
+	for i := range containers {
+		if containers[i].Name != snapshotGCContainerName {
+			continue
+		}
+		overlay := &corev1.Container{Name: containers[i].Name}
+		customization.FromContainerSpec(integrationSpec.SnapshotGarbageCollector)(overlay, customization.DeploymentContext{})
+		withGCField(envPRSnapshotsToKeep, integrationSpec.PRSnapshotsToKeep)(overlay, customization.DeploymentContext{})
+		withGCField(envNonPRSnapshotsToKeep, integrationSpec.NonPRSnapshotsToKeep)(overlay, customization.DeploymentContext{})
+		withGCField(envMinSnapshotsToKeepPerComponent, integrationSpec.MinSnapshotsToKeepPerComponent)(overlay, customization.DeploymentContext{})
+		if err := customization.StrategicMerge(&containers[i], overlay); err != nil {
```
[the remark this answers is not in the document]
avi-biton: Based on everything we investigated and tested, here's a concise response to the review comment:

The concern is valid on the surface, but the implementation is correct. The GC binary (cmd/snapshotgc/snapshotgc.go) explicitly reads all three env vars after flag.Parse(), with the comment:

"We want env vars args to take precedence over cli args"
```
flag.Parse()  // picks up --pr-snapshots-to-keep=70 from the manifest
// We want env vars args to take precedence over cli args
if value, ok := os.LookupEnv("PR_SNAPSHOTS_TO_KEEP"); ok {
    prSnapshotsToKeep, err = strconv.Atoi(value)  // overrides with CR value
}
```
([source, lines 546–568](https://github.com/konflux-ci/integration-service/blob/6e2751b24d1c25f14c657accf845f8351b1f54ef/cmd/snapshotgc/snapshotgc.go#L546-L568))

The env var injection is the intended override mechanism for this binary. We verified this end-to-end on a local Kind cluster: with prSnapshotsToKeep: 3 and nonPRSnapshotsToKeep: 7 set in the CR, the operator correctly injects PR_SNAPSHOTS_TO_KEEP=3 and NON_PR_SNAPSHOTS_TO_KEEP=7 into the CronJob container spec — while the command args remain at their upstream defaults (--pr-snapshots-to-keep=70, --non-pr-snapshots-to-keep=640). When the job runs, the binary reads the env vars and uses 3 and 7.

avi-biton:
>[Correctness] operator/pkg/manifests/integration/manifests.yaml:1633-1635 — Removing the kustomize patch (snapshot-gc-resource-patch.yaml) changes the GC container resource requests from cpu: 50m / memory: 64Mi to cpu: 1000m / memory: 2000Mi (matching limits). Existing clusters that relied on the old low requests will see a 20x CPU and 31x memory request increase on the next reconciliation, which may cause scheduling failures on resource-constrained clusters. The sample YAMLs demonstrate setting lower values via the CR, but existing deployments won't have those CR fields set.
Remediation: Document this as a breaking change in release notes. Consider whether the operator should default to the previous lower requests when snapshotGarbageCollector.resources is not set, or provide a migration guide.


The resource increase on production is intentional — the static patch was precisely what blocked production clusters from using the appropriate upstream defaults. The PR description explicitly calls this out.

For resource-constrained environments (local Kind, e2e), operator/config/samples/konflux_v1alpha1_konflux.yaml and konflux-e2e.yaml already include snapshotGarbageCollector.resources with cpu: 50m / memory: 64Mi. Users of those samples need to add this field to their existing CRs - the PR description documents this migration path: "Users who need low resource limits (e.g. local Kind) can set them in the CR; production clusters can omit the field to use the upstream defaults."

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -247,6 +265,55 @@ func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploy
 	)
 }
 
+// withGCField returns a ContainerOption that sets the given env var to the string
+// representation of value. It is a no-op when value is nil, leaving the upstream default.
+// The GC binary reads these env vars after flag.Parse(), so they take precedence over
+// the hard-coded command args in the upstream manifest.
+func withGCField(name string, value *int32) customization.ContainerOption {
```
gbenhaim: This can be replaced with the generic function you've added in https://github.com/konflux-ci/konflux-ci/pull/6728/changes#diff-c310d6b463e872cd5f6e989734865afe6256c50f64241b50b317585515fe7315R121

also, I think that the value should be string, since env vars values are strings.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -57,8 +58,18 @@ const (
 	// Deployment names
 	controllerManagerDeploymentName = "integration-service-controller-manager"
 
+	// CronJob names
+	snapshotGCCronJobName = "integration-service-snapshot-garbage-collector"
+
 	// Container names
-	managerContainerName = "manager"
+	managerContainerName    = "manager"
+	snapshotGCContainerName = "test-gc"
```
gbenhaim, line 66: **Concern**: `"test-gc"` looks like a development/testing name in the upstream manifests. It matches what's actually in the embedded manifest today, so it's correct — but it's fragile. If the upstream renames this to something more conventional (e.g. `snapshot-gc`), this constant would silently stop matching. The `hasCustomizations` guard at the bottom of `applySnapshotGCCustomizations` handles the case where customizations are set but the container is missing, which is good.

Not blocking, just noting.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -247,6 +265,55 @@ func buildControllerManagerOverlay(spec *konfluxv1alpha1.ControllerManagerDeploy
...
+// applySnapshotGCCustomizations applies user-defined customizations to the snapshot GC CronJob.
+// Typed retention fields are applied last and take precedence over any same-named entry in
+// snapshotGarbageCollector.env. The GC binary reads env vars after flag.Parse(), so injected
+// env vars override the command-arg defaults in the upstream manifest. When not set, the
+// upstream defaults apply.
+//
+// An error is returned if the user has configured any GC fields but the expected container
+// (snapshotGCContainerName) is not found in the CronJob spec — this prevents misconfigurations
+// from silently passing when the upstream container name changes.
+func applySnapshotGCCustomizations(cj *batchv1.CronJob, integrationSpec konfluxv1alpha1.KonfluxIntegrationServiceSpec) error {
```
gbenhaim, line 300: **Suggestion — follow the established `PodOverlay` pattern**: Every other reconciler in this codebase (build-service, release-service, UI, image-controller, namespace-lister, and even the integration controller-manager overlay in this same file) uses `customization.BuildPodOverlay` + `WithContainerBuilder` to build an overlay, then applies it via `ApplyToDeployment`/`ApplyToPodTemplateSpec`. The container iteration and strategic merge are handled internally by `mergeContainerList` — reconcilers never touch the container list directly.

This function manually iterates over containers, manually invokes `ContainerOption` functions, and calls `StrategicMerge` directly, bypassing the `PodOverlay` abstraction.

The idiomatic approach would be:

```go
func buildSnapshotGCOverlay(spec konfluxv1alpha1.KonfluxIntegrationServiceSpec) *customization.PodOverlay {
    return customization.BuildPodOverlay(
        customization.DeploymentContext{},
        customization.WithContainerBuilder(
            snapshotGCContainerName,
            customization.FromContainerSpec(spec.SnapshotGarbageCollector),
            withGCField(envPRSnapshotsToKeep, spec.PRSnapshotsToKeep),
            withGCField(envNonPRSnapshotsToKeep, spec.NonPRSnapshotsToKeep),
            withGCField(envMinSnapshotsToKeepPerComponent, spec.MinSnapshotsToKeepPerComponent),
        ),
    )
}
```

Then at the call site, add an `ApplyToCronJob` convenience method to `PodOverlay` (one-liner following the `ApplyToDeployment` pattern), or call `ApplyToPodTemplateSpec(&cj.Spec.JobTemplate.Spec.Template)` directly.

The "container not found" validation (`hasCustomizations` guard) is valuable and not provided by `mergeContainerList` (which silently ignores unmatched overlays). You could keep it as a separate pre-check before applying the overlay.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -57,8 +58,18 @@ const (
...
+	// CronJob names
+	snapshotGCCronJobName = "integration-service-snapshot-garbage-collector"
+
 	// Container names
-	managerContainerName = "manager"
+	managerContainerName    = "manager"
+	snapshotGCContainerName = "test-gc"
+
+	// Env var names for snapshot garbage collector retention configuration.
+	// The binary reads these after flag.Parse(), so they override the command-line args.
```
gbenhaim, line 74: **Concern**: The comment states the GC binary reads these env vars after `flag.Parse()`, so they override the command-line args. However, looking at the upstream CronJob command args in the embedded manifest, only `--pr-snapshots-to-keep` and `--non-pr-snapshots-to-keep` appear — there's no `--min-snapshots-to-keep-per-component` flag at all.

Could you confirm the upstream integration-service binary at the pinned SHA (`ef2610e`) actually reads `PR_SNAPSHOTS_TO_KEEP`, `NON_PR_SNAPSHOTS_TO_KEEP`, and `MIN_SNAPSHOTS_TO_KEEP_PER_COMPONENT` as environment variables? If not, the typed retention fields would silently have no effect.

operator/internal/controller/integrationservice/konfluxintegrationservice_customization_test.go
```diff
@@ -36,6 +37,24 @@ const (
 )
 
 // getIntegrationServiceDeployment returns a deep copy of the IntegrationService controller-manager deployment from the manifests.
```
gbenhaim: **Nit**: This comment is orphaned — it's the doc comment for `getIntegrationServiceDeployment` (line 58), but it ended up above `getIntegrationSnapshotGCCronJob` when the new function was inserted. As a result, `getIntegrationServiceDeployment` has no doc comment and this function has two stacked comments.

Move `// getIntegrationServiceDeployment returns a deep copy...` down to line 58.

operator/pkg/manifests/integration/manifests.yaml
```diff
@@ -1631,8 +1631,8 @@ spec:
                 cpu: 1000m
                 memory: 2000Mi
               requests:
-                cpu: 50m
-                memory: 64Mi
+                cpu: 1000m
```
gbenhaim, line 1634: **Note for release**: With the kustomize patch removed, GC container requests jump from 50m CPU / 64Mi memory to 1000m CPU / 2000Mi memory (matching limits). This is the upstream default, which is the right call for production, but existing clusters upgrading the operator will see a ~20x CPU and ~31x memory request increase on this container.

The sample CRs (`konflux-e2e.yaml`, `konflux_v1alpha1_konflux.yaml`) include lowered values for Kind, so local dev is covered. Worth calling out in release notes for production users who may want to set `snapshotGarbageCollector.resources` explicitly.

gbenhaim (commented):

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Acknowledged. The `hasCustomizations` guard already converts the "silent no-op" into an explicit error for the cases that matter: if a user has set any GC fields on the CR and the container can't be found, reconciliation fails loudly rather than silently dropping the customization.

For the no-customizations path (the common case where the CR has no GC fields set), a container rename upstream would be a no-op — which is fine, since there's nothing to apply.

In practice, the risk is lower than it looks: the container name comes from the embedded `manifests.yaml` that we pin to an exact upstream SHA. If upstream renames the container, the embedded manifest and our constant would diverge simultaneously, which would be caught immediately by our own tests (the envtest integration test asserts the container is found and the env vars land on it). The rename wouldn't silently slip through to a live cluster.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Done. 
- The values are strings now 
- `withGCField ` removed - using `customization.WithOptionalEnvOverride` instead

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Fixed

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Confirmed at the exact pinned SHA (97f400b4). The binary reads all three env vars explicitly in cmd/snapshotgc/snapshotgc.go:

https://github.com/konflux-ci/integration-service/blob/main/cmd/snapshotgc/snapshotgc.go#L552-L573
```
if value, ok := os.LookupEnv("PR_SNAPSHOTS_TO_KEEP"); ok {
    prSnapshotsToKeep, err = strconv.Atoi(value)
    ...
}
if value, ok := os.LookupEnv("NON_PR_SNAPSHOTS_TO_KEEP"); ok {
    nonPrSnapshotsToKeep, err = strconv.Atoi(value)
    ...
}
if value, ok := os.LookupEnv("MIN_SNAPSHOTS_TO_KEEP_PER_COMPONENT"); ok {
    minSnapShotsToKeepPerComponent, err = strconv.Atoi(value)
    ...
}
```
The observation is correct that `--min-snapshots-to-keep-per-component` doesn't appear in the CronJob command args - because MIN_SNAPSHOTS_TO_KEEP_PER_COMPONENT is only configurable via env var in the upstream binary, not via a flag. All three env vars are read after flag.Parse(), so they take precedence over the hardcoded command args for the other two fields. Injecting them via the container env is the correct and only mechanism for all three.

operator/pkg/manifests/integration/manifests.yaml
avi-biton: @gbenhaim 
I added a note to the commit message and to the PR description.
We also have comments in:
- https://github.com/konflux-ci/konflux-ci/pull/6781/changes#diff-681887fe5e905c32a5f46a07dde3ba0815a2c29e2234e13755eb12b3ea6547d0R68
- https://github.com/konflux-ci/konflux-ci/pull/6781/changes#diff-a7c5b76e7cafec5889aa1ec3e9c573a52fed0f8ee53cfdad6e0796ab9e4309e2R49
- https://github.com/konflux-ci/konflux-ci/pull/6781/changes#diff-61bed93dbb4f835b356c996935e13d8e223076efe03169e3b26bf84c553664b9R63

Do you think we should add it in another place?

gbenhaim (approved):
All review comments have been addressed. The code now follows the established PodOverlay pattern, uses WithOptionalEnvOverride with string-typed fields, and the orphaned comment is fixed. Upstream env var support confirmed. LGTM.

PR #6803, opened 2026-05-12
feat: cluster-hosted script provides onboarding manifests

yftacherzog:
Introduce a script that will be hosted in the cluster and will allow
users to generate the manifests required for manifest-based onboarding.

Also enhancing the defaulttenant controller to attach the registry
secret to the konflux-integration-runner service account. This allows
the Conforma ITS to run successfully.
    
Together, this allows (and documents) onboarding by:
1. Forking & cloning testrepo.
2. Fetching the script from the cluster and running it.
3. Copying the pipelines dir to .tekton and creating a PR.

This will get the user up until successful build pipeline and EC test.

Next, we'll need to verify what happens in release and see if this
requires any tweaks.

Assisted-by: Cursor


operator/upstream-kustomizations/cli/setup-component.sh
```diff
@@ -0,0 +1,344 @@
+#!/bin/bash -e
+
+# Script to set up application/component onboarding resources for a Konflux component.
+# Creates Application, Component, Pipelines-as-Code Repository, a Conforma
+# IntegrationTestScenario (same shape as the Konflux UI, build-definitions @ main), and
+# an optional additional IntegrationTestScenario from custom git (-i/-j/-k).
+# Note: this script needs to be compatible with Bash 3.x to support both macOS and Linux.
+
+set -o pipefail
+set -eu
```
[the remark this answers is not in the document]
yftacherzog: It follows the same flags of the other cluster-hosted scripts that we have at the moment.

operator/upstream-kustomizations/cli/setup-component.sh
```diff
@@ -0,0 +1,344 @@
...
+    echo "Waiting for ServiceAccount '${BUILD_PIPELINE_SA}' to be created by Build Service..."
+    wait_for_serviceaccount "${BUILD_PIPELINE_SA}"
+
+    # imagePullSecrets: kubelet injects these for pulling *container* images in the pod spec.
+    # secrets: Konflux/Tekton build steps (e.g. buildah push) still expect dockerconfigjson secrets
+    # listed here so creds are materialized under /tekton/creds for in-pod registry auth — imagePullSecrets alone is not enough.
+    echo "Attaching secret '${REGISTRY_SECRET_NAME}' to ServiceAccount '${BUILD_PIPELINE_SA}' (secrets + imagePullSecrets)..."
+    kubectl patch serviceaccount "${BUILD_PIPELINE_SA}" -n "${TENANT_NS}" \
+      --type merge \
+      -p "{\"secrets\":[{\"name\":\"${REGISTRY_SECRET_NAME}\"}],\"imagePullSecrets\":[{\"name\":\"${REGISTRY_SECRET_NAME}\"}]}" >/dev/null
```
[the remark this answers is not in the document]
yftacherzog: fixed

operator/upstream-kustomizations/cli/setup-component.sh
```diff
@@ -0,0 +1,351 @@
...
+  resolverRef:
+    resolver: git
+    resourceKind: pipeline
+    params:
+      - name: url
+        value: https://github.com/konflux-ci/build-definitions
+      - name: revision
+        value: main
+      - name: pathInRepo
+        value: pipelines/enterprise-contract.yaml
```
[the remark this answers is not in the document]
yftacherzog: This is the same as what onboarding via the UI does.
The pipline is very stable.
The actual policy is taken from the cluster and pinned by digest.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -215,9 +222,140 @@ func (r *KonfluxDefaultTenantReconciler) syncInternalRegistryCredentialsSecret(
...
+		if apierrors.IsNotFound(err) {
+			return fmt.Errorf("serviceaccount %s/%s not found yet", DefaultTenantNamespace, IntegrationRunnerServiceAccountName)
+		}
+		return fmt.Errorf("get serviceaccount %s/%s: %w", DefaultTenantNamespace, IntegrationRunnerServiceAccountName, err)
+	}
+	original := sa.DeepCopy()
+	if !mergeRegistryCredRefsIntoServiceAccount(sa) {
+		return nil
+	}
+	if err := r.Patch(ctx, sa, client.MergeFrom(original)); err != nil {
```
gbenhaim: can we use server side apply?

gbenhaim (commented):

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -215,9 +222,140 @@ func (r *KonfluxDefaultTenantReconciler) syncInternalRegistryCredentialsSecret(
...
+
 	return ctrl.Result{}, nil
 }
 
+// ensureIntegrationRunnerRegistryCreds patches the konflux-integration-runner ServiceAccount in
+// default-tenant so imagePullSecrets and secrets list RegistryCredentialsSecretName. That lets
+// integration and Conforma PipelineRuns authenticate to the internal registry after the copied
+// dockerconfig secret exists. It is a no-op when the references are already present. Returns an
+// error if the ServiceAccount is missing (bootstrap ordering) or the patch fails.
+func (r *KonfluxDefaultTenantReconciler) ensureIntegrationRunnerRegistryCreds(ctx context.Context) error {
```
gbenhaim: **Suggestion**: Consider handling the integration runner SA registry creds inside `applyManifests` rather than as a separate post-apply patch.

Instead of the current two-step approach (SSA-apply the SA, then patch it to add cred refs), you could check whether the internal registry is enabled before applying manifests and, if so, inject the cred refs into the SA object in-memory before the SSA apply. For example, `applyManifests` could accept the secret name to attach to the SA (empty string when no internal registry), and mutate the SA in the object list before applying.

Benefits:
- The SA reaches its desired state atomically in a single SSA operation — no create-then-patch two-step
- Eliminates the hard error when the SA doesn't exist yet at patch time (`"serviceaccount ... not found yet"`)
- Removal is handled naturally: when the registry CR is absent, the SA is applied without the cred refs and SSA reconciles the diff
- Simplifies the code — `ensureIntegrationRunnerRegistryCreds`, `ensureIntegrationRunnerRegistryCredsRemoved`, and their error-handling branches can be dropped

This would keep things simple for now; if more customizations are needed later, you can introduce a more general abstraction.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -80,6 +83,7 @@ type KonfluxDefaultTenantReconciler struct {
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxdefaulttenants/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=konflux.konflux-ci.dev,resources=konfluxdefaulttenants/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=create;get;list;patch;update;watch
```
gbenhaim: **Suggestion (least-privilege)**: The `update` verb on serviceaccounts appears unnecessary. The controller only uses `Patch` (both SSA in `applyManifests` and `MergeFrom` in the SA patching), never `Update`. Consider removing `update` from this annotation and regenerating `role.yaml`.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -215,9 +222,140 @@ func (r *KonfluxDefaultTenantReconciler) syncInternalRegistryCredentialsSecret(
...
+	}
+	if err := r.Patch(ctx, sa, client.MergeFrom(original)); err != nil {
+		return fmt.Errorf("patch serviceaccount %s/%s to remove registry credentials: %w", DefaultTenantNamespace, IntegrationRunnerServiceAccountName, err)
+	}
+	return nil
+}
+
+// mergeRegistryCredRefsIntoServiceAccount mutates sa by appending RegistryCredentialsSecretName to
+// imagePullSecrets and secrets when not already listed. Returns true if either slice was modified.
+func mergeRegistryCredRefsIntoServiceAccount(sa *corev1.ServiceAccount) bool {
```
gbenhaim: **Suggestion**: These helper functions (`mergeRegistryCredRefsIntoServiceAccount`, `stripRegistryCredRefsFromServiceAccount`, `localObjectRefContainsName`, `objectRefContainsName`, `filterLocalObjectRefs`, `filterObjectRefs`) are all general-purpose SA utilities that don't depend on controller state. Consider moving them to a `pkg/serviceaccount/` package with generic signatures (accepting the secret name as a parameter rather than using the package-level constant). They'd be reusable by other controllers that need to attach secrets to SAs.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
```diff
@@ -215,9 +222,140 @@ func (r *KonfluxDefaultTenantReconciler) syncInternalRegistryCredentialsSecret(
...
+// default-tenant so imagePullSecrets and secrets list RegistryCredentialsSecretName. That lets
+// integration and Conforma PipelineRuns authenticate to the internal registry after the copied
+// dockerconfig secret exists. It is a no-op when the references are already present. Returns an
+// error if the ServiceAccount is missing (bootstrap ordering) or the patch fails.
+func (r *KonfluxDefaultTenantReconciler) ensureIntegrationRunnerRegistryCreds(ctx context.Context) error {
+	sa := &corev1.ServiceAccount{}
+	key := types.NamespacedName{Namespace: DefaultTenantNamespace, Name: IntegrationRunnerServiceAccountName}
+	if err := r.Get(ctx, key, sa); err != nil {
+		if apierrors.IsNotFound(err) {
+			return fmt.Errorf("serviceaccount %s/%s not found yet", DefaultTenantNamespace, IntegrationRunnerServiceAccountName)
```
gbenhaim: **Suggestion**: When the SA is NotFound, this returns a hard error which causes the reconciliation to be marked as failed. Since `applyManifests` creates the SA synchronously before this runs, it should normally be present — but returning a `RequeueAfter` (similar to how missing source credentials are handled on line 203) would be more resilient to transient timing issues and produce a cleaner reconciliation log.

gbenhaim (commented):
Follow-up comments on specific lines.

To clarify the earlier suggestion: the SSA approach is preferred for its simplicity, but if other controllers also modify this SA's `secrets`/`imagePullSecrets` (making SSA field-ownership tricky), the current Patch approach is fine. In that case, these additional suggestions apply.

operator/pkg/manifests/cli/manifests.yaml
```diff
@@ -79,6 +80,365 @@ metadata:
   namespace: konflux-cli
 ---
 apiVersion: v1
+data:
+  setup-component.sh: |
```
yftacherzog, line 84: Not sure I understand the question. What would be a better name?

operator/upstream-kustomizations/cli/setup-component.sh
```diff
@@ -0,0 +1,351 @@
...
+
+  # Configure a component on a custom branch with in-cluster registry mode
+  $(basename "$0") -t user-ns2 -a testrepo -c testrepo \\
+    -g https://github.com/my-org/testrepo -r my-branch -M in-cluster
+
+  # Include integration test scenario
+  $(basename "$0") -a my-app -c my-component \\
+    -i https://github.com/my-org/testrepo -j main -k integration-tests/my-test.yaml
+EOF
+    exit 1
```
gbenhaim: **Nit**: `usage()` exits with code 1, which is appropriate when called from the unknown-option error path (line 151), but not when the user explicitly asks for help via `-h`/`--help`. Convention is to exit 0 for successful help display. Consider `exit "${1:-0}"` here and calling `usage 1` from the error path.

gbenhaim (commented):

avi-biton:
Do we have a path to a situation where the application already exists and we want to add a component to it?
It might not be in scope

gbenhaim:
https://github.com/konflux-ci/konflux-ci/pull/6803#discussion_r3239794497
@yftacherzog no issue with the name of the script. I didn't read it well.

yftacherzog:
> Do we have a path to a situation where the application already exists and we want to add a component to it? It might not be in scope

We don't have such scenario documented. I didn't check what happens if the application is already in place.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: Moved to use SSA now. On removal it's still patching, as there's some concern about the ability of SSA to delete only the things that should be deleted.

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: Done

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: done

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: done

operator/internal/controller/defaulttenant/konfluxdefaulttenant_controller.go
yftacherzog: This changed after attending to the previous comments, but on the removal path, there was still something similar. Changed it there.

operator/upstream-kustomizations/cli/setup-component.sh
yftacherzog: Done. Opened an issue to do the same in the other scripts we have there.
https://github.com/konflux-ci/konflux-ci/issues/6868

operator/pkg/serviceaccount/refs_test.go
```diff
@@ -0,0 +1,147 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestMergeSecretRefsIntoServiceAccount(t *testing.T) {
+	const reg = "regcred-internal-registry"
+	otherPull := corev1.LocalObjectReference{Name: "other-pull"}
+	otherSec := corev1.ObjectReference{Name: "other-docker"}
+
+	t.Run("adds cred to both lists when SA lists are empty", func(t *testing.T) {
+		g := gomega.NewWithT(t)
```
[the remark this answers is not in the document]
yftacherzog: Sibling test dirs use t.Run + gomega assertions, which is the same as we do here.

PR #6839, opened 2026-05-13
feat(ui): replace Nginx reverse proxy with Caddy

gbenhaim:
## Summary

Replace the Nginx-based reverse proxy with Caddy for the Konflux UI deployment.

- **Caddy configuration**: New Caddyfile with `forward_auth` to oauth2-proxy, `reverse_proxy` to Kubernetes API, namespace-lister, and extensible backend services (Tekton Results, KubeArchive) via import snippets
- **Dynamic configuration**: Init container (`generate-proxy-config.sh`) generates Caddy snippets at startup; `config-refresh` sidecar reloads Caddy via Admin API when SA tokens or CA certificates change
- **Group impersonation**: `map` directive splits comma-separated groups from `X-Auth-Request-Groups` into individual `Impersonate-Group` headers (supports up to 10 groups); always adds `system:authenticated`
- **Backend token isolation**: Dedicated projected SA token with `audience: "konflux-backend"` for backend services using TokenReview (Tekton Results temporarily uses the general token due to [upstream limitation](https://github.com/tektoncd/results/issues/1331))
- **Platform-aware TLS**: Mounts OpenShift service CA (`openshift-service-ca.crt`) when available; falls back to `tls_insecure_skip_verify` for non-OpenShift
- **Controller**: Uses `pkg/customization.WithRunAsUser` for conditional `runAsUser` injection on non-OpenShift clusters
- **Liveness probing**: Heartbeat-based probe verifies the config-refresh sidecar loop is active
- **Documentation**: Architecture README with Mermaid diagrams, user docs for OIDC group support and namespace group access

**Depends on:** #6836 (WithRunAsUser), #6838 (API rename)

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] Unit tests pass (`go test ./internal/controller/ui/...`)
- [x] Embedded manifests regenerated and in sync
- [ ] CI passes
- [ ] Manual validation on local Kind cluster


operator/upstream-kustomizations/ui/core/proxy/generate-proxy-config.sh
```diff
@@ -0,0 +1,90 @@
...
+    echo "${ocp_host}"
+    return 0
+  fi
+
+  return 1
+}
+
+# Process tekton-results template
+if [ -f "${TEMPLATES_DIR}/tekton-results.caddy" ]; then
+  hostname=$(resolve_tekton_results) && {
```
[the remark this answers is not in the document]
gbenhaim: Fixed — replaced A && B || C with a proper if/else.

— Gal's Cursor

operator/upstream-kustomizations/ui/core/proxy/generate-proxy-config.sh
```diff
@@ -0,0 +1,90 @@
...
+TOKEN_FILEPATH=/var/run/secrets/konflux-ci.dev/serviceaccount/token
+BACKEND_TOKEN_FILEPATH=/var/run/secrets/konflux-ci.dev/backend/token
+SERVICE_CA_PATH=/mnt/service-ca/service-ca.crt
+
+for cmd in sed nslookup cat; do
+  command -v "${cmd}" >/dev/null 2>&1 || { echo "required command not found: ${cmd}"; exit 1; }
+done
+
+log() { echo "$(date -Iseconds) generate-proxy-config: $*"; }
+
```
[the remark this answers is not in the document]
gbenhaim: Dismissed — these scripts run exclusively inside a UBI container in a pod, not on the user's host. GNU coreutils (`date -Iseconds`) are always available. The `$*` in `log()` is intentional for simple log messages. The `/bin/sh` scripts use `set -eu` which is POSIX; `pipefail` is not available in `/bin/sh`.

— Gal's Cursor

operator/internal/controller/ui/konfluxui_customization_test.go
```diff
@@ -980,3 +982,79 @@ func TestProxyDeploymentTokenHotReload(t *testing.T) {
...
+		g.Expect(err).NotTo(gomega.HaveOccurred())
+
+		rpContainer := testutil.FindContainer(deployment.Spec.Template.Spec.Containers, reverseProxyContainerName)
+		g.Expect(rpContainer).NotTo(gomega.BeNil())
+		if rpContainer.SecurityContext != nil {
+			g.Expect(rpContainer.SecurityContext.RunAsUser).To(gomega.BeNil(),
+				"OpenShift SCCs manage runAsUser; the operator must not set it")
+		}
+	})
+}
```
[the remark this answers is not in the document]
gbenhaim: Dismissed — the existing `konfluxui_customization_test.go` file consistently uses standard `testing.T`/`t.Run`. Mixing Ginkgo into a file that already follows a different convention would hurt readability. The `container_test.go` file in PR 1 has the same pattern.

— Gal's Cursor

operator/pkg/manifests/ui/manifests.yaml
```diff
@@ -580,20 +731,40 @@ spec:
...
+          readOnly: true
+          subPath: impersonate-groups.conf
+        - mountPath: /etc/caddy/impersonate-headers.conf
+          name: proxy-caddyfile
+          readOnly: true
+          subPath: impersonate-headers.conf
+        - mountPath: /etc/caddy/ns-lister-headers.conf
+          name: proxy-caddyfile
+          readOnly: true
+          subPath: ns-lister-headers.conf
```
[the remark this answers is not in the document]
gbenhaim: Fixed — added `trusted-ca`, `service-ca`, and `serving-cert` volume mounts to the `config-refresh` container. Also updated `CA_WATCH_PATHS` to include `/mnt/serving-cert/tls.crt` so the proxy's own TLS certificate rotation triggers a Caddy reload.

— Gal's Cursor

.github/workflows/caddy-fmt.yaml
```diff
@@ -0,0 +1,36 @@
+name: Verify Caddyfile formatting
+
+on:
+  pull_request:
+    paths:
+      - operator/upstream-kustomizations/ui/core/proxy/Caddyfile
```
yftacherzog, line 6: Maybe add the path to the workflow (this file) as well.

yftacherzog:
> **Fixed (1):**
> 
> 1. **Empty Impersonate-Group headers from unused placeholders need explicit validation** (`test/go-tests/proxy_test.go`): Added an explicit e2e test 'Test empty Impersonate-Group header tolerance' that validates user2 (who has no IdP groups via Dex local password connector, so all 10 placeholders resolve to empty strings) can successfully access Kube API resources through the proxy. This makes the implicit validation explicit.

@gbenhaim did the bot actually made a change to the PR? `/` + `fs-fix-stop` should add a label that should get it not to apply fixes. I'm not sure if this is a bug or not for it to trying to fix. If it is, might already be fixed.

Edit: This seems to suggest that the behavior was changed: https://docs.google.com/document/d/14bSk5khKlt4YO4W_JWsTs4yR1b8-ZuT3twzolxCAZJ8/edit?tab=t.ituvyi7744ql

yftacherzog:
> * **[documentation currency]** `operator/docs/content/docs/overview.md:85` — The component table describes the UI as "The Konflux web interface, nginx reverse proxy, and Dex identity provider" but this PR replaces nginx with Caddy. This reference is now stale and misleading.
>   Remediation: Update the description to reference "Caddy reverse proxy" instead of "nginx reverse proxy".

Is this addressed?

yftacherzog (approved):

.github/workflows/caddy-fmt.yaml
gbenhaim: Good call — added `.github/workflows/caddy-fmt.yaml` to the trigger paths so the check also runs when the workflow itself is modified.

— Gal's Cursor

gbenhaim:
> did the bot actually made a change to the PR?

No, the bot did not make any code changes to the PR. The "Fixed" items in the bot's summary were all done manually by us — the bot just listed what it flagged and we addressed each one (either fixed or dismissed with rationale). The `/fs-fix-stop` label should prevent it from auto-applying fixes.

> Is this addressed? (stale "nginx" reference in overview.md)

Yes — just pushed a fix updating `overview.md` to say "Caddy reverse proxy" instead of "nginx reverse proxy". Also checked the rest of the docs; the only other "nginx" references are in `custom-url.md` where they refer to nginx as an ingress controller, which is correct.

— Gal's Cursor

.github/workflows/caddy-fmt.yaml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Extract Caddy image reference
+        id: image
+        run: |
+          digest=$(grep -A1 'name: registry.access.redhat.com/hi/caddy' \
+            operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml \
+            | grep 'digest:' | awk '{print $2}')
+          echo "ref=registry.access.redhat.com/hi/caddy@${digest}" >> "$GITHUB_OUTPUT"
```
filariow, line 26: I'd suggest to use `yq` to have a more reliable extraction:

```suggestion
          ref=$(yq '.images[] | select(.name=="registry.access.redhat.com/hi/caddy") | .name + "/" + .digest' \
        operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml)
          echo "ref=${ref}" >> "$GITHUB_OUTPUT"
```

operator/upstream-kustomizations/ui/core/proxy/Caddyfile
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+	}
+
+	# Static SPA (catch-all, must be last)
+	handle {
+		root * /opt/app-root/src/static-content
+		try_files {path} /index.html
+		file_server
+	}
```
filariow, line 167: this is an ingress for an API server. should we just return 404 here?

filariow: we don't have a SPA (Single Page Application I'd guess) here, right?

gbenhaim: we server also the konflux ui pages and its a single page application.

.github/workflows/caddy-fmt.yaml
gbenhaim: Good idea — switched to `yq` for reliable YAML parsing. I used `@` as the separator instead of `/` since Docker image references use the `name@digest` format.

— Gal's Cursor

gbenhaim: apparently the bot didn't commit the file with the fix. it was added here https://github.com/konflux-ci/konflux-ci/pull/6973/changes

PR #6866, opened 2026-05-14
fix(KFLUXVNGD-992): preserve base container args during strategic merge

### LINKED ISSUE #6860: Strategic merge replaces container args instead of appending
`mergeContainerList` in `pkg/customization/pod.go` uses `StrategicMerge` to apply container overlays. 
Kubernetes Container.Args is tagged +listType=atomic, which means strategic merge replaces the entire args list rather than merging element-by-element.

This causes any overlay that sets Args (via WithArgs or WithLeaderElection) to silently wipe out the base container's args from the embedded manifest.

## Impact:

Any controller running with replicas > 1 would lose all base args (--metrics-bind-address, --health-probe-bind-address, -webhook-config-path) when WithLeaderElection appends --leader-elect=true. The service would fail to bind its health and metrics endpoints.

## Proposed Fix 
Save overlay args before merge, clear them, then append after merge.

avi-biton:
Container.Args is tagged +listType=atomic in the Kubernetes API, so strategic merge replaces the entire list. This caused overlay args (e.g. --leader-elect=true from WithLeaderElection) to silently wipe all base args from the embedded manifest.

Save overlay args before merge and append them after, so base args are preserved and overlay args are added.

fix https://github.com/konflux-ci/konflux-ci/issues/6860

Assisted-by: Cursor

### DIFF
--- a/operator/pkg/customization/pod.go
+++ b/operator/pkg/customization/pod.go
@@ -232,16 +232,25 @@ func (p *PodOverlay) ApplyToDaemonSet(daemonSet *appsv1.DaemonSet) error {
 }
 
 // mergeContainerList merges container overlay configurations into a list of containers.
+// Container.Args is tagged +listType=atomic in the Kubernetes API, so strategic merge
+// replaces the entire args list. We handle args separately: save them before merge,
+// then append after merge so overlay args are added to (not replace) the base args.
 func mergeContainerList(containers []corev1.Container, overlays map[string]*corev1.Container) error {
 	for i := range containers {
 		base := &containers[i]
 		if overlay := overlays[base.Name]; overlay != nil {
-			// Set the overlay's name to match the base to prevent strategic merge
-			// from overwriting the name with an empty string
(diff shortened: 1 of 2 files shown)


avi-biton:
> [correctness] operator/pkg/customization/pod.go — The fix does not deduplicate args. If a base container already has an arg that the overlay also adds, the result will contain duplicates. In practice this is unlikely since base args come from embedded manifests and overlay args from operator options, but worth noting for future callers.

It's technically accurate but not actionable. Deduplication would be wrong here -- some CLI tools accept repeated flags with different values (e.g. --allow=X --allow=Y), and silently dropping duplicates would be a surprising behavior change.

The current append semantics match how Kubernetes itself handles args -- it's just a []string, no uniqueness contract. If a future caller accidentally duplicates an arg, the worst case is a harmless redundant flag; deduplication could silently swallow intentional repeats.

yftacherzog (approved):
That got me a bit confused, but I think we should be good.

PR #6895, opened 2026-05-15
fix: increase kyverno deployment timeout

tisutisu:
Kyverno deployment taking longer, so increasing the timeout to 120s instead of 30s

```
  ⏳ Waiting for Kyverno admission controller...
error: timed out waiting for the condition on deployments/kyverno-admission-controller
🔄 Retrying command (attempt 2/3)...
error: timed out waiting for the condition on deployments/kyverno-admission-controller
🔄 Retrying command (attempt 3/3)...
error: timed out waiting for the condition on deployments/kyverno-admission-controller
kubectl wait --for=condition=Available deployment/kyverno-admission-controller -n kyverno --timeout=30s: Kyverno admission controller did not become available within the allocated time.

❌ Deployment failed during: Kyverno
```

### DIFF
--- a/deploy-deps.sh
+++ b/deploy-deps.sh
@@ -446,7 +446,7 @@ deploy_kyverno() {
     # by kyverno-admission-controller; if we apply policies before that deployment is ready,
     # the API server gets "connection refused" from the webhook. Wait for it to be Available.
     echo "  ⏳ Waiting for Kyverno admission controller..." >&2
-    retry "kubectl wait --for=condition=Available deployment/kyverno-admission-controller -n kyverno --timeout=30s" \
+    retry "kubectl wait --for=condition=Available deployment/kyverno-admission-controller -n kyverno --timeout=60s" \
           "Kyverno admission controller did not become available within the allocated time"
     # Default SET_SKIP_CHECKS=false: reduce-tekton only (pipeline security checks are not defaulted off).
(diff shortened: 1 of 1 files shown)


yftacherzog:
On what sort of environment do you see this happening? Does increasing the timeot address  that for you? Things working properly afterwards?

tisutisu:
> On what sort of environment do you see this happening? Does increasing the timeot address that for you? Things working properly afterwards?

@yftacherzog 
> On what sort of environment do you see this happening?

I am observing the issue while deploying konflux cluster locally sometimes (maybe network speed can be one factor), even observed this on latest CI runs [here](https://github.com/konflux-ci/build-service/pull/595/changes)

> Does increasing the timeot address that for you? Things working properly afterwards?

yes after increasing the timeout, it worked for me locally, the issue was deployment was only taking longer time than current timeout of 3*30sec

yftacherzog:
> yes after increasing the timeout, it worked for me locally, the issue was deployment was only taking longer time than current timeout of 3*30sec

So this is potentially growing from 90sec to 6 minutes.
Can we try a smaller increase?

tisutisu:
> > yes after increasing the timeout, it worked for me locally, the issue was deployment was only taking longer time than current timeout of 3*30sec
> 
> So this is potentially growing from 90sec to 6 minutes. Can we try a smaller increase?

I hope 3 min would be sufficient and deployment should succeed, changed it to 180 sec now.

yftacherzog (approved):

PR #6905, opened 2026-05-18
feat(KFLUXVNGD-966): add pacWebhookInsecureSSL CRD field to disable TLS verification

avi-biton:
Add a typed *bool field to KonfluxBuildServiceSpec that controls the
PAC_WEBHOOK_INSECURE_SSL environment variable with three-state semantics:

- true: injects PAC_WEBHOOK_INSECURE_SSL=1, overrides manager.env
- false: removes PAC_WEBHOOK_INSECURE_SSL, overrides manager.env
- omitted (nil): defers to upstream default or manager.env value

Using *bool instead of bool lets the operator distinguish "explicitly false"
from "not set", so the typed field always takes precedence when the user expresses intent in either direction.

Refactored buildBuildControllerManagerOverlay to extract nullable fields upfront
and build a single linear option chain - no early return, no duplication.
Typed CRD fields are applied after FromContainerSpec so they always win over same-named entries in manager.env.

Added WithoutEnv container option to cleanly remove env vars (used for the explicit-false case), with unit tests.

Updated the e2e sample CR and webhook URL documentation to use the new field instead of the raw env var.

Assisted-by: Cursor

### DIFF
--- a/operator/internal/controller/buildservice/konfluxbuildservice_controller.go
+++ b/operator/internal/controller/buildservice/konfluxbuildservice_controller.go
@@ -260,7 +260,8 @@ func applyBuildServiceDeploymentCustomizations(deployment *appsv1.Deployment, sp
 // On OpenShift, PAC_WEBHOOK_URL is never set — the build-service auto-discovers the PaC Route.
 // When webhookConfigMapName is non-empty, the webhook-config volume (baked into the manifest
 // with optional: true) is updated to reference the hashed ConfigMap.
-// User-provided env vars from the CR spec always take precedence.
+// Typed CRD fields (e.g. pacWebhookInsecureSSL) take precedence over
+// user-provided env vars in buildControllerManager.manager.env.
 func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildServiceSpec, clusterInfo *clusterinfo.Info, webhookConfigMapName string) *customization.PodOverlay {
 	var containerOpts []customization.ContainerOption
 	var podOpts []customization.PodOverlayOption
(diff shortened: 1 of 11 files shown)


yftacherzog (approved):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -285,20 +286,24 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
 	}
 
 	containerOpts = append(containerOpts,
-		customization.FromContainerSpec(spec.BuildControllerManager.Manager),
+		customization.FromContainerSpec(managerSpec),
 		customization.WithLeaderElection(),
 	)
 
+	if spec.PACWebhookInsecureSSL {
+		containerOpts = append(containerOpts, customization.WithEnvOverride("PAC_WEBHOOK_INSECURE_SSL", "1"))
```
yftacherzog: I wonder if this provides a no-surprises behavior.
Currently, the new field only takes precedence if it's true. If it's false, then the env var still wins.

Maybe if it's false, we need to clear the env var or something.

yftacherzog: Maybe we can change the CRD field to behave more like a flag than a boolean (not sure how that should look).

avi-biton: The question is: is that a problem? 
For this specific field, false and "not set" both mean the same thing - "keep SSL verification enabled (the secure upstream default)." 
There's no scenario where a user needs to say "I explicitly want false and override the env var" because:
1- If they want SSL enabled, they just don't set `PAC_WEBHOOK_INSECURE_SSL` in manager.env either
2 - Both manager.env and the typed field are under the same user's control — there's no external actor injecting the env var

yftacherzog: An external actor might be me from 6 months ago that changed things in another place and don't recall anymore. It isn't very much different comparing to the ""I explicitly want **true** and override the env var". No?

I think the contract is not clear enough. It seems to suggest that the specific fields wins over the env var, but it's not the case.

Is it a problem? Not a huge one. Not sure if anyone will ever run into it. I think we can do a simple fix to the code or the definition of the field to avoid it altogether.

avi-biton: I will try something

avi-biton: I updated the type to be `*bool` to get the expected behavior:

- true: injects `PAC_WEBHOOK_INSECURE_SSL=1`, overrides manager.env
- false: removes `PAC_WEBHOOK_INSECURE_SSL`, overrides manager.env
- omitted (nil): defers to upstream default or manager.env value

I also added a function to remove env vars from the container and using it in this update

yftacherzog (approved):

PR #6962, opened 2026-05-21
fix: use short lived tokens in UI's Dex

filariow:
Dex was still using a Long-Lived Token. With this change we want to use a bound ServiceAccount token.

Dex doesn't support hot-reloading, so this change proposes a reload mechanism inspired by [ArgoCD's Dex wrapper](https://github.com/argoproj/argo-cd/blob/b6c358436dc8657f2d976f482626cf14440a7b24/cmd/argocd-dex/commands/argocd_dex.go#L104).


operator/internal/controller/ui/konfluxui_controller.go
```diff
@@ -343,6 +349,23 @@ func applyUIServiceCustomizations(service *corev1.Service, ui *konfluxv1alpha1.K
...
+	}
+
+	// initialize Annotations if needed
+	if serviceAccount.Annotations == nil {
+		serviceAccount.Annotations = make(map[string]string, 1)
+	}
+
+	// set OpenShift Annotation
+	serviceAccount.Annotations[dex.OpenShiftRedirectURIAnnotation] = endpoint.JoinPath(dex.DexCallbackPath).String()
+}
```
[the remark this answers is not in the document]
filariow: this change is out of scope for this PR.

@gbenhaim can I have your opinion on the suggested change?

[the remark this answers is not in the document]
gbenhaim: Agree this is out of scope for this PR — the same mismatch existed in the old `reconcileOpenShiftOAuth` which also used the raw `endpoint`.

Opened #7018 to track the fix separately.

_This comment was posted by **Gal's Cursor**._

operator/internal/controller/ui/konfluxui_controller_test.go
```diff
@@ -656,12 +640,13 @@ var _ = Describe("KonfluxUI Controller", func() {
...
-			}, &corev1.ServiceAccount{}))).To(BeTrue())
-			Expect(errors.IsNotFound(k8sClient.Get(ctx, types.NamespacedName{
-				Name: dex.DexClientSecretName, Namespace: uiNamespace,
-			}, &corev1.Secret{}))).To(BeTrue())
+			Eventually(func(g Gomega) {
+				sa := &corev1.ServiceAccount{}
+				g.Expect(k8sClient.Get(ctx, types.NamespacedName{
+					Name: "dex", Namespace: uiNamespace,
+				}, sa)).To(Succeed())
+				g.Expect(sa.Annotations).NotTo(HaveKey(dex.OpenShiftRedirectURIAnnotation))
```
gbenhaim, line 651: **Nit (test naming):** This test is now checking that the `dex` SA exists but *without* the redirect annotation — it's no longer verifying that resources are absent. Consider renaming from "Should NOT create OpenShift OAuth resources when NOT running on OpenShift" to something like "Should not annotate dex ServiceAccount when NOT running on OpenShift". Same applies to the other similar tests below (ClusterInfo nil, explicitly disabled).

operator/internal/controller/ui/konfluxui_controller_test.go
```diff
@@ -750,12 +738,11 @@ var _ = Describe("KonfluxUI Controller", func() {
...
-					Name: dex.DexClientServiceAccountName, Namespace: uiNamespace,
-				}, &corev1.ServiceAccount{}))).To(BeTrue())
-				g.Expect(errors.IsNotFound(k8sClient.Get(ctx, types.NamespacedName{
-					Name: dex.DexClientSecretName, Namespace: uiNamespace,
-				}, &corev1.Secret{}))).To(BeTrue())
+				sa := corev1.ServiceAccount{}
+				g.Expect(k8sClient.Get(ctx, types.NamespacedName{
+					Name: "dex", Namespace: uiNamespace,
+				}, &sa)).To(Succeed())
+				g.Expect(sa.Annotations).NotTo(HaveKey(dex.OpenShiftRedirectURIAnnotation))
```
gbenhaim, line 748: **Nit (test naming):** This test now verifies that the annotation is removed from the `dex` SA — not that the SA/Secret are deleted. Consider renaming from "Should delete OpenShift OAuth resources when disabled after being enabled" to something like "Should remove OpenShift redirect annotation when disabled after being enabled".

gbenhaim (commented):

operator/upstream-kustomizations/ui/dex/kustomization.yml
```diff
@@ -1,10 +1,16 @@
...
+- digest: sha256:ef2f9c2ed6177d0a9c4759a7ada65c9196efc76353d870f8b6eca0039dd8b983
+  name: quay.io/konflux-ci/dex
+  newName: quay.io/konflux-ci/dex
+
+configMapGenerator:
+- files:
+  - run-dex.sh
+  name: run-dex
+  options:
+    disableNameSuffixHash: true
```
gbenhaim: this shold be false, so when we change the script the deployment will get updated. This can happen when upgrading the operator.

operator/upstream-kustomizations/ui/dex/run-dex.sh
```diff
@@ -0,0 +1,87 @@
...
+shutdown() {
+  log "received signal, shutting down"
+  if [ -n "${dex_pid:-}" ] && [ -d "/proc/${dex_pid}" ]; then
+    kill -s TERM "${dex_pid}" 2>/dev/null || true
+    wait_for_halt || true
+  fi
+  exit 143
+}
+
+wait_for_halt() (
```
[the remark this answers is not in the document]
gbenhaim: +1 on this. Since `wait_for_halt` runs in a subshell (parentheses), the parent shell never calls `wait` on the dex child process, so it becomes a zombie until bash implicitly reaps it. Over many token rotation cycles (~every 8 minutes), this could accumulate transiently.

The fix is straightforward — add `wait "$dex_pid" 2>/dev/null || true` after the `wait_for_halt` call in `start_with_reload` to explicitly reap the child.

_This comment was posted by **Gal's Cursor**._

operator/upstream-kustomizations/ui/dex/run-dex.sh
```diff
@@ -0,0 +1,87 @@
...
+    # run dex in a subprocess
+    OPENSHIFT_OAUTH_CLIENT_SECRET="$(cat ${ORIGINAL_TOKEN})" \
+    OPENSHIFT_OAUTH_CLIENT_ID="${OPENSHIFT_OAUTH_CLIENT_ID:-system:serviceaccount:${NAMESPACE}:${POD_SERVICE_ACCOUNT}}" \
+      /usr/local/bin/dex serve /etc/dex/cfg/config.yaml &
+    dex_pid="$!"
+    log "run Dex in background with pid ${dex_pid}"
+
+    # wait for the token to be refresh by kubelet
+    log "waiting for the token to be refreshed"
+    while [ "$(cksum < "${ORIGINAL_TOKEN}")" = "$(cksum < "${TOKEN_PATH}")" ]; do sleep "${WAIT_TIME}"; done
```
gbenhaim, line 56: **Concern: synchronized restarts across replicas**

In production we typically run 3 dex replicas. Since all pods start at the same time during a rollout, their projected tokens share the same creation time and `expirationSeconds: 600`. The kubelet rotates at ~80% lifetime, so all pods would rotate within seconds of each other — causing all replicas to restart nearly simultaneously and defeating the purpose of multiple replicas for availability.

Consider adding random jitter to the polling interval so pods detect the token change at staggered times, e.g.:

```bash
jitter=$(( RANDOM % 30 ))
while [ "$(cksum < "${ORIGINAL_TOKEN}")" = "$(cksum < "${TOKEN_PATH}")" ]; do sleep $(( WAIT_TIME + jitter )); done
```

This would spread restarts across ~30 seconds, so at least some replicas remain available during rotation.

gbenhaim (commented):

operator/upstream-kustomizations/ui/dex/run-dex.sh
```diff
@@ -0,0 +1,87 @@
...
+  done
+
+  log "the loop broke unexpectedly"
+  exit 1
+}
+
+main() {
+  trap shutdown TERM INT
+
+  if [ "${OPENSHIFT_LOGIN_ENABLED:-false}" == "true" ]; then
```
gbenhaim: **Question: testability of the OpenShift-only code path**

The `if/else` here creates two completely different runtime behaviors — on non-OpenShift (e.g., Kind), dex runs directly with a simple `wait`; on OpenShift, the entire reload loop runs (signal handling during restarts, token comparison, graceful shutdown/restart, zombie reaping).

Since our pre-merge CI only runs on Kind, the reload loop and all its edge cases are never exercised in CI. Have you considered always running the reload loop regardless of platform (just without setting `OPENSHIFT_OAUTH_CLIENT_SECRET` / `OPENSHIFT_OAUTH_CLIENT_ID` when not on OpenShift)? The projected token volume is already mounted unconditionally, so the rotation mechanism would work the same way — it would just restart dex periodically without injecting OpenShift credentials.

This would mean the same code path runs everywhere, and Kind CI would catch issues in the reload logic (signal races, zombie accumulation, restart timing, etc.) before they hit production on OpenShift.

gbenhaim (commented):

operator/upstream-kustomizations/ui/dex/run-dex.sh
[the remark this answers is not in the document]
filariow: this is intentional and should not accumulate. `wait` doesn't time out. `wait_for_halt` waits for `HALT_TIMEOUT ` seconds the process to terminate, if it doesn't the [whole process stop](https://github.com/filariow/konflux-ci/blob/e07d2698b0dfdc47510ed70a85677ef19b3a77d8/operator/upstream-kustomizations/ui/dex/run-dex.sh#L63-L64).

operator/upstream-kustomizations/ui/dex/run-dex.sh
filariow: this will add un-needed complexity in non-OpenShift clusters. How are we usually testing OpenShift-only code in konflux-ci? I'd rather suggest to add a CI job testing on OpenShift instead.

gbenhaim: We have a ci job for openshift, but in order to be productive, we run it on demand or when doing a release.

filariow: let's keep it simple then and only have one workflow at cost of a slightly higher complexity for non-OpenShift clusters

gbenhaim (approved):
_Reviewed by **Gal's Cursor**._

LGTM. The security improvement (short-lived projected tokens replacing long-lived SA token secrets) is solid, and the ArgoCD-inspired reload wrapper is well-implemented. The SSA-based annotation lifecycle, signal handling, jitter for replica coordination, and orphan cleanup of the old `dex-client` resources all look correct.

Two very minor nits (not blocking):
- The `UICleanupGVKs` comments (lines 105-108 in the controller) still say "ServiceAccount is optional" but the `dex` SA is now always-applied. The entries are still needed for migration cleanup, but the comments are slightly misleading.
- The `log "the loop broke unexpectedly"` at the end of `start_with_reload` is unreachable (the `while true` has no `break`).

Nice work!

PR #6973, opened 2026-05-25
Switch to custom Caddy build with impersonate plugin

gbenhaim:
## Summary

- Replace the regex-based `map` + `request_header` workaround (limited to 10 groups) with a custom Caddy build from [konflux-ci/reverse-proxy](https://github.com/konflux-ci/reverse-proxy) that includes the `impersonate` handler plugin
- The plugin splits `X-Auth-Request-Groups` into individual `Impersonate-Group` headers natively with no arbitrary group limit
- Switch the Caddy image from `registry.access.redhat.com/hi/caddy` to `quay.io/konflux-ci/reverse-proxy`

## Changes

- Replace `import impersonate-groups.conf` + `import impersonate-headers.conf` with the `impersonate` directive in the Caddyfile and `tekton-results.caddy`
- Delete `impersonate-groups.conf`, `impersonate-headers.conf`, and `ns-lister-headers.conf`
- Remove corresponding volumeMounts and ConfigMap entries from `proxy.yaml`
- Update `caddy-fmt` CI workflow for the new image name
- Update README and proxy tests

## Test plan

- [x] All 26 proxy e2e tests pass against a local Kind deployment with the new image


test/go-tests/proxy_test.go
avi-biton: We might want to also update this error message as we no longer have placeholders

.github/workflows/caddy-fmt.yaml
```diff
@@ -20,18 +20,17 @@ jobs:
       - name: Extract Caddy image reference
         id: image
         run: |
-          digest=$(grep -A1 'name: registry.access.redhat.com/hi/caddy' \
-            operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml \
-            | grep 'digest:' | awk '{print $2}')
-          echo "ref=registry.access.redhat.com/hi/caddy@${digest}" >> "$GITHUB_OUTPUT"
+          ref=$(yq '.images[] | select(.name == "quay.io/konflux-ci/reverse-proxy") | .newName + "@" + .digest' \
+            operator/upstream-kustomizations/ui/core/proxy/kustomization.yaml)
+          echo "ref=${ref}" >> "$GITHUB_OUTPUT"
```
avi-biton, line 26: The yq query can succeed (exit 0) but return empty/null if the image entry is renamed. Consider adding set -euo pipefail and a guard:
```
run: |
  set -euo pipefail
  ref=$(yq '...' kustomization.yaml)
  if [[ -z "$ref" || "$ref" == "null" ]]; then
    echo "ERROR: Failed to extract reverse-proxy image ref" >&2; exit 1
  fi
  echo "ref=${ref}" >> "$GITHUB_OUTPUT"
```
Posted by Avi's Cursor

avi-biton (approved):

filariow (approved):

PR #6974, opened 2026-05-25
fix(KFLUXVNGD-1005): truncate segment-bridge image tag to commit hash

Omeramsc:
The segment-bridge newTag contained a 64-char value (commit hash + stale manifest-list digest suffix) instead of the expected 40-char git commit hash. This caused ImagePullBackOff on every Konflux install since the 64-char tag does not exist in the registry.

Introduced in b042688b which set newTag to a full manifest-list digest; every subsequent Renovate and automation update propagated the extra 24-char suffix via substring replacement.

Truncate newTag to the 40-char commit hash matching ?ref= so the CronJob image reference resolves correctly.

Ref: https://redhat.atlassian.net/browse/KFLUXVNGD-1005

### DIFF
--- a/operator/pkg/manifests/segment-bridge/manifests.yaml
+++ b/operator/pkg/manifests/segment-bridge/manifests.yaml
@@ -94,7 +94,7 @@ spec:
             - secretRef:
                 name: segment-bridge-config
                 optional: true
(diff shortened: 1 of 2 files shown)


gbenhaim:
thanks @Omeramsc. did you chack that the pod runs properly and that telemetry is reported? also, this fix should be backported to the relese-.0.1 branch.

gbenhaim (approved):

PR #6985, opened 2026-05-26
feat(KFLUXVNGD-984): Add logLevel to KonfluxNamespaceLister CRD

avi-biton:
This change adds a new field to the KonfluxNamespaceLister CRD called logLevel. This field is used to specify the minimum severity for namespace-lister log output.

The field is optional and when omitted, the namespace-lister defaults to error level. logLevel is an enum with the following values: debug, info, warn, error.

- Add logLevel to KonfluxNamespaceLister CRD
- Add tests for logLevel
- Add tests for namespace-lister customization
- Add logLevel to sample yamls

The logLevel field is mapped to the slog integer the upstream namespace-lister expects. See See https://github.com/konflux-ci/namespace-lister/blob/main/internal/log/log.go
 for more details.
 
 ## Upgrade notes

Existing `KonfluxNamespaceLister` CRs (or `Konflux` CRs with a `namespaceLister` section)
should add the following field to preserve current behavior:

    spec:
      logLevel: info

When omitted, `logLevel` defaults to error-only logging.
 
 Assited-by: Cursor + Opus 4.6

### DIFF
--- a/operator/api/v1alpha1/konfluxnamespacelister_types.go
+++ b/operator/api/v1alpha1/konfluxnamespacelister_types.go
@@ -45,6 +45,12 @@ type KonfluxNamespaceListerSpec struct {
 	// +optional
 	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
 	CacheResyncPeriod string `json:"cacheResyncPeriod,omitempty"`
+
+	// LogLevel sets the minimum log severity for the namespace-lister.
+	// Accepted values: debug, info, warn, error.
+	// When omitted, the namespace-lister defaults to error level.
+	// +optional
+	LogLevel LogLevel `json:"logLevel,omitempty"`
 }
 
 // KonfluxNamespaceListerStatus defines the observed state of KonfluxNamespaceLister.
--- a/operator/api/v1alpha1/common_types.go
+++ b/operator/api/v1alpha1/common_types.go
@@ -32,6 +32,17 @@ type ContainerSpec struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
 
+// LogLevel sets the minimum severity for log output.
+// +kubebuilder:validation:Enum=debug;info;warn;error
+type LogLevel string
+
+const (
+	LogLevelDebug LogLevel = "debug"
+	LogLevelInfo  LogLevel = "info"
+	LogLevelWarn  LogLevel = "warn"
+	LogLevelError LogLevel = "error"
+)
+
 // ControllerManagerDeploymentSpec defines customizations for the controller-manager deployment.
 type ControllerManagerDeploymentSpec struct {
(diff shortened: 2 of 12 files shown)


avi-biton:
>The API introduces type LogLevel string at the shared v1alpha1 package level, which is likely to
collide with future components that also need a LogLevel type. Renaming it to a component-scoped
name (e.g., NamespaceListerLogLevel) reduces future API churn and ambiguity in the codebase.

I disagree. 
The log level enum values (debug, info, warn, error) are universal. 
If other components need a log-level field in the future, they can reuse this same type. 
A generic name in the shared v1alpha1 package is the right choice for a reusable type. 
Only rename if a component needs different enum values, which is unlikely for log levels.

avi-biton:
> applyNamespaceListerCustomizations converts spec.LogLevel via a map index and passes the result to
WithOptionalEnvOverride; any non-empty LogLevel value not present in the map results in an empty
string and becomes a silent no-op. This makes future enum expansion (or accidental LogLevel("...")
usage in Go) fail silently by omitting LOG_LEVEL and falling back to upstream defaults.

This issue was addressed

operator/api/v1alpha1/konfluxnamespacelister_types.go
```diff
@@ -31,6 +31,17 @@ type NamespaceListerDeploymentSpec struct {
...
+// LogLevel sets the minimum severity for namespace-lister log output.
+// +kubebuilder:validation:Enum=debug;info;warn;error
+type LogLevel string
+
+const (
+	LogLevelDebug LogLevel = "debug"
+	LogLevelInfo  LogLevel = "info"
+	LogLevelWarn  LogLevel = "warn"
+	LogLevelError LogLevel = "error"
+)
```
yftacherzog: I wonder if this can be generalized.
I suppose more services have it, but not sure they use the same log provider.
Maybe can just generalize per the log provider.

avi-biton: All the log interfaces I know are using these values (some of them adds more) so generalize them per provider might not needed here.
Since we are only use them here, I think we can leave it as it is for now and if we will need to reuse them or equivalents in other place we will generalize them.

yftacherzog: For example, in integration we have log level for the snapshot gc.
It's probably supported for others as well as an env var, but not used.

yftacherzog: Not saying we should implement for all, but maybe there are some aspects of the implementation that could be shared and then used once we need them.

avi-biton: Moved to `operator/api/v1alpha1/common_types.go`

yftacherzog (approved):

PR #6998, opened 2026-05-27
test: add concurrency group of e2e tests

yftacherzog:
This is mainly for making bots queue their e2e tests rather than running them in parallel.

Assisted-by: Cursor


yftacherzog:
closing for now

PR #7060, opened 2026-05-31
fix: use repositories instead of repository in RPAs

yftacherzog:
repository field is deprecated in favor of repositories. We need to have this updated as repository was recently removed from the RPA spec, which fails CI now if trying to update the upstream ref for release-service-catalog.

Assisted-by: Cursor

Addresses the failure in: https://github.com/konflux-ci/konflux-ci/pull/7030


yftacherzog:
I rebased the original PR on this branch, so this PR might not be needed after all.

PR #7065, opened 2026-05-31
fix: temp set toolchain to auto for tests on openshift

yftacherzog:
We moved to use go 1.26, but there's no
rhel-9-release-golang-1.26-openshift-4.21 base image yet, so for now we set the toolchain to auto for the tests on openshift, to allow us access to the go 1.26 toolchain.

Assisted-by: Cursor

### DIFF
--- a/deploy-konflux-on-ocp.sh
+++ b/deploy-konflux-on-ocp.sh
@@ -24,6 +24,11 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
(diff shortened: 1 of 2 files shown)


gbenhaim (changes requested):
We can use humming bird https://images.redhat.com/?name=go

yftacherzog:
> We can use humming bird https://images.redhat.com/?name=go

Our problem at the moment is with the image openshift ci is using for the tests. Are you suggesting we use it as base image there?

yftacherzog:
Planning to address it [here](https://github.com/openshift/release/pull/79899) instead.

PR #7077, opened 2026-06-01
chore(KFLUXVNGD-1007): ai skill for reviewing go upgrades

yftacherzog:
Recently, upgrading the go toolchain in this repo has caused CI to fail in unexpected places. This change introduces a skill that will facilitate AI agents in identifying potential future cases like that and ask the PR author to make sure the required steps are taken in this repo and others to minimize CI downtime.

Assisted-by: Cursor

### DIFF
--- a/skills/go-toolchain-upgrade/SKILL.md
+++ b/skills/go-toolchain-upgrade/SKILL.md
@@ -1,0 +1,76 @@
+---
+name: go-toolchain-upgrade
+description: >-
+  Use when konflux-ci/konflux-ci PRs touch go.mod, go.sum, the go directive,
+  golangci-lint or controller-gen pins, or deploy/e2e scripts that run go or make;
+  or when CI logs show go.mod requires go >=, GOTOOLCHAIN/local mismatch, or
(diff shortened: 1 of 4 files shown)


skills/go-toolchain-upgrade/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+**Triage first.** Most `go.mod`/`go.sum` PRs are dependency-only—do not block them for
+cross-repo checklists. Escalate only when minimum Go or CI **images** must change.
+
+## Triage
+
+| Signal | Result | Action |
+|--------|--------|--------|
+| Only `require`/`replace`/indirect changes; **`go` unchanged** | Routine | Short note; **stop** |
+| `go.sum` only | Routine | Same |
+| **`go` minor increases** (e.g. 1.25→1.26) in any module | Significant | Step 2 |
```
avi-biton: There's no "Step 2" heading - the section is called "Significant". 
"Step 2" in the table doesn't match the actual heading name. 
An agent would still find it since the Result column says "Significant" which matches, so it's purely cosmetic.

avi-biton (approved):

PR #7197, opened 2026-06-07
fix(ui): use shared cluster CA for namespace-lister TLS verification

gbenhaim:
Previously, the UI proxy verified namespace-lister's TLS certificate using a trust-manager Bundle (trusted-ca ConfigMap). This had two problems:

1. The Bundle was part of the internal-registry component, which defaults to disabled — leaving the ConfigMap missing on most deployments.
2. It introduced a dependency on trust-manager, a separate controller that watches Bundle CRs and distributes CA certs as ConfigMaps.

The root cause was that the UI had an independent CA hierarchy (self-signed via self-signed-cluster-issuer) with no trust relationship to the cluster-wide ca-issuer that signs namespace-lister's cert.

Fix this by issuing ui-ca from ca-issuer instead of self-signed-cluster-issuer, making it a sub-CA under the shared cluster root. Since cert-manager stores the issuing CA in the Secret's ca.crt field, the ui-ca Secret now contains the root CA cert — the same CA that signed namespace-lister's certificate. The proxy mounts this and uses it for TLS verification, eliminating the trust-manager dependency entirely.

Trust chain (after):

  root-secret (root CA, in cert-manager ns)
    └── ca-issuer (ClusterIssuer, signs all Konflux service certs)
          ├── namespace-lister cert (leaf)
          └── ui-ca (sub-CA, in konflux-ui ns)
                └── ui-ca-issuer (namespace Issuer)
                      ├── serving-cert (proxy TLS)
                      └── dex-cert (Dex TLS)

The OpenShift Route annotation (destination-ca-certificate-secret) continues to work: ui-ca Secret's tls.crt is still a CA cert (CA:TRUE), which the Ingress-to-Route controller requires.

Assisted-by: Cursor

### DIFF
--- a/operator/pkg/manifests/ui/manifests.yaml
+++ b/operator/pkg/manifests/ui/manifests.yaml
@@ -127,7 +127,7 @@ data:
   Caddyfile: "{\n\tauto_https off\n\tadmin off\n\torder inject_cached_vars before
     reverse_proxy\n\tfile_watcher {\n\t\tcache kube_token /var/run/secrets/konflux-ci.dev/serviceaccount/token\n\t\tcache
     backend_token /var/run/secrets/konflux-ci.dev/backend/token\n\t\twatch /var/run/secrets/kubernetes.io/serviceaccount\n\t\twatch
-    /mnt/trusted-ca\n\t\twatch /mnt/service-ca\n\t\twatch /mnt/serving-cert\n\t}\n}\n\n#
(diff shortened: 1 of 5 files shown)


yftacherzog (approved):
Other than the documentation comments raised by the bots, lgtm

avi-biton:
Other than the 2 comments raised: 
- Updating `ui/README` 
- Consider removing ` optional: true` from the secret part in the manifest

All looks good

avi-biton (approved):

PR #7207, opened 2026-06-10
fix(ui): make ui-ca a self-signed root to fix route TLS re-encryption

gbenhaim:
## Summary

- **Root cause**: cert-manager omits self-signed root CAs from `tls.crt` (per TLS spec). When `ui-ca` was an intermediate CA issued by `ca-issuer`, the OpenShift route's `destinationCACertificate` contained only the intermediate — HAProxy couldn't build a complete chain to a self-signed root, causing 503 errors on TLS re-encryption.
- **Fix**: Make `ui-ca` a self-signed root (issued by `self-signed-cluster-issuer`). Its `tls.crt` is now inherently self-signed, giving HAProxy a valid trust anchor: `serving-cert → ui-ca (self-signed root)`.
- **Namespace-lister trust**: A new `cluster-root-ref` Certificate (issued by `ca-issuer`) surfaces the cluster root CA in the namespace so the proxy can still verify namespace-lister TLS.

## Changes

| File | Change |
|------|--------|
| `operator/upstream-kustomizations/ui/certmanager/certificate.yaml` | Changed `ui-ca` issuerRef to `self-signed-cluster-issuer`; added `cluster-root-ref` Certificate |
| `operator/upstream-kustomizations/ui/core/proxy/proxy.yaml` | Changed `cluster-ca` volume secret from `ui-ca` to `cluster-root-ref` |
| `operator/upstream-kustomizations/cert-manager/README.md` | Updated architecture diagram and docs |
| `operator/pkg/manifests/ui/manifests.yaml` | Regenerated via `kustomize build` |

## Test plan

- [x] `make test` passes (all operator unit/integration tests)
- [x] Deploy to OpenShift and verify:
  - `kubectl get secret ui-ca -n konflux-ui -o jsonpath='{.data.tls\.crt}' | base64 -d | openssl x509 -text -noout` shows self-signed (Subject == Issuer)
  - Route no longer returns 503
  - Namespace-lister accessible through the UI proxy
  - Dex login flow works end-to-end

### DIFF
--- a/operator/pkg/manifests/ui/manifests.yaml
+++ b/operator/pkg/manifests/ui/manifests.yaml
@@ -590,7 +590,7 @@ spec:
           items:
           - key: ca.crt
             path: ca-bundle.crt
-          secretName: ui-ca
+          secretName: cluster-root-ref
       - configMap:
           name: openshift-service-ca.crt
           optional: true
@@ -606,6 +606,22 @@ spec:
             path: url
           optional: true
           secretName: segment-bridge-config
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: cluster-root-ref
+  namespace: konflux-ui
+spec:
+  commonName: cluster-root-ref
+  dnsNames:
+  - cluster-root-ref.konflux-ui.svc
+  duration: 87600h
+  issuerRef:
+    group: cert-manager.io
+    kind: ClusterIssuer
+    name: ca-issuer
+  secretName: cluster-root-ref
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
@@ -665,7 +681,7 @@ spec:
   issuerRef:
     group: cert-manager.io
     kind: ClusterIssuer
-    name: ca-issuer
+    name: self-signed-cluster-issuer
   privateKey:
     algorithm: ECDSA
     size: 256
--- a/operator/pkg/manifests/ui/manifests.yaml
+++ b/operator/pkg/manifests/ui/manifests.yaml
@@ -616,7 +616,6 @@ spec:
   commonName: cluster-root-ref
   dnsNames:
   - cluster-root-ref.konflux-ui.svc
-  duration: 87600h
   issuerRef:
     group: cert-manager.io
     kind: ClusterIssuer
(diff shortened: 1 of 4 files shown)


operator/upstream-kustomizations/cert-manager/README.md
avi-biton: I think thats not accurate anymore

avi-biton (approved):

operator/pkg/manifests/ui/manifests.yaml
```diff
@@ -609,6 +609,22 @@ spec:
...
 apiVersion: cert-manager.io/v1
 kind: Certificate
+metadata:
+  name: cluster-root-ref
+  namespace: konflux-ui
+spec:
+  commonName: cluster-root-ref
+  dnsNames:
+  - cluster-root-ref.konflux-ui.svc
+  duration: 87600h
```
yftacherzog: What's the reasoning behind the 10 years period? Do we need that to correspond to the numbers in other certificates (currently unset)?

yftacherzog: Cursor saying: If root-secret / ca-issuer rotates, cluster-root-ref's ca.crt won't refresh until this cert renews — potentially leaving the proxy with a stale cluster root and breaking namespace-lister TLS.

yftacherzog:
### Stale documentation (not updated in this PR)

The cert-manager README updates in this PR are good, but a few related docs still describe the old single-trust-domain / sub-CA model:

| File | Issue |
|------|-------|
| `operator/upstream-kustomizations/cert-manager/README.md` (bootstrap sequence, ~L41–51) | Still says `self-signed-cluster-issuer`'s "only job" is the cluster root, and that "all Konflux services" use `ca-issuer` — contradicts the updated architecture diagram in this PR. Likely what @avi-biton meant with the earlier comment. |
| `operator/upstream-kustomizations/ui/README.md` (~L262) | TLS table still implies the cluster root comes from a generic cert-manager Secret; should name `cluster-root-ref`. |
| `operator/docs/content/docs/guides/custom-url.md` | Still accurate functionally, but could note `ui-ca` is now a self-signed root rather than a sub-CA under the cluster root. |

Not blocking for the fix itself, but worth a follow-up commit before merge to avoid the doc drift that caused confusion here.

— Yftach's Cursor

yftacherzog (approved):
/test konflux-e2e-v420-optional
/test konflux-e2e-v420-arm64-optional

operator/pkg/manifests/ui/manifests.yaml
gbenhaim: fixed

PR #7210, opened 2026-06-14
test(KFLUXVNGD-1081): resolve ui url from cr

yftacherzog:
Proxy tests were hard coded to use localhost. To be able to run them on OpenShift, we need to be able to run them remotely. Resolving the UI URL from the CR provides a unified approach for getting the URL.

Assisted-by: Cursor

### DIFF
--- a/test/go-tests/k8s_client.go
+++ b/test/go-tests/k8s_client.go
@@ -1,0 +1,48 @@
+package go_tests
+
+import (
+	"context"
+	"fmt"
+
+	konfluxv1alpha1 "github.com/konflux-ci/konflux-ci/operator/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+var (
+	proxyScheme     = runtime.NewScheme()
+	proxyKubeConfig *rest.Config
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(proxyScheme))
+	utilruntime.Must(konfluxv1alpha1.AddToScheme(proxyScheme))
+}
+
+// NewClient builds a controller-runtime client from KUBECONFIG or ~/.kube/config.
+func NewClient() (crclient.Client, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	proxyKubeConfig = cfg
+	return crclient.New(cfg, crclient.Options{Scheme: proxyScheme})
+}
+
+func serviceProxyGet(ctx context.Context, namespace, serviceName, portName, path string) ([]byte, error) {
+	if proxyKubeConfig == nil {
+		return nil, fmt.Errorf("kubernetes client not initialized")
+	}
+	clientset, err := kubernetes.NewForConfig(proxyKubeConfig)
+	if err != nil {
+		return nil, err
+	}
+	return clientset.CoreV1().Services(namespace).
+		ProxyGet("http", serviceName, portName, path, nil).
+		DoRaw(ctx)
+}
--- a/test/go-tests/proxy_setup.go
+++ b/test/go-tests/proxy_setup.go
@@ -1,0 +1,112 @@
+package go_tests
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	konfluxv1alpha1 "github.com/konflux-ci/konflux-ci/operator/api/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	konfluxCRName              = "konflux"
+	konfluxReadyWaitTimeout    = 5 * time.Minute
+	konfluxHealthWaitTimeout   = 5 * time.Minute
+	konfluxReadyPollInterval   = 2 * time.Second
+	proxyHTTPRequestTimeout    = 30 * time.Second
+)
+
+var (
+	proxyBase       *url.URL
+	proxyHome       string
+	tokenURL        string
+	proxyHTTPClient *http.Client
+	proxyClient     crclient.Client
+)
+
+var _ = BeforeSuite(func() {
+	var err error
+	proxyClient, err = NewClient()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(proxyClient).NotTo(BeNil())
+
+	proxyHTTPClient = &http.Client{
+		Timeout: proxyHTTPRequestTimeout,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+
+	ctx := context.Background()
+
+	Eventually(func(g Gomega) {
+		konflux := &konfluxv1alpha1.Konflux{}
+		getErr := proxyClient.Get(ctx, crclient.ObjectKey{Name: konfluxCRName}, konflux)
+		g.Expect(getErr).NotTo(HaveOccurred())
+		g.Expect(konflux.IsReady()).To(BeTrue(), konflux.ReadyConditionMessage())
+
+		baseURL := strings.TrimSpace(os.Getenv("KONFLUX_PROXY_URL"))
+		if baseURL == "" {
+			baseURL = konflux.Status.UIURL
+			g.Expect(baseURL).NotTo(BeEmpty(), "expected konflux/konflux status.uiURL to be set once Ready")
+		}
+		g.Expect(setProxyBase(baseURL)).NotTo(HaveOccurred())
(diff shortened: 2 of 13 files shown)


test/go-tests/k8s_client.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"os"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func kubeconfigPath() (string, error) {
```
gbenhaim: Can we use the client provided by controller-runtime? It's much easier to use. Here is an example of how to create it https://github.com/konflux-ci/tekton-kueue/blob/1a1db766894a303eac441247e25bd0e1d37319dc/test/e2e/e2e_test.go#L788

test/go-tests/proxy_endpoint.go
```diff
@@ -0,0 +1,152 @@
...
+
+	healthTimeout := 5 * time.Minute
+	if override := os.Getenv("KONFLUX_PROXY_HEALTH_TIMEOUT"); override != "" {
+		if parsed, parseErr := time.ParseDuration(override); parseErr == nil {
+			healthTimeout = parsed
+		}
+	}
+
+	Eventually(func(g Gomega) {
+		resp, doErr := proxyHTTPClient.Get(proxyHome + "/health")
```
gbenhaim: /ask does the caddyfile expose the /health endpoint for external access?

test/go-tests/proxy_endpoint.go
```diff
@@ -0,0 +1,152 @@
...
+		g.Expect(doErr).NotTo(HaveOccurred())
+		defer resp.Body.Close()
+		g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+	}).WithTimeout(healthTimeout).WithPolling(pollInterval).Should(Succeed())
+})
+
+func konfluxReadyTimeout() time.Duration {
+	for _, key := range []string{"KONFLUX_READY_TIMEOUT", "E2E_KONFLUX_READY_TIMEOUT"} {
```
gbenhaim: Why do we need 2 different env vars?

test/go-tests/proxy_endpoint.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+		if cond["type"] == "Ready" && cond["status"] == "True" {
+			return true
+		}
+	}
+	return false
+}
+
+func konfluxReadyMessage(obj *unstructured.Unstructured) string {
```
gbenhaim: Let's add IsReady method to the konflux struct

test/go-tests/proxy_setup.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+	return "konflux Ready condition not found in status"
+}
+
+func konfluxUIURL(obj *unstructured.Unstructured) (string, bool, error) {
+	return unstructured.NestedString(obj.Object, "status", "uiURL")
+}
+
+func proxyURL(path string) string {
```
gbenhaim, line 93: I thin that we should use the endpoint structure from go's http lib

test/go-tests/proxy_endpoint.go
yftacherzog: Yes. `/health` is routed on the main public Caddy listener (`:9443`), not on a separate internal-only port. In the Caddyfile it is an unauthenticated handler that returns 200:

```caddyfile
# Health endpoint for liveness/readiness probes
handle /health {
    respond 200
}
```

See `operator/upstream-kustomizations/ui/core/proxy/Caddyfile` (same block is embedded in the deployed UI manifests).

That is the same site address used for `/api/k8s/*`, `/idp/*`, etc. There is no `forward_auth` on `/health`, so external callers (CI pod, laptop) get 200 without a token — which matches the proxy test expectations.

We also have empirical confirmation: OpenShift CI ran all 26 proxy specs against `status.uiURL` from the Konflux CR (no port-forward), including the unauthenticated `/health` entry, in ~28s.

So the `BeforeSuite` poll against `proxyHome + "/health"` is intentional and correct for remote/OCP runs.

— Yftach's Cursor

test/go-tests/k8s_client.go
yftacherzog: done

test/go-tests/proxy_endpoint.go
yftacherzog: I removed it. Initially, I was thinking about making the wait part of the test, but I don't think that's a good idea anymore. Anyhow, when that was done, a new env var was added on top of a pre-existing one that had a somewhat similar meaning.

test/go-tests/proxy_endpoint.go
yftacherzog: done

test/go-tests/proxy_setup.go
yftacherzog: done

test/go-tests/proxy_setup.go
```diff
@@ -0,0 +1,118 @@
...
+	Eventually(func(g Gomega) {
+		resp, doErr := proxyHTTPClient.Get(proxyURL("/health"))
+		g.Expect(doErr).NotTo(HaveOccurred())
+		defer resp.Body.Close()
+		g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+	}).WithTimeout(konfluxHealthWaitTimeout).WithPolling(konfluxReadyPollInterval).Should(Succeed())
+})
+
+func setProxyBase(raw string) error {
+	u, err := url.Parse(strings.TrimSpace(raw))
```
avi-biton, line 73: Don't we also need to handle `http://`?

test/go-tests/k8s_client.go
```diff
@@ -0,0 +1,36 @@
...
+
+var proxyScheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(proxyScheme))
+	utilruntime.Must(konfluxv1alpha1.AddToScheme(proxyScheme))
+}
+
+// CreateK8sClient builds a Kubernetes clientset from KUBECONFIG or ~/.kube/config.
+func CreateK8sClient() (*kubernetes.Clientset, error) {
```
gbenhaim: can we use `ctrl.GetConfigOrDie()` from controller runtime? I would expect to see a single method in this file which returns a client. Everything else should be handled by controller-runtime.

test/go-tests/proxy_setup.go
yftacherzog: Good catch. I now dropped the http case from `proxyWebSocketURL()`. Don't think there's a reason to support it in the test as the proxy is not exposing http to users. Cursor's Pavlovian addition most probably :)

test/go-tests/k8s_client.go
yftacherzog: done

test/go-tests/proxy_setup.go
```diff
@@ -0,0 +1,152 @@
...
+	return "konflux Ready condition not found in status"
+}
+
+func konfluxUIURL(obj *unstructured.Unstructured) (string, bool, error) {
+	return unstructured.NestedString(obj.Object, "status", "uiURL")
+}
+
+func proxyURL(path string) string {
+	return proxyHome + path
+}
```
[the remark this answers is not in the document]
yftacherzog: we only support https

test/go-tests/proxy_endpoint.go
```diff
@@ -0,0 +1,152 @@
...
+			g.Expect(found).To(BeTrue(), "expected konflux/konflux status.uiURL to be set once Ready")
+			g.Expect(baseURL).NotTo(BeEmpty())
+		}
+		proxyHome = strings.TrimRight(baseURL, "/")
+		tokenURL = proxyHome + "/idp/token"
+	}).WithTimeout(readyTimeout).WithPolling(pollInterval).Should(Succeed())
+
+	healthTimeout := 5 * time.Minute
+	if override := os.Getenv("KONFLUX_PROXY_HEALTH_TIMEOUT"); override != "" {
+		if parsed, parseErr := time.ParseDuration(override); parseErr == nil {
```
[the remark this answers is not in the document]
yftacherzog: That should be ok for now

test/go-tests/proxy_setup.go
```diff
@@ -0,0 +1,118 @@
...
+		g.Expect(konflux.IsReady()).To(BeTrue(), konflux.ReadyConditionMessage())
+
+		baseURL := strings.TrimSpace(os.Getenv("KONFLUX_PROXY_URL"))
+		if baseURL == "" {
+			baseURL = konflux.Status.UIURL
+			g.Expect(baseURL).NotTo(BeEmpty(), "expected konflux/konflux status.uiURL to be set once Ready")
+		}
+		g.Expect(setProxyBase(baseURL)).NotTo(HaveOccurred())
+	}).WithTimeout(konfluxReadyWaitTimeout).WithPolling(konfluxReadyPollInterval).Should(Succeed())
+
```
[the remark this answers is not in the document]
yftacherzog: That should be ok for now

test/go-tests/k8s_client.go
```diff
@@ -0,0 +1,48 @@
...
+func NewClient() (crclient.Client, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	proxyKubeConfig = cfg
+	return crclient.New(cfg, crclient.Options{Scheme: proxyScheme})
+}
+
+func serviceProxyGet(ctx context.Context, namespace, serviceName, portName, path string) ([]byte, error) {
```
gbenhaim, line 37: /ask is this function used anywhere?

yftacherzog: It's called here: https://github.com/konflux-ci/konflux-ci/blob/proxy-tests/test/go-tests/proxy_test.go#L314

gbenhaim (approved):

avi-biton (approved):

PR #7354, opened 2026-06-21
test(KFLUXVNGD-1081): default to not deploying test resources

yftacherzog:
Only Kind deployments for CI actually use the test resources (for specific proxy tests only relevant in Kind).

This change make it so that we don't deploy test resources unless we specifically need those.

Assisted-by: Cursor


gbenhaim:
> Only Kind deployments for CI actually use the test resources (for specific proxy tests only relevant in Kind).

We also run the proxy tests on Openshift.

> 
> This change make it so that we don't deploy test resources unless we specifically need those.
> 
> Assisted-by: Cursor

yftacherzog:
> > Only Kind deployments for CI actually use the test resources (for specific proxy tests only relevant in Kind).
> 
> We also run the proxy tests on Openshift.

We run a slightly different set of proxy tests on openshift. Namely, we only run the ones that run against default-tenant. So we don't need the test resources.

gbenhaim (approved):

PR #7563, opened 2026-06-25
feat(metrics): add configurable HTTPS metrics scraping

yftacherzog:
Enable Prometheus to scrape operator and component /metrics over HTTPS
using short-lived bearer tokens minted via TokenRequest instead of
static secrets.

Gate scrape resources with spec.componentMetrics on the Konflux CR
(enabled by default). The operator ensures a metrics-scraper
ServiceAccount per operand namespace, binds metrics-reader CRBs to it,
and rotates prometheus-scrape-token Secrets. Operand controllers refresh
tokens on a shared ticker; the operator uses a dedicated rotator with
adaptive timing because it has no operand controller and also owns the
operator ServiceMonitor.

Add cluster integration tests (metricsauth.DefaultCatalog) that
port-forward and scrape endpoints the way Prometheus would, to catch
RBAC, token, and ServiceMonitor regressions on Kind.

Assisted-by: Cursor


test/fixtures/metrics-targets.yaml
```diff
@@ -0,0 +1,48 @@
...
+targets:
+  - id: konflux-operator
+    group: operator
+    namespace: konflux-operator
+    service: konflux-operator-controller-manager-metrics-service
+    scheme: https
+    portName: https
+    port: 8443
+    path: /metrics
+    tlsInsecureSkipVerify: true
```
yftacherzog: Can we use tls already here and for build-service?
```suggestion
    tlsInsecureSkipVerify: false
```

yftacherzog: We'll do it in next PRs

operator/config/samples/konflux_v1alpha1_konflux.yaml
```diff
@@ -168,6 +168,12 @@ spec:
   # Defaults to true if not specified.
   defaultTenant:
     enabled: true
+  # componentMetrics:
```
gbenhaim: **Concern — default local deploy regression on vanilla Kubernetes**

The `componentMetrics` section is commented out here, but `IsComponentMetricsEnabled()` defaults to `true` when unset. On vanilla Kubernetes (Kind), `ResolveMetricsScraperServiceAccount` returns `ErrMetricsScraperNotConfigured` because there's no OpenShift UWM default to fall back to.

This means a plain `./scripts/deploy-local.sh` (without `QUAY_*` vars, which resolves to this file) will leave build-service and image-controller in an error loop:

1. `applyManifests` reaches the `prometheus-build-service-metrics-reader` CRB (annotated with `metrics-scraper-binding: "true"`)
2. `ApplyMetricsScraperBindingSubjects` → `ResolveMetricsScraperServiceAccount` → `ErrMetricsScraperNotConfigured`
3. `applyManifests` returns error → reconciler reports `ApplyFailed`

Core resources (namespace, deployments, services) are applied before the CRB so the services themselves run, but the reconciler stays in an error state.

Suggestion — either:
- Uncomment this block and point it to the `konflux-monitoring-test/prometheus` SA that `deploy-deps.sh` now creates, or
- Make the reconcilers gracefully skip CRB subject resolution when no scraper is configured (log a warning, leave subjects empty) instead of returning an error

deploy-deps.sh
```diff
@@ -392,15 +397,24 @@ deploy_prometheus_operator_crds() {
...
     kubectl apply -k "${script_path}/dependencies/prometheus-operator-crds"
 }
 
+deploy_metrics_scraper_rbac() {
+    : "${SKIP_METRICS_SCRAPER_RBAC:=false}"
+    if [[ "${SKIP_METRICS_SCRAPER_RBAC}" == "true" ]]; then
+        echo "⏭️  Skipping metrics scraper RBAC (SKIP_METRICS_SCRAPER_RBAC=true)" >&2
+        return 0
+    fi
+    kubectl apply -f "${script_path}/test/fixtures/metrics-scraper/rbac.yaml"
```
gbenhaim: **Suggestion — test fixtures deployed into the production dependency path**

This applies `test/fixtures/metrics-scraper/rbac.yaml` during regular dependency deployment, creating a `konflux-monitoring-test` namespace and `prometheus` ServiceAccount in every Kind cluster — including non-E2E ones.

Blurring test infrastructure into the deployment path could confuse users who inspect their cluster and find a `konflux-monitoring-test` namespace they didn't ask for.

Consider either:
- Moving this to E2E test scripts only (and having the default Konflux sample set `componentMetrics.enabled: false` to avoid the scraper requirement), or
- Promoting this to a proper deployment dependency (e.g. `dependencies/metrics-scraper/`) to signal it's a runtime need, not just a test fixture

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -193,12 +233,26 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 func (r *KonfluxBuildServiceReconciler) applyManifests(ctx context.Context, tc *tracking.Client, owner *konfluxv1alpha1.KonfluxBuildService, webhookConfigMapName string) error {
 	log := logf.FromContext(ctx)
 
+	metricsEnabled, err := common.ComponentMetricsEnabled(ctx, r.Client)
```
gbenhaim: **Nit — `ComponentMetricsEnabled` called twice per reconcile**

This is the second call to `ComponentMetricsEnabled` within the same reconciliation (the first is at line 183 in `Reconcile()`). Both read the Konflux CR from the API server. If the CR changes between the two calls, they could disagree — e.g. metrics get disabled between `Reconcile` and `applyManifests`, causing the scrape token to be rotated for resources that were just skipped.

The race window is tiny and the consequence is just a retry on the next reconcile, but it would be cleaner to call once at the top of `Reconcile()` and pass the result to `applyManifests()` as a parameter. Same pattern applies to image-controller.

gbenhaim (commented):

operator/config/samples/konflux_v1alpha1_konflux.yaml
gbenhaim: /ask this a valid concern?

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -165,6 +179,32 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return errHandler.HandleApplyError(ctx, err)
 	}
 
+	result := ctrl.Result{}
+	metricsEnabled, err := common.ComponentMetricsEnabled(ctx, r.Client)
+	if err != nil {
+		return errHandler.HandleWithReason(ctx, err, condition.ReasonApplyFailed, "read component metrics setting")
+	}
+	if metricsEnabled && r.TokenCreator != nil {
+		scraper, scrapeResolveErr := common.ResolveMetricsScraperServiceAccount(ctx, r.Client, r.ClusterInfo)
```
gbenhaim: **Request: avoid reading the parent Konflux CR from sub-CR reconcilers.**

This call (and the `ResolveMetricsScraperServiceAccount` call below) reaches back to the Konflux CR from the build-service reconciler, creating a circular dependency. Every other config field in this operator flows top-down: the Konflux reconciler writes fields into the sub-CR spec, and the sub-CR reconciler reads its own spec.

Please add a `componentMetrics` field to the sub-CR specs and have the Konflux reconciler forward the global config to each sub-CR, the same way it forwards `spec.buildService.spec` → `KonfluxBuildServiceSpec`, etc. The sub-CR reconciler should then read `owner.Spec.ComponentMetrics` (its own spec) instead of calling `common.ComponentMetricsEnabled(ctx, r.Client)` which reads the parent Konflux CR.

This also eliminates the need for the `Watches(&konfluxv1alpha1.Konflux{})` + `KonfluxComponentMetricsChangedPredicate` in `SetupWithManager` — the sub-CR's own spec update from the Konflux reconciler will trigger reconciliation naturally via the existing `For()` watch.

gbenhaim (changes requested):

**Architecture concern: sub-CR reconcilers should not read the parent Konflux CR directly.**

Currently, the build-service, image-controller, and integration-service reconcilers reach back to the Konflux CR at reconcile time (via `common.ComponentMetricsEnabled` and `common.ResolveMetricsScraperServiceAccount`) instead of reading their own CR's spec. This creates a circular dependency between the Konflux CR and the sub-CRs and is inconsistent with how every other config field flows in this operator (Konflux CR → sub-CR spec → sub-CR reconciler).

The suggested change:

1. Add a `componentMetrics` field to each sub-CR spec (e.g. `KonfluxBuildServiceSpec`, `KonfluxImageControllerSpec`, `KonfluxIntegrationServiceSpec`)
2. Have the Konflux reconciler forward `spec.componentMetrics` from the Konflux CR to each sub-CR's spec (same pattern as all other config fields)
3. Sub-CR reconcilers read metrics config from their own CR spec instead of reaching back to the Konflux CR
4. Remove the `Watches(&konfluxv1alpha1.Konflux{}, ...)` + `KonfluxComponentMetricsChangedPredicate` from each sub-CR controller — the sub-CR's own spec change triggers reconciliation naturally via the existing `For()` watch

This eliminates:
- The circular dependency (sub-CRs reading back up to their parent)
- The extra `Watches` + custom predicate machinery on each controller
- The duplicate `ComponentMetricsEnabled` calls per reconcile (once in `Reconcile()`, once in `applyManifests()`)
- The need for `common.ComponentMetricsEnabled` and `common.ResolveMetricsScraperServiceAccount` to read the Konflux CR themselves

And aligns with the established config-forwarding pattern used by every other field in this operator.

operator/pkg/clusterinfo/metrics_scraper.go
```diff
@@ -0,0 +1,50 @@
...
+import "k8s.io/apimachinery/pkg/types"
+
+const (
+	// OpenShiftUWMNamespace is the namespace for OpenShift user-workload Prometheus.
+	OpenShiftUWMNamespace = "openshift-user-workload-monitoring"
+	// OpenShiftUWMScraperSA is the default OpenShift UWM Prometheus scraper identity.
+	OpenShiftUWMScraperSA = "prometheus-user-workload"
+	// LegacyOpenShiftUWMScraperSA is the legacy OpenShift UWM Prometheus ServiceAccount name.
+	LegacyOpenShiftUWMScraperSA = "prometheus-k8s"
+)
```
gbenhaim: **Suggestion: the configurable scraper SA and the OpenShift UWM defaults are unnecessary in this architecture.**

Since the operator controls the entire auth chain — it mints the token, writes it to the `prometheus-scrape-token` Secret, and manages the CRB subjects — it doesn't matter which SA the token is minted for. The token is a bearer token: whoever presents it (Prometheus) is authenticated as the SA it was minted for, not as Prometheus's own identity. Prometheus gets the token from `bearerTokenSecret`, not from its pod-mounted SA token.

The operator could instead create its own dedicated SA (e.g. `metrics-scraper` in each operand namespace), mint tokens for it, and bind it in the CRB. No user configuration or platform detection needed.

This would eliminate:
- `spec.componentMetrics.scraper` (no config needed on any platform)
- The OpenShift UWM defaults and legacy `prometheus-k8s` subject handling
- `ErrMetricsScraperNotConfigured` (the error that breaks vanilla K8s deploys without explicit config)
- This entire file (`clusterinfo/metrics_scraper.go`)
- `common.ResolveMetricsScraperServiceAccount` and `MetricsReaderBindingSubjects`
- The `deploy-deps.sh` test scraper RBAC step

The legacy subject (`prometheus-k8s`) is also ineffective in the current design: the token is minted for `prometheus-user-workload`, so even if `prometheus-k8s` is authorized by the CRB, it wouldn't have the matching token to authenticate with.

gbenhaim (commented):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -165,6 +179,32 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			Scraper:          scraper,
+			OperandNamespace: webhookConfigNamespace,
+			Apply: func(applyCtx context.Context, secret *corev1.Secret) error {
+				return tc.ApplyOwned(applyCtx, secret)
+			},
+		})
+		if scrapeErr != nil {
+			return errHandler.HandleWithReason(ctx, scrapeErr, condition.ReasonApplyFailed, "reconcile prometheus scrape token")
+		}
+		result = common.MergeRequeueAfter(result, scrapeResult)
```
gbenhaim: **Suggestion: replace `RequeueAfter` with a channel source for periodic token rotation.**

`RequeueAfter` is a convergence mechanism ("I couldn't reach desired state, try again later"), not a scheduling mechanism. Token rotation is periodic maintenance — recurring work that never converges and has nothing to do with the CR's desired state. Using `RequeueAfter` for it turns an event-driven controller into a polling controller.

Concrete problems with the `RequeueAfter` approach:

- **Timing drift.** Any event (deployment rollout, config change, owned resource update) cancels the pending requeue and resets it after the new reconcile. In a busy cluster, the requeue fires at unpredictable times.
- **Concern conflation in the return value.** `MergeRequeueAfter` merges "did I converge?" with "when should I do maintenance?" into one `ctrl.Result`. If the reconciler has a real convergence-related requeue (e.g. waiting for a deployment), it fights with the token rotation requeue for the shortest interval.
- **Full reconcile on every tick.** Every periodic requeue runs the entire `Reconcile()` — all manifests re-applied, all customizations re-evaluated, orphan cleanup, status update — just to refresh a token.
- **Invisible intent.** When a reconcile fires, there's no way to tell if it was triggered by a CR change, an owned resource update, or a periodic requeue.

**Suggested alternative:** a single `manager.Runnable` ticker that emits `GenericEvent`s to a `source.Channel` watched by each sub-CR controller:

```
Single ticker Runnable ──timer──┬─▶ build-service controller channel
  (one goroutine, no K8s calls) ├─▶ image-controller controller channel
                                └─▶ future components...
```

The Runnable is trivial — just a ticker that sends events on a fixed interval. It does no K8s work, needs no namespace knowledge, no ownership handling. All token logic (freshness check, minting, `applyOwned`) stays in the reconciler.

Because the reconciler checks freshness before minting, duplicate triggers are harmless — if a normal reconcile already refreshed the token, the timer-triggered reconcile sees it's fresh and skips. The `RequeueAfter` plumbing and `MergeRequeueAfter` can be removed entirely.

gbenhaim (commented):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
gbenhaim: _Posted by **Gal's Cursor**._

**Acknowledged trade-off:** the ticker emits events even when `componentMetrics` is disabled, causing unnecessary reconciles.

In practice this is negligible — the reconciler already guards with `if metricsEnabled && r.TokenCreator != nil`, so a timer-triggered reconcile when metrics are off is a no-op: read the Konflux CR from the informer cache (no API call), see `metricsEnabled == false`, return immediately. No manifest application, no K8s writes, no token minting. On a ~1h interval, a few extra no-op reconciles are the same cost as any other spurious event (owned resource update, informer resync), and the reconciler is already idempotent by design.

test/fixtures/metrics-targets.yaml
```diff
@@ -0,0 +1,64 @@
+# Metrics scrape targets for cluster integration tests.
+# Each target has a Ginkgo group label: operator (Konflux operator) or component (operands).
+scraper:
+  namespace: konflux-monitoring-test
+  serviceAccount: prometheus
+
```
gbenhaim: **Suggestion: replace YAML catalog with Go structs.**

This YAML file is only consumed by the Go test suite (`metricsintegration`). The `Catalog` and `Target` struct definitions already exist in Go — the YAML just duplicates that schema in a less safe format.

A `var defaultTargets = []Target{...}` in a Go file would:
- Give compile-time type safety and IDE autocomplete
- Catch typos at build time instead of at runtime via `panic` in `loadMetricsCatalog`
- Eliminate the `validate()` function (the compiler does it for free)
- Remove the `findRepoRoot()` helper and `KONFLUX_REPO_ROOT` env var dependency
- Remove the runtime deserialization step

The typical justification for externalizing test data into YAML is when non-developers need to edit it, or when the same data is consumed by multiple tools (e.g., a CI script and Go tests). Neither applies here — only developers add/edit targets, and the catalog is only consumed by Go.

test/go-tests/metricsintegration/metrics_test.go
```diff
@@ -0,0 +1,63 @@
...
+	if target.ScrapeTokenSecret != "" {
+		Eventually(func(g Gomega) {
+			var tokenErr error
+			token, tokenErr = metricsauth.SecretToken(ctx, kubeClient, target.Namespace, target.ScrapeTokenSecret, "token")
+			g.Expect(tokenErr).NotTo(HaveOccurred())
+			g.Expect(token).NotTo(BeEmpty())
+		}).WithTimeout(metricsReadyTimeout).WithPolling(metricsReadyInterval).Should(Succeed())
+	} else {
+		token, err = metricsauth.ServiceAccountToken(ctx, kubeREST, metricsCatalog.Scraper.Namespace, metricsCatalog.Scraper.ServiceAccount)
+		Expect(err).NotTo(HaveOccurred())
```
gbenhaim: **Concern: inconsistent token acquisition weakens E2E coverage.**

The test uses two different token paths depending on the target:
- `build-service` and `image-controller` → read from the operator-managed `prometheus-scrape-token` secret
- `integration-service` and `konflux-operator` → mint a fresh token directly via the K8s TokenRequest API

The second path bypasses the operator's token management entirely. This means if the operator had a bug where it never wrote the scrape token secret for these components, these tests would still pass — they're testing "does RBAC allow scraping?" rather than "did the operator create a valid scrape token?".

For consistency, every target that uses bearer token auth should read from the operator-created secret. The E2E test should validate the full flow: operator creates secret → test reads secret → test scrapes with that token.

(I understand integration-service currently uses HTTP and may not be wired into the token-based auth flow yet, and the operator target uses a separate `ScrapeTokenRotator` — but the test should still validate the secret the rotator creates rather than minting its own token.)

gbenhaim (commented):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
yftacherzog: This is something I was considering before implementation. I decided against it at first for simplicity, but I'll make the change now.

avi-biton:
Since we are now creating secrets and apply them by `ApplyOwned`, we want to also watch these secrets by the controllers that creates them.
We need to add `Owns(&corev1.Secret{})` to the controllers that creates these secrets (`build-service`, `image-controller`) and the needed tests for them

avi-biton:
### Broadcaster interval vs. token freshness

`TokenRotationBroadcaster` ticks every 1 minute (`DefaultScrapeTokenMinRequeue`), triggering a full reconcile on every subscribed controller. With a 1-hour TTL and 50% refresh threshold, the token only needs refreshing every 30 minutes.

This means ~29 out of 30 broadcaster-triggered reconciles are effectively no-ops: they `Get` the Secret, check the expiry annotation, re-`Apply` for ownership tracking, then exit. That's acceptable overhead, but worth being explicit about the intent.

**Question:** Is the 1-minute tick intentionally compensating for the missing `Owns(&corev1.Secret{})` watch (so deleted Secrets are re-created within ~1 min)? If so, a comment on the `Interval` field or `NewTokenRotationBroadcaster` explaining this tradeoff would help — otherwise it looks like it could be safely bumped to 30 min without anyone realizing the self-healing window would widen.

If adding `Owns(&corev1.Secret{})` (with a name predicate) is on the table, the broadcaster interval could be relaxed to the refresh period and the two mechanisms would complement each other:
- Watch → immediate recovery from deletion
- Broadcaster → guaranteed rotation even if the watch event is missed

_Posted by Avi's Cursor_

yftacherzog:
> Since we are now creating secrets and apply them by `ApplyOwned`, we want to also watch these secrets by the controllers that creates them. We need to add `Owns(&corev1.Secret{})` to the controllers that creates these secrets (`build-service`, `image-controller`) and the needed tests for them

done

yftacherzog:
> ### Broadcaster interval vs. token freshness
> `TokenRotationBroadcaster` ticks every 1 minute (`DefaultScrapeTokenMinRequeue`), triggering a full reconcile on every subscribed controller. With a 1-hour TTL and 50% refresh threshold, the token only needs refreshing every 30 minutes.
> 
> This means ~29 out of 30 broadcaster-triggered reconciles are effectively no-ops: they `Get` the Secret, check the expiry annotation, re-`Apply` for ownership tracking, then exit. That's acceptable overhead, but worth being explicit about the intent.
> 
> **Question:** Is the 1-minute tick intentionally compensating for the missing `Owns(&corev1.Secret{})` watch (so deleted Secrets are re-created within ~1 min)? If so, a comment on the `Interval` field or `NewTokenRotationBroadcaster` explaining this tradeoff would help — otherwise it looks like it could be safely bumped to 30 min without anyone realizing the self-healing window would widen.
> 
> If adding `Owns(&corev1.Secret{})` (with a name predicate) is on the table, the broadcaster interval could be relaxed to the refresh period and the two mechanisms would complement each other:
> 
> * Watch → immediate recovery from deletion
> * Broadcaster → guaranteed rotation even if the watch event is missed
> 
> _Posted by Avi's Cursor_

The 1 minute value was not set because owns() were omitted. I think it could be loosened nonetheless, but as it's a no-op, I think it should be safe to leave it as-is. I added some doc comments.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -78,23 +85,31 @@ const (
...
 	client.Client
-	Scheme      *runtime.Scheme
-	ObjectStore *manifests.ObjectStore
-	ClusterInfo *clusterinfo.Info
+	Scheme        *runtime.Scheme
+	ObjectStore   *manifests.ObjectStore
+	ClusterInfo   *clusterinfo.Info
+	TokenCreator  kubernetes.TokenCreator
+	Clock         clock.Clock
+	TokenRotation *common.TokenRotationBroadcaster
```
gbenhaim: **Suggestion: accept a channel instead of a `*TokenRotationBroadcaster` reference.**

The reconciler never calls any method on `TokenRotation` during `Reconcile()` — it only calls `Subscribe()` once in `SetupWithManager`. The reconciler doesn't need to know about the broadcaster; it just needs a channel of events.

Instead of storing `TokenRotation *common.TokenRotationBroadcaster` on the struct, accept a `chan event.TypedGenericEvent[client.Object]` (or leave it nil when rotation isn't needed). The caller in `main.go` would call `broadcaster.Subscribe()` and pass the channel directly:

```go
// In the reconciler struct:
TokenRotationEvents <-chan event.TypedGenericEvent[client.Object]

// In main.go:
rotationCh := broadcaster.Subscribe()
reconciler := &KonfluxBuildServiceReconciler{
    // ...
    TokenRotationEvents: rotationCh,
}
```

This also lets you **remove the `sync.RWMutex` from `TokenRotationBroadcaster`** entirely. If `main.go` calls `Subscribe()` for all controllers before adding the broadcaster as a Runnable, the `subscribers` slice is built sequentially during setup and only read during `Start()` — no concurrent access. The `controller-runtime` manager already guarantees that `SetupWithManager` completes before `mgr.Start()` runs Runnables.

Benefits:
- Reduces coupling — the reconciler depends on a Go channel, not a concrete broadcaster type
- Simplifies testing — tests pass a plain channel without constructing a broadcaster
- Removes the mutex — the broadcaster becomes ~20 lines simpler
- Follows the principle that the wiring layer (main.go) owns dependency construction, not the consumer

gbenhaim (commented):

operator/internal/common/scrape_token_rotation.go
```diff
@@ -0,0 +1,105 @@
...
+		Object: &metav1.PartialObjectMetadata{
+			TypeMeta:   metav1.TypeMeta{Kind: "ScrapeTokenRotation", APIVersion: "konflux.konflux-ci.dev/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: "rotation"},
+		},
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
```
gbenhaim, line 82: **Bug: subscriber channels are never closed on shutdown.**

When the context is cancelled, `Start` returns without closing the subscriber channels. This leaves the `source.Channel` watchers hanging indefinitely. Close them before returning:

```go
case <-ctx.Done():
	for _, ch := range b.subscribers {
		close(ch)
	}
	return nil
```

This is safe — after `ctx.Done()`, the ticker is stopped and `broadcast()` will never run again, so there's no risk of sending on a closed channel.

gbenhaim (commented):

operator/internal/operatormetrics/scrape_token_rotator.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	for {
+		requeue, err := r.reconcile(ctx, namespace)
+		if err != nil {
+			log.Error(err, "failed to reconcile operator metrics scrape wiring")
+			requeue = kubernetes.DefaultScrapeTokenMinRequeue
+		}
+
+		timer := time.NewTimer(requeue)
```
gbenhaim: **Question: why does this use a per-iteration `time.NewTimer` while `TokenRotationBroadcaster` uses a fixed `time.Ticker`?**

The `requeue` value here comes from `ScrapeTokenRequeueAfter`, which in practice will almost always be 30 minutes (half the 1h TTL). It would only differ if the token had an unusual TTL or the clock drifted between calls. Since the reconciler already checks freshness and skips minting when the token is still valid, a fixed-interval ticker would work the same way — tick every 30 minutes, check, skip if fresh.

That's exactly what `TokenRotationBroadcaster` does for operand controllers. Having two different scheduling mechanisms for the same problem (periodic token refresh) is confusing. Could this be simplified to a ticker for consistency?

gbenhaim (commented):

gbenhaim (commented):

operator/internal/operatormetrics/scrape_token_rotator.go
```diff
@@ -0,0 +1,143 @@
...
+	clk := r.Clock
+	if clk == nil {
+		clk = clock.RealClock{}
+	}
+
+	metricsConfig, err := r.componentMetricsConfig(ctx)
+	if err != nil {
+		return kubernetes.DefaultScrapeTokenMinRequeue, err
+	}
+	if metricsConfig != nil && !metricsConfig.IsEnabled() {
```
gbenhaim: **Concern: the operator's `ScrapeTokenRotator` reads the Konflux CR to check `componentMetrics`.**

The operator can run without a Konflux CR existing — it's started before any CR is created. Coupling the operator's own metrics enablement to the Konflux CR introduces a dependency that doesn't belong here. The operator's metrics scraping should be controlled independently (e.g., via CLI flags or environment variables at startup), not by reading a CR that may or may not exist.

The sub-CR controllers correctly receive their `componentMetrics` config via the push model (Konflux controller forwards it to the sub-CR spec). The operator itself isn't a sub-CR, so it shouldn't be reading the Konflux CR directly either — it should use its own configuration mechanism.

operator/pkg/kubernetes/component_metrics.go
```diff
@@ -0,0 +1,119 @@
...
+				return true
+			}
+			switch o := obj.(type) {
+			case *corev1.Secret:
+				if o.Type != corev1.SecretTypeServiceAccountToken {
+					return false
+				}
+				return isMetricsReaderServiceAccountName(o.Annotations["kubernetes.io/service-account.name"])
+			case *unstructured.Unstructured:
+				name, _, _ := unstructured.NestedString(o.Object, "metadata", "name")
```
gbenhaim, line 77: **Nit: redundant name re-read for unstructured Secret.**

`name` is already assigned from `obj.GetName()` on line 51, which works for `*unstructured.Unstructured`. This block re-reads `metadata.name` via `unstructured.NestedString` and shadows the outer `name` variable with an identical value:

```go
case *unstructured.Unstructured:
    name, _, _ := unstructured.NestedString(o.Object, "metadata", "name")
    if name == ScrapeTokenSecretName {
        return true
    }
```

The `ScrapeTokenSecretName` check on the outer `name` could be done before the type switch (it's already done for typed `*corev1.Secret` on line 67), and this inner re-read removed.

operator/internal/common/scrape_token_rotation.go
```diff
@@ -0,0 +1,105 @@
...
+
+	mu          sync.RWMutex
+	subscribers []chan event.TypedGenericEvent[client.Object]
+}
+
+// NewTokenRotationBroadcaster returns a broadcaster with the default poll interval.
+func NewTokenRotationBroadcaster() *TokenRotationBroadcaster {
+	return &TokenRotationBroadcaster{
+		Interval: kubernetes.DefaultScrapeTokenMinRequeue,
+	}
```
gbenhaim, line 45: **Concern: broadcaster ticks every 1 minute, but tokens only need refreshing every 30 minutes.**

`DefaultScrapeTokenMinRequeue` is 1 minute. With a 1h TTL and a 50% refresh threshold, `ScrapeTokenNeedsRefresh` only returns true at the 30-minute mark. So each controller gets 29 no-op reconciles (read secret, check freshness, skip) for every 1 useful mint.

With 3 subscriber controllers, that's ~90 no-op reconciles per hour across the operator. Each no-op still reads the CR, checks `IsEnabled()`, resolves the scraper SA, reads the secret, and returns — not expensive individually, but unnecessary.

Consider setting the broadcaster interval closer to the refresh window (e.g., 15 minutes). The freshness check is idempotent, so a slightly longer interval doesn't risk token expiry — it just means the token might be refreshed up to `interval` minutes late, still well before the 1h TTL expires.

gbenhaim (commented):

operator/internal/operatormetrics/scrape_token_rotator.go
yftacherzog: Good point!

operator/internal/common/scrape_token_rotation.go
yftacherzog: As this was raised several times already by multiple humans and humanoids, I'll change the interval.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
yftacherzog: done

operator/internal/common/scrape_token_rotation.go
yftacherzog: done

operator/internal/operatormetrics/scrape_token_rotator.go
yftacherzog: Tickers approach is aligned now

operator/internal/operatormetrics/scrape_token_rotator.go
yftacherzog: done

operator/pkg/kubernetes/component_metrics.go
yftacherzog: removed

operator/internal/common/scrape_token_rotation.go
yftacherzog: done

gbenhaim (approved):
All concerns from the previous review rounds have been addressed:

- Mutex removed from TokenRotationBroadcaster (subscriptions moved to main.go init)
- Reconcilers decoupled from broadcaster (receive-only channel)
- Channels closed on shutdown
- Rotation interval increased to 15 min (DefaultScrapeTokenRotationInterval)
- Timer/Ticker inconsistency resolved (both use Ticker now)
- ScrapeTokenRotator no longer reads the Konflux CR

Test coverage is solid and documentation is accurate.

PR #7655, opened 2026-06-29
test: track Prometheus Operator CRDs

yftacherzog:
We plan to enable monitoring for the Konflux Operator and the underlying services it deploys. For that purpose we need the Prometheus Operator CRDs available in CI so that we can install ServiceMonitors CRs.

Assisted-by: Cursor


.github/scripts/update-third-party-manifests.sh
```diff
@@ -76,6 +84,20 @@ helm template trust-manager jetstack/trust-manager \
...
+PROM_CRD_URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v${PROM_TAG}/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml"
+PROM_ENVTEST_CRD="${REPO_ROOT}/operator/test/crds/prometheus/servicemonitors.monitoring.coreos.com.yaml"
+PROM_DEPS_CRD="${REPO_ROOT}/dependencies/prometheus-operator-crds/servicemonitors.monitoring.coreos.com.yaml"
+mkdir -p "$(dirname "$PROM_ENVTEST_CRD")" "$(dirname "$PROM_DEPS_CRD")"
+{
+  echo "# Source: prometheus-operator ${PROMETHEUS_OPERATOR_VERSION}"
+  echo "# Generated by .github/scripts/update-third-party-manifests.sh — do not edit."
+  curl -fsSL "$PROM_CRD_URL"
+} > "$PROM_ENVTEST_CRD"
+/bin/cp -f "$PROM_ENVTEST_CRD" "$PROM_DEPS_CRD"
```
avi-biton: Minor nit: /bin/cp -f hardcodes the absolute path to cp. While /bin/cp exists on both Linux and macOS, the convention elsewhere in this repo is plain cp. The /bin/cp trick is typically used to bypass interactive shell aliases (e.g. cp -i), which isn't a concern in non-interactive set -euo pipefail scripts.

`cp -f "$PROM_ENVTEST_CRD" "$PROM_DEPS_CRD"`

_Posted by Avi's Cursor_

.github/scripts/update-third-party-manifests.sh
```diff
@@ -76,6 +84,20 @@ helm template trust-manager jetstack/trust-manager \
   --set 'defaultPackage.resources.limits.memory=250Mi' \
   > dependencies/trust-manager/trust-manager.yaml
 
+# prometheus-operator ServiceMonitor CRD (Kind envtest + deploy-deps prerequisite)
+PROM_TAG="${PROMETHEUS_OPERATOR_VERSION#v}"
+PROM_CRD_URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v${PROM_TAG}/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml"
```
avi-biton, line 89: Problem: The CRD URL assumes the prometheus-operator repo structure at example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml. This path is not guaranteed to be stable across major releases. If the repo reorganizes, the curl -fsSL will fail with an opaque exit code, and the error won't clearly say "CRD not found at URL".

Fix: Consider adding an explicit error message on curl failure:

```
if ! curl -fsSL "$PROM_CRD_URL" >> "$PROM_ENVTEST_CRD"; then
  echo "Failed to download ServiceMonitor CRD from $PROM_CRD_URL" >&2
  exit 1
fi
```
Currently the `set -e` will catch it, but the user sees only curl's generic error rather than what URL was tried. This is a minor robustness improvement.

_Posted by Avi's Cursor_

deploy-deps.sh
```diff
@@ -97,6 +97,10 @@ deploy() {
     deploy_cert_manager
     COMPLETED_STEPS+=("$CURRENT_STEP")
 
+    CURRENT_STEP="Prometheus Operator CRDs"
```
avi-biton, line 100: Only cosmetics, but we can add an emoji here like the rest of the steps

avi-biton:
Optional thought: 
the CRD is downloaded once and then `/bin/cp`'d to a second location — the two files are identical 1400-line copies. You could have `dependencies/prometheus-operator-crds/kustomization.yml` reference `../../operator/test/crds/prometheus/servicemonitors.monitoring.coreos.com.yaml` (or use a symlink) to keep a single source of truth.

That said, keeping them separate is consistent with how cert-manager CRDs are handled (operator/test/crds/ vs dependencies/), so the current approach is perfectly fine too. Just flagging in case deduplication appeals to you.

_Posted by Avi's Cursor_

yftacherzog:
> Optional thought: the CRD is downloaded once and then `/bin/cp`'d to a second location — the two files are identical 1400-line copies. You could have `dependencies/prometheus-operator-crds/kustomization.yml` reference `../../operator/test/crds/prometheus/servicemonitors.monitoring.coreos.com.yaml` (or use a symlink) to keep a single source of truth.
> 
> That said, keeping them separate is consistent with how cert-manager CRDs are handled (operator/test/crds/ vs dependencies/), so the current approach is perfectly fine too. Just flagging in case deduplication appeals to you.
> 
> _Posted by Avi's Cursor_

Not sure if there are any caveats for handling this with ref/symlink. We can give it a try on a dedicated PR.

.github/scripts/update-third-party-manifests.sh
yftacherzog: done

.github/scripts/update-third-party-manifests.sh
yftacherzog: done

deploy-deps.sh
yftacherzog: done

avi-biton (approved):

yftacherzog:
> Optional thought: the CRD is downloaded once and then `/bin/cp`'d to a second location — the two files are identical 1400-line copies. You could have `dependencies/prometheus-operator-crds/kustomization.yml` reference `../../operator/test/crds/prometheus/servicemonitors.monitoring.coreos.com.yaml` (or use a symlink) to keep a single source of truth.
> 
> That said, keeping them separate is consistent with how cert-manager CRDs are handled (operator/test/crds/ vs dependencies/), so the current approach is perfectly fine too. Just flagging in case deduplication appeals to you.
> 
> _Posted by Avi's Cursor_

Looking into it again, for cert-manager, those are not direct copies, but one is using yq to extract from the other. It might be that we should use some sort of link here instead of duplicating. I'm looking into that.

avi-biton:
> > Optional thought: the CRD is downloaded once and then `/bin/cp`'d to a second location — the two files are identical 1400-line copies. You could have `dependencies/prometheus-operator-crds/kustomization.yml` reference `../../operator/test/crds/prometheus/servicemonitors.monitoring.coreos.com.yaml` (or use a symlink) to keep a single source of truth.
> > That said, keeping them separate is consistent with how cert-manager CRDs are handled (operator/test/crds/ vs dependencies/), so the current approach is perfectly fine too. Just flagging in case deduplication appeals to you.
> > _Posted by Avi's Cursor_
> 
> Looking into it again, for cert-manager, those are not direct copies, but one is using yq to extract from the other. It might be that we should use some sort of link here instead of duplicating. I'm looking into that.

Since we already have a working pattern, lets try to use it (if possible) instead of re-inventing

gbenhaim (approved):

PR #7669, opened 2026-06-29
test(KFLUXVNGD-1109): add tests for CRD self-healing and drift correction

avi-biton:
Add integration tests for the CRD watch path (Watches(&CRD{},
MapCRDToRequest)) across all three controllers that use it:
- applicationapi (10 CRDs)
- enterprisecontract (1 CRD)
- releaseservice (6 CRDs)
- imagecontroller (2 CRDs)
- integrationservice (3 CRDs)

Self-healing: delete each CRD, verify removal via UID check, then
assert recreation with correct spec.names.kind and ownership labels.

Drift correction: tamper spec.versions[0].served (set to false), then
verify SSA restores it. A ResourceVersion gate ensures the assertion
evaluates post-reconciliation state, preventing false passes from
matching the tampered value.

The served field was chosen because it is present on every CRD, mutable
after creation, and represents real operational damage — disabling it
breaks the API for all consumers. SSA manages it as part of the full
CRD spec, exercising the same restore path as any other spec field.

### DIFF
--- a/operator/internal/controller/applicationapi/konfluxapplicationapi_controller_test.go
+++ b/operator/internal/controller/applicationapi/konfluxapplicationapi_controller_test.go
@@ -21,14 +21,52 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	konfluxv1alpha1 "github.com/konflux-ci/konflux-ci/operator/api/v1alpha1"
 	"github.com/konflux-ci/konflux-ci/operator/internal/condition"
+	"github.com/konflux-ci/konflux-ci/operator/internal/constant"
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/testutil"
+	"github.com/konflux-ci/konflux-ci/operator/pkg/manifests"
 )
+
+type crdInfo struct {
+	name string
+	kind string
+}
+
+var managedCRDs = []crdInfo{
+	{"applications.appstudio.redhat.com", "Application"},
+	{"componentdetectionqueries.appstudio.redhat.com", "ComponentDetectionQuery"},
+	{"components.appstudio.redhat.com", "Component"},
+	{"deploymenttargetclaims.appstudio.redhat.com", "DeploymentTargetClaim"},
+	{"deploymenttargetclasses.appstudio.redhat.com", "DeploymentTargetClass"},
+	{"deploymenttargets.appstudio.redhat.com", "DeploymentTarget"},
+	{"environments.appstudio.redhat.com", "Environment"},
+	{"promotionruns.appstudio.redhat.com", "PromotionRun"},
+	{"snapshotenvironmentbindings.appstudio.redhat.com", "SnapshotEnvironmentBinding"},
+	{"snapshots.appstudio.redhat.com", "Snapshot"},
+}
+
+func crdEntries() []TableEntry {
+	entries := make([]TableEntry, len(managedCRDs))
+	for i, c := range managedCRDs {
+		entries[i] = Entry(c.kind, c.name, c.kind)
+	}
+	return entries
+}
+
+func crdEntriesNameOnly() []TableEntry {
+	entries := make([]TableEntry, len(managedCRDs))
+	for i, c := range managedCRDs {
+		entries[i] = Entry(c.kind, c.name)
+	}
+	return entries
+}
 
 var _ = Describe("KonfluxApplicationAPI Controller", func() {
 	Context("When reconciling a resource", func() {
@@ -37,7 +75,7 @@ var _ = Describe("KonfluxApplicationAPI Controller", func() {
 				ObjectMeta: metav1.ObjectMeta{Name: CRName},
 			}
 			Expect(k8sClient.Create(ctx, appRes)).To(Succeed())
-			testutil.DeferCleanupParentAndChildren(k8sClient, appRes)
+			DeferCleanup(testutil.DeleteAndWait, k8sClient, appRes)
 
 			Eventually(func(g Gomega) {
 				updated := &konfluxv1alpha1.KonfluxApplicationAPI{}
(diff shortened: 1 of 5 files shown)


yftacherzog (approved):

operator/internal/controller/applicationapi/konfluxapplicationapi_controller_test.go
```diff
@@ -21,23 +21,61 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
```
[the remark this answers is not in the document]
avi-biton: Acknowledged - we considered using GetCRDNamesForComponent() for the drift-correction table but opted to keep both tables sourcing from the same hardcoded list for a few reasons:

1. Consistency - self-healing needs Kind which that API doesn't provide, so it must use the hardcoded list. Having drift-correction pull from a different source adds inconsistency without meaningful benefit.
2. Explicit test contracts - hardcoded names act as a documented expectation. If a CRD is accidentally dropped from manifests, the test fails and signals that coverage changed — with dynamic extraction it would silently stop testing that CRD.
3. Low churn - CRD additions/removals are rare and deliberate; the maintenance cost is negligible.

Keeping as-is.

_Posted by Avi's Cursor_

avi-biton:
I updated the scope of this PR to include all the controllers that watches CRDs.
Added tests to all of them.

@yftacherzog @gbenhaim PTAL

operator/internal/controller/applicationapi/konfluxapplicationapi_controller_test.go
```diff
@@ -48,4 +86,94 @@ var _ = Describe("KonfluxApplicationAPI Controller", func() {
...
 	})
+
+	Context("Self-healing", func() {
+		DescribeTable("recreates CRD when deleted",
+			func(ctx context.Context, crdName, expectedKind string) {
+				cr := &konfluxv1alpha1.KonfluxApplicationAPI{
+					ObjectMeta: metav1.ObjectMeta{Name: CRName},
+				}
+				Expect(k8sClient.Create(ctx, cr)).To(Succeed())
+				DeferCleanup(testutil.DeleteAndWait, k8sClient, cr)
```
yftacherzog, line 97: The other CRD-watching controllers pass the CRD under test to `DeferCleanupParentAndChildren` so cluster-scoped CRDs are removed after the parent CR is deleted. Here only the parent is cleaned up; with 10 table entries that can leave CRDs behind in envtest. Consider matching the pattern used in release/integration/imagecontroller, e.g.:

```go
testutil.DeferCleanupParentAndChildren(k8sClient, cr,
    &apiextensionsv1.CustomResourceDefinition{
        ObjectMeta: metav1.ObjectMeta{Name: crdName},
    },
)
```

The same applies to the drift-correction table at line 146.

— Yftach's Cursor

operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller_test.go
```diff
@@ -278,6 +281,48 @@ var _ = Describe("KonfluxEnterpriseContract Controller", Ordered, func() {
...
+			testutil.DeferCleanupParentAndChildren(k8sClient, ec, ecClusterScopedChildren()...)
+
+			crdNN := types.NamespacedName{Name: ecCRDName}
+
+			By("waiting for CRD with owner labels")
+			var originalUID types.UID
+			Eventually(func(g Gomega) {
+				crd := &apiextensionsv1.CustomResourceDefinition{}
+				g.Expect(k8sClient.Get(ctx, crdNN, crd)).To(Succeed())
+				g.Expect(crd.Labels).To(HaveKeyWithValue(constant.KonfluxOwnerLabel, CRName))
```
yftacherzog, line 300: The other controllers assert both `KonfluxOwnerLabel` and `KonfluxComponentLabel` on recreated CRDs. The reconciler sets both via `tracking.OwnershipConfig`, so consider adding:

```go
g.Expect(crd.Labels).To(HaveKeyWithValue(constant.KonfluxComponentLabel, string(manifests.EnterpriseContract)))
```

in the wait and recreation assertions (lines 299 and 323).

— Yftach's Cursor

operator/internal/controller/imagecontroller/konfluximagecontroller_controller_test.go
```diff
@@ -43,6 +46,32 @@ const (
...
+
+var icManagedCRDs = []icCRDInfo{
+	{"imagerepositories.appstudio.redhat.com", "ImageRepository"},
+	{"imagerepositories.konflux-ci.dev", "ImageRepository"},
+}
+
+func icCRDEntries() []TableEntry {
+	entries := make([]TableEntry, len(icManagedCRDs))
+	for i, c := range icManagedCRDs {
+		entries[i] = Entry(c.name, c.name, c.kind)
```
yftacherzog, line 62: Minor consistency nit: the other controllers use `Entry(c.kind, c.name, c.kind)` for the Ginkgo description. Using `c.name` here works but produces less readable output in test reports.

— Yftach's Cursor

yftacherzog (commented):
A few line-level notes below. The PR description still says "three controllers" but the change now covers five — worth updating for accuracy.

— Yftach's Cursor

operator/internal/controller/applicationapi/konfluxapplicationapi_controller_test.go
avi-biton: Switched all CRD tests to `DeferCleanup(testutil.DeleteAndWait, k8sClient, cr)` - cleaning up only the parent CR.

`DeferCleanupParentAndChildren` exists to prevent Ginkgo's LIFO ordering from deleting cluster-scoped children while the reconciler is still active. 
In the CRD tests we register a single cleanup call and each table entry tests a different CRD name, so there's no LIFO risk and no interference between entries. 
Deleting the parent stops the reconciler, and leftover CRDs are harmless — they're keyed by name (no accumulation) and get reconciled fresh by the next entry.

operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller_test.go
avi-biton: Added

operator/internal/controller/imagecontroller/konfluximagecontroller_controller_test.go
avi-biton: Kept `c.name` (FQDN) here intentionally - `imagecontroller` has two CRDs with the same Kind (`ImageRepository`) but different API groups (`appstudio.redhat.com` vs `konflux-ci.dev`), so using `c.kind` would produce duplicate Entry descriptions. 
The FQDN keeps them unambiguous in test output.

avi-biton:
Looking at the failure

operator/internal/controller/integrationservice/konfluxintegrationservice_controller_test.go
```diff
@@ -49,6 +52,33 @@ const (
 	managerClusterRoleBindingName = "integration-service-manager-rolebinding"
 )
 
+type isCRDInfo struct {
```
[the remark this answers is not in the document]
avi-biton: is stands for IntegrationService.
No need to take action.

PR #7718, opened 2026-06-30
feat(KFLUXVNGD-1108): install EnterpriseContractPolicy CRD and CR as part of operator lifecycle

avi-biton:
The EnterpriseContractPolicy (ECP) CRD is shipped by the enterprise-contract upstream but is needed before the KonfluxEnterpriseContract controller can reconcile its default CR instance. This creates a chicken-and-egg problem: the controller cannot start an Owns() watch on a GVK whose CRD does not yet exist at manager startup time.

Solution:
- Add a `preInstallCRDs()` function in `cmd/main.go` that runs before `mgr.Start()`. It creates a direct `client.Client` and applies the ECP CRD via Server-Side Apply (field owner: "konflux-operator-bootstrap") so the CRD is registered before any controller sets up watches.
- Add `Owns()` on unstructured ECP CRs in the `KonfluxEnterpriseContract` controller with `IgnoreStatusUpdatesPredicate`, enabling self-healing: if the default ECP CR is deleted or its spec is tampered with, the controller is immediately re-enqueued and restores the desired state.

Supporting changes:
- rebuild-upstream-manifests.sh: extract the ECP CRD from the enterprise- contract manifests.yaml into `operator/test/crds/enterprise-contract/` so envtest can register the GVK during unit tests.
- `testutil.go`: add the ECP CRD path to `CRDDirectoryPaths` for envtest.
- `verify-manifests-in-sync.sh` + `path-filters`: add CI verification that the extracted test CRD stays in sync with the upstream manifests.
- Controller tests: add drift-correction test cases verifying that a deleted ECP is recreated and a modified ECP spec is restored (skipPolicies=false), and that no recreation occurs when skipPolicies=true.
- operator-test-e2e.yaml: set INSTALL_PROMETHEUS_OPERATOR_CRDS=true.

Note: the ECP CRD is not installed by `make install` (which only covers the operator's own Konflux* CRDs from config/crd/). Instead, the operator itself installs it at startup via preInstallCRDs(), before the manager starts. This ensures the same behavior in both development (make install + make run) and production (OLM) — the operator binary is the single owner of the ECP CRD lifecycle, regardless of how it is deployed.

Tested on a local Kind cluster (deploy-local.sh, OPERATOR_INSTALL_METHOD=none):
- [x] Operator startup pre-installs the ECP CRD before manager start
- [x] All controllers start, including Owns() watch on unstructured ECP
- [x] Applying Konflux CR creates the default ECP in enterprise-contract-service
- [x] Deleting the default ECP triggers reconcile and immediate recreation
- [x] Patching the ECP spec (injecting drift) triggers reconcile and SSA restore
- [x] Setting skipPolicies=true on Konflux CR deletes the default ECP via orphan cleanup
- [x] User-created ECP with a different name is not touched by the operator
- [x] Restoring skipPolicies=false recreates the default ECP, user policies unaffected
- [x] User-created ECP with the same name ("default") is overwritten by SSA when skipPolicies=false (expected: operator owns that name)

### DIFF
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -34,6 +34,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
@@ -63,6 +64,7 @@ import (
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/segmentbridge"
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/ui"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/clusterinfo"
+	"github.com/konflux-ci/konflux-ci/operator/pkg/kubernetes"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/manifests"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/segment"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/version"
@@ -273,6 +275,18 @@ func main() {
 		setupLog.Error(err, "unable to parse embedded manifests")
 		os.Exit(1)
 	}
+	// Pre-install CRDs managed by controllers that also watch CRs of those CRDs.
+	// The CRD must exist before mgr.Start() so the Owns() informer can sync.
+	directClient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		setupLog.Error(err, "unable to create client for CRD pre-install")
+		os.Exit(1)
+	}
+	if err := preInstallCRDs(context.Background(), directClient, objectStore); err != nil {
+		setupLog.Error(err, "unable to pre-install CRDs")
+		os.Exit(1)
+	}
+
 	// Detect cluster information (platform, version, capabilities)
 	clusterInfo, err := clusterinfo.Detect(cfg)
 	if err != nil {
@@ -465,4 +479,48 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+}
+
+// preInstallCRDs applies CRDs from embedded manifests before the manager starts.
+//
+// Why this is needed (and why Watches(CRD) + reconcile is insufficient):
+// Controllers that use Owns() to watch CRs of upstream CRDs (e.g., EnterpriseContractPolicy)
+// require the CRD to be registered in the API server at manager startup time. Without it,
+// controller-runtime cannot create the informer for the owned GVK, and mgr.Start() blocks
+// indefinitely waiting for the cache to sync. The existing Watches(CRD) pattern only handles
+// *re-applying* a CRD after out-of-band deletion — it cannot bootstrap a CRD that was never
+// present. This function solves the ordering problem by ensuring CRDs exist before any
+// controller sets up watches, using a direct client (bypassing the not-yet-started cache).
+// The controller then takes over full ownership of the CRD during its first reconcile via SSA.
+//
+// The components list below is intentionally limited to those whose controllers use Owns()
+// on CRs of CRDs they install. Components that only Watches(CRD) do not need pre-install.
+func preInstallCRDs(ctx context.Context, c client.Client, store *manifests.ObjectStore) error {
+	components := []manifests.Component{
+		manifests.EnterpriseContract,
+	}
+	for _, comp := range components {
+		objects, err := store.GetForComponent(comp)
+		if err != nil {
+			return fmt.Errorf("getting objects for %s: %w", comp, err)
+		}
+		for _, obj := range objects {
+			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
+			if !ok {
+				continue
+			}
+			if err := c.Patch(ctx, crd, kubernetes.SSAPatch,
+				client.FieldOwner("konflux-operator-bootstrap"),
+				client.ForceOwnership,
+			); err != nil {
+				return fmt.Errorf("pre-installing CRD %s: %w", crd.Name, err)
+			}
+			setupLog.Info("Pre-installed CRD",
+				"name", crd.Name,
+				"resourceVersion", crd.ResourceVersion,
+				"versions", len(crd.Spec.Versions),
+			)
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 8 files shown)


operator/cmd/main.go
```diff
@@ -273,6 +275,18 @@ func main() {
 		setupLog.Error(err, "unable to parse embedded manifests")
```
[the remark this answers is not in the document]
avi-biton: Done

operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller.go
```diff
@@ -168,13 +169,17 @@ func (r *KonfluxEnterpriseContractReconciler) SetupWithManager(mgr ctrl.Manager)
 	if err != nil {
 		return err
 	}
+	ecPolicy := &unstructured.Unstructured{}
+	ecPolicy.SetGroupVersionKind(ecPolicyGVK)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&konfluxv1alpha1.KonfluxEnterpriseContract{}).
 		Named("konfluxenterprisecontract").
 		Owns(&corev1.Namespace{}, builder.WithPredicates(predicate.IgnoreStatusUpdatesPredicate)).
```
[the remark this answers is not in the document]
avi-biton: Done

operator/cmd/main.go
```diff
@@ -466,3 +480,36 @@ func main() {
 		os.Exit(1)
 	}
 }
+
```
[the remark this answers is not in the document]
avi-biton: Done

operator/cmd/main.go
```diff
@@ -466,3 +480,47 @@ func main() {
...
+			return fmt.Errorf("getting objects for %s: %w", comp, err)
+		}
+		for _, obj := range objects {
+			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
+			if !ok {
+				continue
+			}
+			if err := c.Patch(ctx, crd, kubernetes.SSAPatch,
+				client.FieldOwner("konflux-operator-bootstrap"),
+				client.ForceOwnership,
```
yftacherzog, line 514: Do we want this to be owned by the konflux-operator-bootstrap or by the EC reconciler?

operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller_test.go
```diff
@@ -256,6 +257,68 @@ var _ = Describe("KonfluxEnterpriseContract Controller", Ordered, func() {
...
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, ecpNN, newDefaultECPolicy())).
+					To(MatchError(ContainSubstring("not found")))
+			}).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Succeed())
+
+			By("verifying the EnterpriseContractPolicy stays deleted after Owns() reconcile settles")
+			Consistently(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, ecpNN, newDefaultECPolicy())).
+					To(MatchError(ContainSubstring("not found")))
+			}).WithTimeout(5 * time.Second).WithPolling(testutil.EventuallyPolling).Should(Succeed())
```
yftacherzog, line 319: I think this would be more consistent with how we handle this elsewhere:
```suggestion
            By("waiting for the EnterpriseContractPolicy to be removed via orphan cleanup")
            Eventually(func(g Gomega) {
                err := k8sClient.Get(ctx, ecpNN, newDefaultECPolicy())
                g.Expect(errors.IsNotFound(err)).To(BeTrue(), "unexpected error: %v", err)
            }).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Succeed())
            By("verifying the EnterpriseContractPolicy stays deleted after Owns() reconcile settles")
            Consistently(func(g Gomega) {
                err := k8sClient.Get(ctx, ecpNN, newDefaultECPolicy())
                g.Expect(errors.IsNotFound(err)).To(BeTrue(), "unexpected error: %v", err)
            }).WithTimeout(5 * time.Second).WithPolling(testutil.EventuallyPolling).Should(Succeed())
```

operator/cmd/main.go
avi-biton: Good question. The `konflux-operator-bootstrap` field owner is intentional for two reasons:

Observability: If you inspect the CRD between startup and first reconcile (`kubectl get crd -o yaml`), you can clearly see it was applied by the bootstrap step vs. a regular reconcile. After the first reconcile, the EC controller re-applies the CRD with `konflux-enterprisecontract-controller` + `ForceOwnership`, taking over all managed fields.

Generic bootstrap function: `preInstallCRDs()` is designed to handle multiple components (today just `EnterpriseContract`, but `ReleaseServiceConfig` is next per KFLUXVNGD-892). Using a single generic field owner keeps the function decoupled from individual controller identities so no need for a per-component field manager mapping.

Functionally both approaches are equivalent (SSA with `ForceOwnership` means the reconciler takes over regardless), but the separate name keeps concerns cleanly separated.

_Posted by Avi's Cursor_

operator/internal/controller/enterprisecontract/konfluxenterprisecontract_controller_test.go
avi-biton: Done

yftacherzog (approved):

gbenhaim:
I'm thinking of it's the right approach to install the cards using the operator or to place them under config/crds and let them be installed alongside the operator's cards. At least this is the patterns I know other operators are following.

avi-biton:
> I'm thinking of it's the right approach to install the cards using the operator or to place them under config/crds and let them be installed alongside the operator's cards. At least this is the patterns I know other operators are following.

Thanks for the review! The pattern you're describing (CRDs under `config/crd/`) works well for operators that define their own CRDs via Go API types - and that's exactly what we do for the 16 Konflux* CRDs in `config/crd/bases/`, which are generated by `controller-gen` from our types in `api/v1alpha1/`.

The ECP CRD is a different case though - it's an upstream CRD (`enterprisecontractpolicies.appstudio.redhat.com`) that the operator ships and manages as part of deploying the `enterprise-contract component`. 
The reconciler already applies it via SSA during every reconciliation (line 158 of the EC controller's `applyManifests()`), sourcing it from the embedded manifests in `pkg/manifests/enterprise-contract/manifests.yaml`.

`preInstallCRDs()` exists only to solve the startup ordering problem: `Owns()` requires the CRD to be registered at `mgr.Start()` time, but the reconciler hasn't run yet. It reads from the same embedded manifests - no duplication.

Placing the upstream CRD in `config/crd/` would mean:

Two copies of the same CRD - one in `config/crd/` (for make install) and one in the embedded manifests (for reconciliation). They'd need to stay in sync on every upstream bump.
Convention break - `config/crd/bases/` files are all controller-gen output. 
Mixing in manually-maintained upstream CRDs is confusing (and make manifests won't update them).
OLM bundle conflict - `config/manifests/kustomization.yaml` includes `config/default` which includes `config/crd`. The upstream CRD would end up in our OLM bundle as "owned", conflicting with the enterprise-contract operator's own bundle.
Still doesn't cover all flows - For `make deploy` and `kubectl apply -f install.yaml`, CRDs and the Deployment are applied in one shot. There's no guarantee the CRD is Established before the operator pod starts its `Owns()` watch. `preInstallCRDs()` eliminates this race.

The pattern we are using here is in use in several operators that needs the CRDs before owning them, such as:
- https://github.com/tigera/operator/blob/release-v1.34/main.go#L291-L295
- https://github.com/stolostron/backplane-operator/blob/main/main.go#L346-L349

Happy to discuss further if you see a gap in this reasoning!

_Posted by Avi's Cursor_

PR #7932, opened 2026-07-07
test(conformance): reorder tests and validate retrigger spec without waiting

avi-biton:
Reorder the conformance test sequence so Integration Testing and Release run to completion before the Build Retrigger is triggered. Previously, the retrigger ran concurrently with integration, producing overlapping PipelineRuns that overwhelmed etcd on resource-constrained Kind clusters (4 vCPU GitHub Actions runners).

The retrigger test no longer waits for the full PipelineRun to complete. Instead it validates the wiring is correct by asserting:
- Labels (event-type, pipeline type, component, application)
- Source commit SHA label is set
- PipelineRef is resolved (bundle resolver wiring)
- Pipeline parameters (output-image, git-url)
- Build request annotation is cleared by build-service

This eliminates ~5-10 minutes of runtime where a second build PipelineRun would run in parallel, reducing peak etcd object count and API server load. Namespace cleanup in AfterAll tears down any in-flight runs.

### DIFF
--- a/test/go-tests/tests/conformance/conformance.go
+++ b/test/go-tests/tests/conformance/conformance.go
@@ -306,30 +306,6 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				})
 			})
 
-			// --- Build Retrigger ---
-
-			ginkgo.When("push pipelinerun is retriggered", func() {
-				ginkgo.It("should eventually succeed", func() {
-					err = fw.AsKubeAdmin.HasController.SetComponentAnnotation(component.GetName(), buildcontrollers.BuildRequestAnnotationName, buildcontrollers.BuildRequestTriggerPaCBuildAnnotationValue, userNamespace)
-					gomega.Expect(err).NotTo(gomega.HaveOccurred())
-
-					gomega.Eventually(func() error {
-						testPipelinerun, err = fw.AsKubeAdmin.HasController.GetComponentPipelineRunWithType(component.GetName(), appSpec.ApplicationName, userNamespace, "build", "", "incoming")
-						if err != nil {
-							dumpDiagnostics(fw.AsKubeAdmin, component.GetName(), appSpec.ApplicationName, userNamespace)
-							return err
-						}
-						if !testPipelinerun.HasStarted() {
-							return fmt.Errorf("pipelinerun %s/%s hasn't started yet", testPipelinerun.GetNamespace(), testPipelinerun.GetName())
-						}
-						return nil
-					}, 10*time.Minute, constants.PipelineRunPollingInterval).Should(gomega.Succeed(), "timed out waiting for retriggered PipelineRun")
-
-					err = fw.AsKubeAdmin.HasController.WaitForComponentPipelineToBeFinished(component, "build", "", "incoming", fw.AsKubeAdmin.TektonController, &has.RetryOptions{Retries: 2, Always: true}, testPipelinerun)
-					gomega.Expect(err).NotTo(gomega.HaveOccurred())
-				})
-			})
-
 			// --- Integration Testing ---
 
 			ginkgo.When("Integration Test PipelineRun is created", ginkgo.Label(upstreamKonfluxTestLabel), func() {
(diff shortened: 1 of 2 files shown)


test/go-tests/tests/conformance/conformance.go
```diff
@@ -432,6 +408,76 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 					}, customResourceUpdateTimeout, defaultPollingInterval).Should(gomega.Succeed(), "release %q not marked as released", release.Name)
 				})
 			})
+
+			// --- Build Retrigger ---
+			// Validates the retrigger wiring without waiting for the PipelineRun to
+			// complete — namespace cleanup in AfterAll tears down any in-flight runs.
```
yftacherzog: AfterAll only deletes `managedNamespace` and Git refs/webhooks; it does not delete `userNamespace`, where the retrigger PipelineRun lives. In CI the Kind cluster is torn down after the job, so impact is limited. Consider rewording this comment to avoid implying user-namespace PipelineRuns are explicitly cleaned up.

— Yftach's Cursor

test/go-tests/tests/conformance/conformance.go
```diff
@@ -432,6 +408,76 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+				})
+
+				ginkgo.It("should have a source commit SHA label", func() {
+					// Confirms the retrigger picked up the correct git revision.
+					gomega.Expect(testPipelinerun.Labels["pipelinesascode.tekton.dev/sha"]).NotTo(gomega.BeEmpty(),
+						"retriggered PipelineRun should carry a source commit SHA label")
+				})
+
+				ginkgo.It("should have a resolved pipeline bundle", func() {
+					// Confirms the Tekton bundle resolver wiring is correct — a nil
```
yftacherzog: A non-nil `PipelineRef` indicates PaC populated the spec, not that bundle/task resolution completed. The recent `CouldntGetTask` flake happened while resolving tasks after the PLR existed. Consider narrowing the comment (and possibly the `It` name) to "pipeline ref is set" rather than "resolved pipeline bundle".

— Yftach's Cursor

test/go-tests/tests/conformance/conformance.go
```diff
@@ -432,6 +408,76 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+					// PipelineRef would indicate a misconfigured PaC repository.
+					gomega.Expect(testPipelinerun.Spec.PipelineRef).NotTo(gomega.BeNil(),
+						"retriggered PipelineRun should have a PipelineRef (bundle resolution)")
+				})
+
+				ginkgo.It("should have correct pipeline parameters", func() {
```
yftacherzog, line 457: The loop only asserts `output-image` and `git-url` when those names appear in `Spec.Params`. If either param is missing, the test passes without failing. Consider collecting both values (or using explicit `Expect` after the loop) and requiring each param to be present.

— Yftach's Cursor

test/go-tests/tests/conformance/conformance.go
```diff
@@ -432,6 +408,76 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 					}, customResourceUpdateTimeout, defaultPollingInterval).Should(gomega.Succeed(), "release %q not marked as released", release.Name)
 				})
 			})
+
+			// --- Build Retrigger ---
```
yftacherzog, line 412: `test/go-tests/tests/conformance/README.md` still describes the old flow (integration/release only) and does not mention retrigger or the new order. A short bullet under "Test scenario" would keep the doc aligned (e.g. retrigger wiring validated after release, without waiting for completion).

— Yftach's Cursor

yftacherzog (commented):
Optional nits on the retrigger assertions and docs — none are blockers.

— Yftach's Cursor

yftacherzog:
Looking at CI, integration + release take much less than retrigger, so we should gain speedup here I think.
It fails at the moment on a wrong assertion. I think it's just a matter of some tweaking to have it passing.

gbenhaim:
There was another PR for tuning Etcd. Do we still need this PR?

gbenhaim:
/ask does reordering the tests (introduced in this PR) going to make the conformance tests longer?

yftacherzog:
> /ask does reordering the tests (introduced in this PR) going to make the conformance tests longer?

@gbenhaim It's not just reordering so it's going to make them shorter.

Currently we have the conformance run duration dominated by the duration of the 2 on-push pipeline, having the integration and release tests running in parallel to the second run and finish before the second run is done.

The parallel running is what we believe stresses the runner.

With this change, we:
1. Don't let them run in parallel.
2. Don't wait for the second on-push pipeline to finish, and instead just verify that we're able to trigger it.

test/go-tests/tests/conformance/conformance.go
```diff
@@ -432,6 +408,76 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+						"retriggered PipelineRun should carry a source commit SHA label")
+				})
+
+				ginkgo.It("should have a resolved pipeline bundle", func() {
+					// Confirms the Tekton bundle resolver wiring is correct — a nil
+					// PipelineRef would indicate a misconfigured PaC repository.
+					gomega.Expect(testPipelinerun.Spec.PipelineRef).NotTo(gomega.BeNil(),
+						"retriggered PipelineRun should have a PipelineRef (bundle resolution)")
+				})
+
```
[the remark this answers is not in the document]
avi-biton: fixed

test/go-tests/tests/conformance/conformance.go
avi-biton: Fixed

test/go-tests/tests/conformance/conformance.go
avi-biton: Fixed

test/go-tests/tests/conformance/conformance.go
avi-biton: Fixed

test/go-tests/tests/conformance/conformance.go
avi-biton: fixed

avi-biton:
> Looking at CI, integration + release take much less than retrigger, so we should gain speedup here I think. It fails at the moment on a wrong assertion. I think it's just a matter of some tweaking to have it passing.

Done

yftacherzog (approved):

PR #8024, opened 2026-07-09
test(metrics): add OpenShift UWM e2e and fix operand scrape ordering

yftacherzog:
Add optional OpenShift e2e coverage that verifies Konflux operand
metrics are scraped by user-workload monitoring (UWM Prometheus):
UWM readiness, scrape contract (ServiceMonitor, prometheus-scrape-token,
metrics-scraper CRB), and up==1 per catalog target.

Fix intermittent UWM flakes where prometheus-operator rejects operand
ServiceMonitors when bearerTokenSecret is missing at evaluation time:
- Phase 1: annotation resync nudges (token-minted/refreshed,
  settle-retry, secret-sync) so PO re-evaluates the SM after the scrape
  token exists
- Phase 2: defer operand ServiceMonitor apply until
  prometheus-scrape-token is readable; re-apply every reconcile for
  tracking-client orphan safety

Wire the suite into OCP optional e2e (enable-uwm.sh, run-e2e.sh), emit
[UWM resync] CI evidence (sm_after_secret, active_targets), and document
operator behavior in component-monitoring.md.

Assisted-by: Cursor

### DIFF
--- a/operator/internal/common/operand_servicemonitor.go
+++ b/operator/internal/common/operand_servicemonitor.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2025 Konflux CI.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/konflux-ci/konflux-ci/operator/pkg/kubernetes"
+)
+
+// OperandServiceMonitorFromObjects returns the embedded operand ServiceMonitor matching
+// namespace and name from a component manifest object list. Used by ApplyServiceMonitor
+// callbacks in deferred ServiceMonitor apply (SM is skipped in applyManifests, applied here).
+func OperandServiceMonitorFromObjects(objects []client.Object, namespace, name string) (client.Object, bool) {
+	for _, obj := range objects {
+		if !kubernetes.IsComponentMetricsServiceMonitor(obj) {
+			continue
+		}
+		if obj.GetNamespace() == namespace && obj.GetName() == name {
+			return obj, true
+		}
+	}
+	return nil, false
+}
--- a/operator/internal/common/scrape_token.go
+++ b/operator/internal/common/scrape_token.go
@@ -20,36 +20,77 @@ import (
 	"context"
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/konflux-ci/konflux-ci/operator/pkg/kubernetes"
 )
 
+var operandServiceMonitorGVK = schema.GroupVersionKind{
+	Group:   "monitoring.coreos.com",
+	Version: "v1",
+	Kind:    "ServiceMonitor",
+}
+
 // ScrapeTokenReconcilerConfig configures operand-namespace Prometheus scrape token reconciliation.
 type ScrapeTokenReconcilerConfig struct {
-	Client           client.Reader
-	Clock            clock.Clock
-	TokenCreator     kubernetes.TokenCreator
-	Scraper          types.NamespacedName
-	OperandNamespace string
-	Apply            kubernetes.ScrapeTokenApplyFunc
+	Client             client.Client
+	Clock              clock.Clock
+	TokenCreator       kubernetes.TokenCreator
+	Scraper            types.NamespacedName
+	OperandNamespace   string
+	ServiceMonitorName string
+	Apply              kubernetes.ScrapeTokenApplyFunc
+	// ApplyServiceMonitor applies the operand ServiceMonitor after prometheus-scrape-token
+	// is readable (deferred ServiceMonitor apply). Must be idempotent and invoked on every
+	// reconcile so the tracking client retains the SM during orphan cleanup; callers
+	// typically use tc.ApplyOwned.
+	ApplyServiceMonitor func(ctx context.Context) error
+}
+
+// DeferredSMApplyResult captures operand ServiceMonitor state from deferred apply without
+// re-reading the informer cache immediately after the write.
+type DeferredSMApplyResult struct {
+	ExistedBeforeApply bool
+	Prior              *unstructured.Unstructured
 }
 
 // ReconcilePrometheusScrapeToken ensures the operand scrape token Secret exists and is fresh.
-// Periodic re-checks are driven by TokenRotationBroadcaster rather than RequeueAfter.
-func ReconcilePrometheusScrapeToken(ctx context.Context, cfg ScrapeTokenReconcilerConfig) error {
+// When ServiceMonitorName is set, it also applies the operand ServiceMonitor after the token
+// is readable (deferred ServiceMonitor apply) and nudges prometheus-operator to re-evaluate
+// the SM (resync nudges).
+//
+// Deferred ServiceMonitor apply: operand reconcilers skip the ServiceMonitor in applyManifests
+// when componentMetrics is enabled. This function applies it here via ApplyServiceMonitor
+// only after the scrape token Secret exists with non-empty token bytes. ApplyServiceMonitor
+// must run every reconcile (not only on create) so tracking-client orphan cleanup does not
+// delete the SM.
+//
+// Resync nudges: patches SM annotations via ResyncOperandServiceMonitor when the token is
+// minted (token-minted) or refreshed (token-refreshed), on settle requeue after
+// DefaultServiceMonitorResyncSettleDelay (settle-retry), or when secret resourceVersion
+// drifts (secret-sync). secret-sync is suppressed while metrics-scrape-resync-settle=pending.
+//
+// Returns RequeueAfter on mint/refresh so settle-retry runs once more before steady state.
+func ReconcilePrometheusScrapeToken(ctx context.Context, cfg ScrapeTokenReconcilerConfig) (reconcile.Result, error) {
 	if cfg.TokenCreator == nil {
-		return fmt.Errorf("token creator is required")
+		return reconcile.Result{}, fmt.Errorf("token creator is required")
 	}
 	if cfg.OperandNamespace == "" {
-		return fmt.Errorf("operand namespace is required")
+		return reconcile.Result{}, fmt.Errorf("operand namespace is required")
 	}
 	if cfg.Scraper.Namespace == "" || cfg.Scraper.Name == "" {
-		return fmt.Errorf("scraper service account is required")
+		return reconcile.Result{}, fmt.Errorf("scraper service account is required")
 	}
-	_, err := kubernetes.EnsurePrometheusScrapeToken(ctx, kubernetes.EnsureScrapeTokenInput{
+
+	tokenResult, err := kubernetes.EnsurePrometheusScrapeToken(ctx, kubernetes.EnsureScrapeTokenInput{
 		Client:           cfg.Client,
 		Clock:            cfg.Clock,
 		TokenCreator:     cfg.TokenCreator,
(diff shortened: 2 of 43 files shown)


test/go-tests/pkg/metricsauth/portforward.go
```diff
@@ -68,8 +58,35 @@ func startServicePortForward(ctx context.Context, cfg *rest.Config, svc ServiceR
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	podName, err := readyPodFromService(ctx, clientset, svc.Namespace, svc.Name, svc.Port)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return startPodPortForward(ctx, cfg, svc.Namespace, podName, svc.Port)
```
avi-biton: instead of directly returning the result of `startPodPortForward`, capture the error and wrap it with the service namespace/name. This way, when `startPodPortForward` fails (e.g. "port-forward to build-service pod xyz not ready: context deadline exceeded"), the caller sees the service context too: "service build-service/build-service-controller-manager-metrics-service: port-forward to build-service pod xyz not ready: ...".

```suggestion
		pf, port, err := startPodPortForward(ctx, cfg, svc.Namespace, podName, svc.Port)
	if err != nil {
		return nil, 0, fmt.Errorf("service %s/%s: %w", svc.Namespace, svc.Name, err)
	}
	return pf, port, nil
```

_Posted by Avi's Cursor_

yftacherzog: done

scripts/operator-e2e/openshift/enable-uwm.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+# Enable OpenShift user-workload monitoring (UWM) when not already configured.
+#
+# Idempotent: safe on clusters where preview/Argo already applied
+# cluster-monitoring-config (infra-deployments monitoring-workload-prometheus).
```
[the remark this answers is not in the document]
yftacherzog: done

test/go-tests/pkg/metricsopenshift/names_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/konflux-ci/konflux-ci/test/go-tests/pkg/metricsauth"
+)
+
+func TestPrometheusBindingName(t *testing.T) {
+	operator := metricsauth.Target{ID: "konflux-operator"}
+	assert.Equal(t, "konflux-operator-prometheus-konflux-operator-metrics-reader", PrometheusBindingName(operator))
+	assert.Equal(t, "prometheus-build-service-metrics-reader", PrometheusBindingName(metricsauth.Target{ID: "build-service"}))
```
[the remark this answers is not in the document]
yftacherzog: following the pattern in sibling files

operator/internal/common/scrape_token.go
```diff
@@ -19,43 +19,234 @@ package common
...
+		Namespace: cfg.OperandNamespace,
+	}, secret); err != nil {
+		return reconcile.Result{}, fmt.Errorf("get scrape token secret after ensure: %w", err)
+	}
+	if len(secret.Data[kubernetes.ScrapeTokenSecretKey]) == 0 {
+		return reconcile.Result{}, fmt.Errorf(
+			"scrape token secret %s/%s has empty %q after ensure",
+			cfg.OperandNamespace, kubernetes.ScrapeTokenSecretName, kubernetes.ScrapeTokenSecretKey,
+		)
+	}
```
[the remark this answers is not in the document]
yftacherzog: For now this is not an issue. We can address it if we see flakiness around this.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -234,6 +254,16 @@ func (r *KonfluxBuildServiceReconciler) applyManifests(ctx context.Context, tc *
...
+		// Phase 2 (UWM): defer operand ServiceMonitor until ReconcilePrometheusScrapeToken
+		// applies it after prometheus-scrape-token is readable.
+		if metricsEnabled && kubernetes.IsComponentMetricsServiceMonitor(obj) {
+			log.V(1).Info("Deferring operand ServiceMonitor apply until scrape token is ready",
+				"kind", tracking.GetKind(obj),
+				"name", obj.GetName(),
+				"namespace", obj.GetNamespace(),
+			)
+			continue
+		}
```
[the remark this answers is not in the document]
yftacherzog: done

operator/internal/operatormetrics/scrape_token_rotator.go
```diff
@@ -122,6 +122,7 @@ func (r *ScrapeTokenRotator) reconcile(ctx context.Context, namespace string) (t
 			return kubernetes.ApplyScrapeTokenSecret(applyCtx, r.Client, secret)
 		},
 	})
```
avi-biton, line 124: ### Operator SM-before-token ordering

The operand fix (defer SM until token is readable + resync nudges) looks correct — but `ScrapeTokenRotator.reconcile` still does the opposite order:

```go
// scrape_token_rotator.go:111-124
EnsureOperatorServiceMonitor(...)       // creates SM referencing prometheus-scrape-token
EnsurePrometheusScrapeToken(...)        // mints the Secret
```
On cold start (empty namespace, first leader reconcile) PO can evaluate the SM before the Secret exists — same rejection class this PR fixes for operands. No resync nudge either.

The UWM suite still asserts up==1 for konflux-operator (UWMCatalogTargets includes it), so this can surface as a flaky optional OCP job.

Would you prefer to swap the ordering here (+ optional resync) in this PR, or track it as a follow-up?

_Posted by Avi's Cursor_

scripts/operator-e2e/openshift/run-metrics-openshift-tests.sh
```diff
@@ -0,0 +1,20 @@
...
+# Intended for OpenShift CI only (via test/e2e/run-e2e.sh or infra overlay e2e). Not invoked from
+# Kind/Tekton operator E2E.
+set -euo pipefail
+
+REPO_ROOT="$(cd "${1:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)}" && pwd)"
+LABEL_FILTER="${METRICS_OPENSHIFT_GINKGO_LABEL_FILTER:-openshift}"
+
+cd "${REPO_ROOT}/test/go-tests"
+echo "Running OpenShift metrics tests (label filter: ${LABEL_FILTER})..."
+go test -mod=mod ./metricsopenshift -v -timeout 45m -ginkgo.label-filter="${LABEL_FILTER}"
```
avi-biton: ### Test timeout budget

`run-metrics-openshift-tests.sh` runs with `-timeout 45m`, but the serial worst case adds up:

- `WaitReady`: up to 3 × 600s (config check + pod ready + canary) = ~30m
- 3 contract specs: 3 × 5m = 15m
- 3 UWM target specs: 3 × 10m = 30m
- 2 up-only specs: 2 × 10m = 20m

If multiple specs hit their max poll windows (slow convergence, PO restart, etc.) the `go test` timeout can kill the suite before Ginkgo reports which spec was slow.

Consider raising `-timeout` (e.g. 90m) or shortening the per-spec ceilings — the `go test` timeout should be a safety net, not the primary bound.

_Posted by Avi's Cursor_

operator/pkg/kubernetes/component_metrics.go
```diff
@@ -36,6 +36,18 @@ var ComponentMetricsOrphanCleanupGVKs = []schema.GroupVersionKind{
...
+// from upstream-kustomizations/*/monitoring/. Operand reconcilers skip this object in
+// applyManifests when componentMetrics is enabled (deferred ServiceMonitor apply); it is
+// applied later from ReconcilePrometheusScrapeToken after prometheus-scrape-token is readable.
+func IsComponentMetricsServiceMonitor(obj client.Object) bool {
+	if obj == nil {
+		return false
+	}
+	gvk := objectGroupVersionKind(obj)
+	return gvk.Group == "monitoring.coreos.com" && gvk.Kind == "ServiceMonitor"
+}
```
avi-biton, line 49: ### IsComponentMetricsServiceMonitor matches any ServiceMonitor

`IsComponentMetricsServiceMonitor` returns true for any `monitoring.coreos.com/v1 ServiceMonitor`, not just the operand metrics SM:

```go
func IsComponentMetricsServiceMonitor(obj client.Object) bool {
	// ...
	return gvk.Group == "monitoring.coreos.com" && gvk.Kind == "ServiceMonitor"
}
```

Both applyManifests loops (build-service, image-controller) use this to decide which SM to defer. The ApplyServiceMonitor callback then applies only the named SM (e.g. "build-service"). If a second ServiceMonitor were added to a component's manifests, it would be deferred by applyManifests but never applied by the callback — silently dropped.

Safe today (one SM per component), but fragile. Consider also matching by name or adding a manifest label/annotation to mark the operand metrics SM.

_Posted by Avi's Cursor_

test/go-tests/pkg/metricsopenshift/wait.go
```diff
@@ -0,0 +1,241 @@
+// Package metricsopenshift provides helpers for OpenShift user-workload monitoring (UWM)
+// metrics integration tests used by test/go-tests/metricsopenshift.
+//
+// Operator scrape wiring is documented in operator/docs/component-monitoring.md (deferred
+// ServiceMonitor apply, resync nudges). This package implements the test-side waits,
+// test-side waits, contract checks, Prometheus queries, and CI log evidence.
```
avi-biton: Duplication
```suggestion
// contract checks, Prometheus queries, and CI log evidence.
```

operator/internal/operatormetrics/scrape_token_rotator.go
yftacherzog: The operator path has the same theoretical race class, but we have not seen it flake in CI.

`ScrapeTokenRotator.reconcile` applies the ServiceMonitor and mints `prometheus-scrape-token` back-to-back in one function call, so prometheus-operator usually sees the Secret before or together with the SM. Operand flakes came from a wider window: the SM was applied during a long `applyManifests` loop while many other resources were still being reconciled.

The operator path also re-patches the ServiceMonitor on every rotator reconcile (immediate start plus the 15-minute ticker), which gives prometheus-operator repeated chances to pick up the target without explicit resync annotations. UWM metrics tests run well after operator install and settle, so they are not racing operator cold-start wiring the way operand targets can.

Swapping order or adding resync would add complexity for symmetry only, not to fix observed behavior. Recommend leaving `ScrapeTokenRotator` as-is for now.

_Posted by Yftach's Cursor_

scripts/operator-e2e/openshift/run-metrics-openshift-tests.sh
yftacherzog: I increased the overall timeout. I don't think it's the best choice, but it's the easiest. If we see that we're hanging on tests that ultimately fail or something like that, then we can explore better approaches.

operator/pkg/kubernetes/component_metrics.go
yftacherzog: It's a good point, but I'd prefer addressing it once we have a component with more than one servicemonitor. This will allow us to pick the most relevant approach rather than trying to predict the actual corner case ahead of time.

operator/internal/common/operand_servicemonitor.go
```diff
@@ -0,0 +1,38 @@
...
+	for _, obj := range objects {
+		if !kubernetes.IsComponentMetricsServiceMonitor(obj) {
+			continue
+		}
+		if obj.GetNamespace() == namespace && obj.GetName() == name {
+			return obj, true
+		}
+	}
+	return nil, false
+}
```
gbenhaim, line 38: **Suggestion (follow-up):** Consider adding a generic `GetByGVK` method on `ObjectStore` in `pkg/manifests` that filters objects by GVK. This would generalize the pattern used by both `GetCRDNamesForComponent` and this function:

```go
func (s *ObjectStore) GetByGVK(component Component, gvk schema.GroupVersionKind) ([]client.Object, error) {
    objects, err := s.GetForComponent(component)
    if err != nil {
        return nil, err
    }
    var result []client.Object
    for _, obj := range objects {
        if obj.GetObjectKind().GroupVersionKind() == gvk {
            result = append(result, obj)
        }
    }
    return result, nil
}
```

Then `OperandServiceMonitorFromObjects` becomes a thin wrapper or can be replaced at call sites. Also opens the door for a type-parameterized variant for typed objects (`ObjectsOfType[*appsv1.Deployment](...)`). Fine as a follow-up.

operator/internal/common/scrape_token.go
```diff
@@ -19,43 +19,236 @@ package common
...
+	if applyErr := cfg.ApplyServiceMonitor(ctx); applyErr != nil {
+		return fmt.Errorf("apply ServiceMonitor %q: %w", cfg.ServiceMonitorName, applyErr)
+	}
+	logApply(
+		"metrics scrape applied ServiceMonitor after token",
+		"namespace", cfg.OperandNamespace,
+		"servicemonitor", cfg.ServiceMonitorName,
+		"existedBeforeApply", smExistedBeforeApply,
+	)
+	return nil
```
gbenhaim: **Nit:** Two log lines are emitted per reconcile here (before + after `ApplyServiceMonitor`). Since `ApplyServiceMonitor` must run every reconcile for tracking-client retention, the "after" log adds no information beyond "apply didn't error". Consider removing the second `logApply` call, or collapsing into a single post-apply line.

test/go-tests/metricsopenshift/uwm_up_only_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func init() {
+	loadMetricsCatalog()
+}
+
+var _ = Describe("UWM Prometheus legacy up targets", Label("openshift", "metrics-uwm-up-only"), func() {
+	Expect(metricsCatalog).NotTo(BeNil())
+	Expect(uwmUpOnlyTargets).NotTo(BeEmpty())
+
+	for _, target := range uwmUpOnlyTargets {
```
gbenhaim, line 18: **Nit:** Unlike `uwm_targets_test.go` which has `target := target` in the loop, this file doesn't. Not a bug with Go 1.22+ loop variable semantics, but inconsistent within the same package. Either add it here for consistency or remove it from `uwm_targets_test.go`.

test/go-tests/pkg/metricsopenshift/prometheus.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	for _, pod := range list.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			return pod.Name, nil
+		}
+	}
+	if len(list.Items) == 0 {
+		return "", fmt.Errorf("no prometheus pods in %s", UWMNamespace)
+	}
+	return list.Items[0].Name, nil
```
gbenhaim: **Minor:** Falling back to `list.Items[0].Name` when no pod is Running means the subsequent port-forward will fail with a less clear error. Consider returning something like `"no running prometheus pods (N pods in non-Running state)"` for better CI diagnostics.

gbenhaim (commented):

Overall this is a high-quality PR — well-designed deferred SM apply, clean resync state machine, excellent test coverage and CI observability. A few suggestions below.

operator/internal/common/scrape_token.go
```diff
@@ -19,43 +19,236 @@ package common
...
+			cfg.OperandNamespace, kubernetes.ScrapeTokenSecretName, kubernetes.ScrapeTokenSecretKey,
+		)
+	}
+
+	smKey := client.ObjectKey{Namespace: cfg.OperandNamespace, Name: cfg.ServiceMonitorName}
+	if err := applyDeferredOperandServiceMonitor(ctx, cfg, secret, smKey); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	sm := &unstructured.Unstructured{}
```
gbenhaim: /qodo why Unstructured is used here an not the ServiceMonitor type?

operator/internal/common/scrape_token.go
```diff
@@ -19,43 +19,236 @@ package common
...
+		Namespace: cfg.OperandNamespace,
+	}, secret); err != nil {
+		return reconcile.Result{}, fmt.Errorf("get scrape token secret after ensure: %w", err)
+	}
+	if len(secret.Data[kubernetes.ScrapeTokenSecretKey]) == 0 {
+		return reconcile.Result{}, fmt.Errorf(
+			"scrape token secret %s/%s has empty %q after ensure",
+			cfg.OperandNamespace, kubernetes.ScrapeTokenSecretName, kubernetes.ScrapeTokenSecretKey,
+		)
+	}
```
gbenhaim: **Suggestion (follow-up):** This second Get reads the secret from the informer cache immediately after `EnsurePrometheusScrapeToken` wrote it to the API server. The cache may not have caught up yet:

- If the secret was just **created**: cache could return NotFound → error + requeue (noisy but self-heals).
- If the secret was just **updated**: cache returns old ResourceVersion → stale RV stored in SM resync annotation → self-corrects on the settle-retry pass.

Similarly, the first Get at line 90 (`secretExistsBefore`) is only needed because `EnsurePrometheusScrapeToken` doesn't report whether it was a create vs update.

Consider having `EnsurePrometheusScrapeToken` return a result struct that carries all the information the caller needs from the write:

```go
type EnsureScrapeTokenResult struct {
    RequeueAfter    time.Duration
    TokenUpdated    bool   // was the token minted/refreshed?
    SecretExisted   bool   // did the secret exist before? (mint vs refresh)
    ResourceVersion string // RV of the secret after the write
}
```

This eliminates both re-reads from a potentially stale cache — `SecretExisted` replaces the first Get (line 90), `ResourceVersion` replaces this second Get. The values come directly from the function that performed the write, so they're correct by construction.

Not blocking — the settle-retry mechanism self-corrects in ≤1 extra reconcile cycle — but would make the code more robust under cache lag. Fine as a follow-up.

gbenhaim (commented):

operator/internal/common/scrape_token.go
gbenhaim: **Change requested:** This second Get reads the secret from the informer cache immediately after `EnsurePrometheusScrapeToken` wrote it to the API server. The cache may not have caught up yet:

- If the secret was just **created**: cache returns NotFound → error + unnecessary requeue.
- If the secret was just **updated**: cache returns stale ResourceVersion → wrong RV stored in SM resync annotation.

Similarly, the first Get at line 90 (`secretExistsBefore`) only exists because `EnsurePrometheusScrapeToken` doesn't report whether it was a create vs update.

Have `EnsurePrometheusScrapeToken` return a result struct that carries the information the caller needs from the write:

```go
type EnsureScrapeTokenResult struct {
    RequeueAfter    time.Duration
    TokenUpdated    bool   // was the token minted/refreshed?
    SecretExisted   bool   // did the secret exist before? (mint vs refresh)
    ResourceVersion string // RV of the secret after the write
}
```

This eliminates both re-reads from a potentially stale cache — `SecretExisted` replaces the first Get (line 90), `ResourceVersion` replaces this second Get. The values come directly from the function that performed the write, so they're correct by construction.

gbenhaim (changes requested):

test/go-tests/pkg/metricsopenshift/wait.go
yftacherzog: done

operator/internal/common/scrape_token.go
```diff
@@ -19,43 +19,236 @@ package common
...
+
+	smKey := client.ObjectKey{Namespace: cfg.OperandNamespace, Name: cfg.ServiceMonitorName}
+	if err := applyDeferredOperandServiceMonitor(ctx, cfg, secret, smKey); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	sm := &unstructured.Unstructured{}
+	sm.SetGroupVersionKind(operandServiceMonitorGVK)
+	smErr := cfg.Client.Get(ctx, smKey, sm)
```
gbenhaim: **Change requested:** Same cache staleness pattern as the secret Get above — `applyDeferredOperandServiceMonitor` just wrote the SM to the API server via `tc.ApplyOwned`, and this Get reads from the informer cache which may not have caught up.

If the SM was just created, this returns NotFound → the resync nudge is skipped entirely and deferred to the settle-retry 15s later.

`applyDeferredOperandServiceMonitor` already probes whether the SM existed before apply (line 218, `smExistedBeforeApply`). Have it return that information so this function doesn't need to re-read from cache to determine the SM's state and annotations.

gbenhaim (commented):

operator/internal/common/scrape_token.go
yftacherzog: ServiceMonitor is optional: the Prometheus Operator CRD may be absent on some clusters. The operator module does not depend on prometheus-operator/api; operand ServiceMonitors come from embedded YAML and are applied as generic client objects, consistent with `operatormetrics/scrape_wiring.go` and `servicemonitor_resync.go`.

Using Unstructured with an explicit GVK supports probe/patch of resync annotations and maps `meta.IsNoMatchError` to "monitoring not installed" without failing reconciliation. This path only needs metadata (existence, annotations), not typed spec fields.

— Yftach's Cursor

operator/internal/common/scrape_token.go
yftacherzog: Good catch. I modified the implementation. Have a look. It might explain some of the unexpected behavior encountered during debugging. I'll need to let it run a few times to make sure things are stable.

operator/internal/common/scrape_token.go
yftacherzog: Good catch. I modified the implementation. Have a look. It might explain some of the unexpected behavior encountered during debugging. I'll need to let it run a few times to make sure things are stable.

yftacherzog:
For retro bot, open an issue with the content of https://github.com/konflux-ci/konflux-ci/pull/8024#discussion_r3579053930

operator/internal/common/operand_servicemonitor.go
yftacherzog: Asked the retro bot to open a followup for this one.

operator/internal/common/scrape_token.go
yftacherzog: done

test/go-tests/metricsopenshift/uwm_up_only_test.go
yftacherzog: removed

test/go-tests/pkg/metricsopenshift/prometheus.go
yftacherzog: done

operator/internal/common/scrape_token.go
yftacherzog: done

yftacherzog:
No flakes since the no-caching change.

gbenhaim (approved):

The cache staleness feedback is addressed — `EnsurePrometheusScrapeToken` now returns an `EnsureScrapeTokenResult` struct with `ResourceVersion`, `SecretExisted`, and `TokenUpdated`, and `applyDeferredOperandServiceMonitor` returns `DeferredSMApplyResult` with the pre-apply SM state. No more post-write cache reads. LGTM.

PR #8070, opened 2026-07-12
Configure Tekton resolver retry/back-off and image pull timeout

gbenhaim:
## Summary

- Adds `configMaps` to `spec.pipeline.options` in the `TektonConfig` CR to improve resilience against transient registry/git failures
- Configures `default-imagepullbackoff-timeout: "15m"` so TaskRuns tolerate transient registry outages (KONFLUX-14906)
- Configures bundle resolver with true exponential back-off: 15s initial, factor 2.0, cap 15m — giving ~15.75 min retry window (KONFLUX-14907)
- Configures git resolver with `fetch-timeout: "2m"` (only supported param for retry tolerance)

## Motivation

Currently:
- TaskRuns fail immediately on the first `ImagePullBackOff` (default timeout is 0)
- Bundle resolver fails in 0 seconds with `CouldntGetTask` on transient registry errors (default: 2 retries only)
- Git resolver has a 1m fetch timeout (increased to 2m for slower networks)

## Bundle resolver back-off progression

```
15s → 30s → 60s → 2m → 4m → 8m → [cap hit, loop exits]
```

Total: ~15.75 minutes across 6 retries with true exponential growth (factor 2.0).

Note: the git resolver does not support `backoff-*` parameters — only `fetch-timeout` is effective.

## Test plan

- [ ] Deploy on Kind with `deploy-deps.sh` and verify:
  - `kubectl get configmap bundleresolver-config -n tekton-pipelines -o yaml` shows backoff params
  - `kubectl get configmap git-resolver-config -n tekton-pipelines -o yaml` shows `fetch-timeout: 2m`
  - `kubectl get configmap config-defaults -n tekton-pipelines -o yaml` shows `default-imagepullbackoff-timeout: 15m`
- [ ] Create a PipelineRun with a non-existent image and confirm it retries for ~15 min
- [ ] Create a PipelineRun with a non-existent bundle and confirm it retries with exponential back-off

### DIFF
--- a/dependencies/tekton-config/tekton-config.yml
+++ b/dependencies/tekton-config/tekton-config.yml
@@ -47,6 +47,20 @@ spec:
     default-service-account: default
     options:
       disabled: false
+      configMaps:
+        config-defaults:
+          data:
+            default-imagepullbackoff-timeout: "15m"
+        bundleresolver-config:
+          data:
+            fetch-timeout: "2m"
+            backoff-steps: "30"
+            backoff-cap: "30s"
+        git-resolver-config:
+          data:
+            fetch-timeout: "2m"
+            backoff-steps: "4"
+            backoff-cap: "30s"
       deployments:
         tekton-pipelines-remote-resolvers:
           spec:
--- a/dependencies/tekton-config/tekton-config.yml
+++ b/dependencies/tekton-config/tekton-config.yml
@@ -54,13 +54,13 @@ spec:
         bundleresolver-config:
           data:
             fetch-timeout: "2m"
-            backoff-steps: "30"
-            backoff-cap: "30s"
+            backoff-duration: "15s"
+            backoff-factor: "2.0"
+            backoff-steps: "10"
+            backoff-cap: "15m"
         git-resolver-config:
           data:
             fetch-timeout: "2m"
-            backoff-steps: "4"
-            backoff-cap: "30s"
       deployments:
         tekton-pipelines-remote-resolvers:
           spec:


dependencies/tekton-config/tekton-config.yml
```diff
@@ -47,6 +47,20 @@ spec:
...
       disabled: false
+      configMaps:
+        config-defaults:
+          data:
+            default-imagepullbackoff-timeout: "15m"
+        bundleresolver-config:
+          data:
+            fetch-timeout: "2m"
+            backoff-steps: "30"
+            backoff-cap: "30s"
```
yftacherzog: With the current settings, the bundle resolver retry window is ~30 seconds, not ~15 minutes.

`backoff-steps` and `backoff-cap` interact with the defaults for `backoff-duration` (2s) and `backoff-factor` (2.0). Exponential growth hits the 30s cap after about four steps, which terminates the loop regardless of `backoff-steps: "30"`.

To approximate a 15-minute window aligned with the image-pull timeout, all four parameters need to be set together, for example:

```yaml
fetch-timeout: "2m"
backoff-duration: "30s"
backoff-factor: "1.0"
backoff-steps: "30"
backoff-cap: "30s"
```

— Yftach's Cursor

dependencies/tekton-config/tekton-config.yml
```diff
@@ -47,6 +47,20 @@ spec:
...
+        git-resolver-config:
+          data:
+            fetch-timeout: "2m"
+            backoff-steps: "4"
+            backoff-cap: "30s"
```
yftacherzog: The git resolver does not support `backoff-steps` or `backoff-cap`. Upstream `ScmConfig` only recognizes keys such as `fetch-timeout`, `default-url`, `default-revision`, and SCM token settings; unknown keys are silently ignored.

These backoff fields apply to the bundle resolver only. For git, only `fetch-timeout: "2m"` will take effect here. Remove the backoff keys and adjust the PR description for KONFLUX-14907 accordingly — git rate-limit handling cannot be configured through resolver backoff in Tekton today.

— Yftach's Cursor

yftacherzog (changes requested):
Two resolver configuration blocks do not match upstream Tekton behavior and will not deliver the stated retry goals. The `default-imagepullbackoff-timeout` change looks correct.

— Yftach's Cursor

yftacherzog (approved):

gbenhaim:
Re: fullsend-ai-review's [comment](https://github.com/konflux-ci/konflux-ci/pull/8070#discussion_r3566944400) on `backoff-steps: 10` vs 6:

The concern is valid in theory, but in practice go-containerregistry's retry implementation (used by the bundle resolver) zeroes the Steps counter once the computed duration exceeds the cap — so the loop terminates after ~6 retries / ~15.75 min regardless of whether `backoff-steps` is 6 or 10. Setting it to 10 is a safe upper bound that doesn't change runtime behavior.

We've verified the actual retry window in production (`stone-prd-rh01`): a PipelineRun with a non-existent image retried for ~16 minutes before failing, confirming the expected behavior.

-- Gal's Cursor

PR #8228, opened 2026-07-16
feat(KFLUXVNGD-1098): add metrics for release service

yftacherzog:
Adding metrics to release-service in the same way it was added to integration-service. I.e. http with no auth, as this is what it currently supports. https + auth will be added once added upstream.

Also cleaning up some copy-pasted ArgoCD annotations from existing metrics-related manifests along the way.

Assisted-by: Cursor


operator/internal/controller/releaseservice/konfluxreleaseservice_controller.go
```diff
@@ -160,12 +169,25 @@ func (r *KonfluxReleaseServiceReconciler) Reconcile(ctx context.Context, req ctr
...
 	for _, obj := range objects {
+		if !metricsEnabled && kubernetes.IsComponentMetricsScrapeResource(obj) {
+			log.V(1).Info("Skipping component metrics scrape resource",
+				"kind", tracking.GetKind(obj),
+				"name", obj.GetName(),
+				"namespace", obj.GetNamespace(),
+			)
+			continue
+		}
+
```
avi-biton, line 190: This skip path is not covered by unit tests. Codecov shows 22.22% patch coverage with 6 missing lines. The integration-service controller has the same gap, so this is a pre-existing pattern, but this PR is a good opportunity to add at least basic coverage for both gating paths (enabled=default, disabled=skip).

_Posted by Avi's Cursor_

yftacherzog: Was repeating the same pattern from integration-service but there the patch was bigger so the missing lines did not get us below 80%.
Added coverage to both now.

avi-biton (approved):

PR #8231, opened 2026-07-16
docs: document per-test scheme isolation for parallel fake-client tests

### LINKED ISSUE #8218: Add fake client per-test scheme isolation guidance to AGENTS.md testing conventions
## What happened

PR #8213 fixed a flaky test race condition caused by parallel tests sharing the global scheme.Scheme via fake.NewClientBuilder() without an explicit scheme. When multiple goroutines concurrently registered ServiceMonitor GVK via AddKnownTypeWithName, Go's runtime panicked with concurrent map writes. The fix was straightforward (use runtime.NewScheme() per test), and the same pattern already existed in metrics_reader_binding_test.go, but neither AGENTS.md nor any skill documented this convention.

## What could go better

The code agent correctly implemented the fix, but only because the issue author explicitly recommended it with code snippets. If AGENTS.md documented the per-test scheme isolation convention, both code and review agents would catch this pattern proactively — the code agent would avoid introducing the race in new tests, and the review agent would flag it during review. Currently, two test files follow the per-test scheme pattern and two others (hashedconfigmap_test.go, resync_evidence_test.go) do not use explicit schemes but are safe only because they lack t.Parallel(). Adding parallelism to those tests without also adding per-test schemes would reintroduce the same bug.

## Proposed change

Add a bullet point to the 'Controller-runtime patterns' subsection in Architecture Notes of AGENTS.md documenting that tests using controller-runtime's fake.NewClientBuilder() in combination with t.Parallel() (or Ginkgo parallel nodes) must provide an explicit per-test scheme via WithScheme(runtime.NewScheme()) rather than relying on the global scheme.Scheme default. Without this, the fake client's Get path may concurrently register GVKs on the shared scheme map, causing fatal panics. The pattern is already established in metrics_reader_binding_test.go and servicemonitor_resync_test.go. Codifying it prevents future contributors from introducing the same race.

## Validation criteria

After the AGENTS.md change is merged: (1) the 'Controller-runtime patterns' section contains guidance about per-test scheme isolation for fake clients in parallel tests, (2) a future code agent writing new parallel tests with fake.NewClientBuilder() includes WithScheme(runtime.NewScheme()) without needing the issue to specify it, and (3) the review agent flags any PR that adds t.Parallel() to a test using fake.NewClientBuilder() without an explicit scheme.

---
_Generated by retro agent from https://github.com/konflux-ci/konflux-ci/pull/8213_

addielaruee:
Closes #8218

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,6 +143,7 @@ skip/review rules, pre-label heuristics, and `ready-for-merge` behavior, apply
 - **Interface minimality** — Reconciler structs should depend on the narrowest type needed. If a reconciler only consumes events from a broadcaster, accept a `<-chan event.TypedGenericEvent[client.Object]`, not the broadcaster itself. Wire via `Subscribe()` in `main.go`.
 - **Component monitoring** — see the controller wiring checklist in [`operator/docs/component-monitoring.md`](operator/docs/component-monitoring.md#controller-wiring-checklist).
 - **Informer cache staleness** — After writing to the API server, do not immediately read the same object via cached `client.Get` — the informer cache may not have caught up. Use write-path return values or carry state forward in reconcile result structs.
+- **Fake client scheme isolation in parallel tests** — Tests using controller-runtime's `fake.NewClientBuilder()` alongside `t.Parallel()` (or Ginkgo parallel nodes) must supply an explicit per-test scheme via `WithScheme(runtime.NewScheme())`, never the shared global `scheme.Scheme`. Concurrent tests registering GVKs on the same shared scheme map cause a fatal "concurrent map writes" panic. See `metrics_reader_binding_test.go` and `servicemonitor_resync_test.go` for the correct pattern.
 
 ## Skills
 


yftacherzog (commented):
@addielaruee thank you for this PR!
Fix the grammar issue and squash the commits so the PR contains a single commit and I think we should be good to go.

addielaruee:
Fixed the subject-verb agreement and squashed into a single commit.

AGENTS.md
```diff
@@ -143,6 +143,7 @@ skip/review rules, pre-label heuristics, and `ready-for-merge` behavior, apply
...
 - **Component monitoring** — see the controller wiring checklist in [`operator/docs/component-monitoring.md`](operator/docs/component-monitoring.md#controller-wiring-checklist).
 - **Informer cache staleness** — After writing to the API server, do not immediately read the same object via cached `client.Get` — the informer cache may not have caught up. Use write-path return values or carry state forward in reconcile result structs.
+- **Fake client scheme isolation in parallel tests** — Tests using controller-runtime's `fake.NewClientBuilder()` alongside `t.Parallel()` (or Ginkgo parallel nodes) must supply an explicit per-test scheme via `WithScheme(runtime.NewScheme())`, never the shared global `scheme.Scheme`. Concurrent tests registering GVKs on the same shared scheme map cause a fatal "concurrent map writes" panic. See `metrics_reader_binding_test.go` and `servicemonitor_resync_test.go` for the correct pattern.
```
yftacherzog, line 146: I think we can soften this a bit. Not sure I want to make too strict generalization.

```suggestion
- **Fake client scheme isolation in parallel tests** — Tests using controller-runtime's `fake.NewClientBuilder()` alongside `t.Parallel()` (or Ginkgo parallel nodes) should typically supply an explicit per-test scheme via `WithScheme(runtime.NewScheme())`, rather than a shared global `scheme.Scheme`. Concurrent tests registering GVKs on the same shared scheme map may cause a fatal "concurrent map writes" panic. See `metrics_reader_binding_test.go` and `servicemonitor_resync_test.go` for the correct pattern.
```

PR #8232, opened 2026-07-16
docs: document upstream CRD field-name casing convention for unstructured writes

### LINKED ISSUE #7930: Document unstructured field-name mapping convention for operator-managed upstream CRs
## What happened

PR #7756 introduces applyReleaseServiceConfigCustomizations() which manipulates an unstructured ReleaseServiceConfig object using string-based field access. The Go KonfluxReleaseServiceSpec type uses json tag 'emptyDirOverrides' (camelCase per Go convention), but the upstream ReleaseServiceConfig CRD schema uses 'EmptyDirOverrides' (PascalCase). The controller code at controller.go lines 241-254 correctly uses the upstream PascalCase names ('EmptyDirOverrides') when writing to the unstructured object. Meanwhile, the 'debug' field happens to be lowercase in both the operator's type and the upstream CRD, avoiding any casing mismatch. This naming difference is intentional but nowhere documented, making it a likely review flag — a reviewer might reasonably question whether 'EmptyDirOverrides' is a bug when the Go struct uses 'emptyDirOverrides'.

## What could go better

The casing difference between the operator's own CRD json tags and the upstream CRD field names should be documented as a deliberate convention. Additionally, using bare string literals for field names in unstructured object access makes the code fragile to upstream changes and harder to review. If upstream renames a field, the string-based access would silently produce incorrect behavior. Defining string constants for unstructured field names (e.g., const releaseServiceConfigOverridesField = 'EmptyDirOverrides') would make them searchable, reviewable, and centrally updatable. Confidence is medium because the existing CI sync checks partially mitigate upstream drift risk.

## Proposed change

Add a section to CONTRIBUTING.md under 'Operator Development' explaining the convention for controllers that manipulate upstream CRs via unstructured objects: (1) field names in unstructured access must match the upstream CRD schema exactly, not the operator's own json tags, (2) define string constants for unstructured field names to make them grep-able and reviewable, (3) note that verify-manifests-in-sync.sh catches upstream CRD schema changes that might affect these field names. Optionally, refactor the existing bare strings in applyReleaseServiceConfigCustomizations() to use named constants.

## Validation criteria

The documentation update should be verifiable by checking that (1) the convention is described in CONTRIBUTING.md, (2) the recommendation for string constants is present, and (3) the review agent can reference this convention when reviewing future PRs that add unstructured field access. A future PR adding similar unstructured manipulation should follow the documented pattern without requiring human reviewers to re-explain the casing convention.

---
_Generated by retro agent from https://github.com/konflux-ci/konflux-ci/pull/7756_

addielaruee:
## What
Adds a short section to `CONTRIBUTING.md` under "Operator Development" documenting the convention for controllers that write to upstream CRDs via `unstructured.Unstructured`: field names must match the upstream CRD schema's casing (e.g. `EmptyDirOverrides`), not the operator's own Go/JSON struct tag casing (`emptyDirOverrides`).

This is based on the existing (correct) pattern in `applyReleaseServiceConfigCustomizations()` in `operator/internal/controller/releaseservice/konfluxreleaseservice_controller.go`, which was previously undocumented and could be mistaken for a typo during review.

Also recommends defining named string constants for these field names instead of inline literals, for searchability and easier updates if the upstream schema changes.

## Why
Without this documented, a reviewer unfamiliar with the upstream `ReleaseServiceConfig` CRD's casing could reasonably flag `EmptyDirOverrides` as a bug, since it doesn't match the operator's own `emptyDirOverrides` struct tag.

Closes #7930

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,7 @@ Contributing Guidelines
 - [Editing Markdown Files](#editing-markdown-files)
 - [Using KubeLinter](#using-kubelinter)
 - [Operator Development](#operator-development)
+  * [Unstructured field names for upstream CRs](#unstructured-field-names-for-upstream-crs)
 - [CI/CD and Testing](#cicd-and-testing)
   * [Operator rendered manifests](#operator-rendered-manifests)
   * [Automated E2E Tests](#automated-e2e-tests)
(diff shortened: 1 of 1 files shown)


addielaruee:
Updated the table of contents with npx markdown-toc so check-toc passes, and dropped the unrelated AGENTS.md commit from this branch since that change belongs to a different issue and PR.

PR #8275, opened 2026-07-18
fix: exclude nested componentMetrics on release and UI

### LINKED ISSUE #8274: Nested componentMetrics under releaseService and ui on the Konflux CR gets ignored
# Title

Nested componentMetrics under releaseService and ui on the Konflux CR gets ignored

# Description

Hey!

On the Konflux parent CR, metrics are supposed to be controlled from the top level field `spec.componentMetrics`. The parent reconciler copies that into each sub CR.

For build, image, and integration that already worked cleanly. Their nested config types on the parent do not include `componentMetrics`, so you cannot set it under those sections.

Release and UI were different. They nested the full sub CR spec on the parent, and that type includes `componentMetrics`. So kubectl and the CRD happily accepted something like this

```yaml
apiVersion: konflux.konflux-ci.dev/v1alpha1
kind: Konflux
metadata:
  name: konflux
spec:
  componentMetrics:
    enabled: true
  releaseService:
    spec:
      componentMetrics:
        enabled: false
  ui:
    spec:
      componentMetrics:
        enabled: false
```

A person reading that would think release and UI metrics are off. They were not. The parent controller copied the nested spec and then overwrote `componentMetrics` from the top level value in `applyKonfluxReleaseService` and `applyKonfluxUI`.

So the nested field showed up in the API, looked real, and did nothing useful.

## Expected

Parent nested specs for release and UI should match build and integration. No `componentMetrics` on `Konflux.spec.releaseService.spec` or `Konflux.spec.ui.spec`. Only top level `spec.componentMetrics` should control this, and that value gets forwarded into the real sub CRs.

## Actual (before the fix)

`ReleaseServiceConfig.Spec` was `*KonfluxReleaseServiceSpec` and `KonfluxUIConfig.Spec` was `*KonfluxUISpec`. Both carried `componentMetrics`, so the nested field was valid on the parent CR and then silently overwritten.

## Fix

Matched release and UI to the build and integration pattern.

1. Added `KonfluxReleaseServiceConfigSpec` and `KonfluxUIConfigSpec` without `componentMetrics`.
2. Embedded those into the real sub CR specs (`KonfluxReleaseServiceSpec` and `KonfluxUISpec`), which still keep `componentMetrics` for forwarding.
3. Pointed parent `ReleaseServiceConfig` and `KonfluxUIConfig` at the ConfigSpec types.
4. Updated `applyKonfluxReleaseService` and `applyKonfluxUI` to copy the ConfigSpec and set `ComponentMetrics` from `ForwardedComponentMetrics`.
5. Added `NewKonfluxReleaseServiceSpec` and `NewKonfluxUISpec` helpers.
6. Regenerated deepcopy and CRDs. The parent Konflux CRD now only exposes top level `componentMetrics`.

## Test

`operator/api/v1alpha1/konflux_nested_component_metrics_test.go`

`TestKonfluxParentReleaseAndUINestedComponentMetricsExcluded` passes. It checks that nested `componentMetrics` in parent JSON is ignored, forwarding uses top level metrics only, and parent Spec types do not expose `ComponentMetrics`.

Also covered by `TestNewKonfluxReleaseServiceSpec` and `TestNewKonfluxUISpec`.

## Files

- `operator/api/v1alpha1/konflux_types.go`
- `operator/api/v1alpha1/konfluxreleaseservice_types.go`
- `operator/api/v1alpha1/konfluxui_types.go`
- `operator/api/v1alpha1/operand_specs.go`
- `operator/api/v1alpha1/operand_specs_test.go`
- `operator/api/v1alpha1/konflux_nested_component_metrics_test.go`
- `operator/api/v1alpha1/zz_generated.deepcopy.go`
- `operator/config/crd/bases/konflux.konflux-ci.dev_konfluxes.yaml`
- `operator/internal/controller/konflux/konflux_controller.go`

false200:
## Summary

Fixes ignored nested `componentMetrics` under `releaseService` and `ui` on the parent Konflux CR by aligning those configs with the build/integration ConfigSpec pattern. Only top-level `spec.componentMetrics` is accepted and forwarded.

Closes #8274 

## Test plan

- [x] `go test ./api/v1alpha1/ -run TestKonfluxParentReleaseAndUINestedComponentMetricsExcluded`
- [x] `go test ./api/v1alpha1/ -run 'TestNewKonfluxReleaseServiceSpec|TestNewKonfluxUISpec'`
- [ ] Confirm parent Konflux CRD no longer documents nested `componentMetrics` under release/UI


yftacherzog:
@false200 thank you for opening the issue and this PR.
Can you take a look at the failures in lint and unit tests?

false200:
> @false200 thank you for opening the issue and this PR. Can you take a look at the failures in lint and unit tests?

Sure.

yftacherzog:
@false200 can we get patch code coverage at 80% (or higher)?

false200:
> @false200 can we get patch code coverage at 80% (or higher)?

Sure!

operator/api/v1alpha1/konflux_nested_component_metrics_test.go
yftacherzog, line 1: I think we should drop this file.
It's testing the move from one model to the other specifically, and I don't think it will be needed in any context other than this PR.

We should make the tests in line with what we have for the components that are modeled correctly.

false200: Yup, makes sense. I'll make that change!

yftacherzog (approved):

false200:
@yftacherzog, it looks like the patch coverage only improved a bit with my changes (77.15% → 77.67% in the latest report). Would you like me to work on increasing it further?

yftacherzog:
> @yftacherzog, it looks like the patch coverage only improved a bit with my changes (77.15% → 77.67% in the latest report). Would you like me to work on increasing it further?

77.67% is the project's coverage. You don't need to solve that :)
The line above it says `✅ All modified and coverable lines are covered by tests.` which means you're good for the patch.

false200:
> > @yftacherzog, it looks like the patch coverage only improved a bit with my changes (77.15% → 77.67% in the latest report). Would you like me to work on increasing it further?
> 
> 77.67% is the project's coverage. You don't need to solve that :) The line above it says `✅ All modified and coverable lines are covered by tests.` which means you're good for the patch.

Got it, thanks for clarifying.

PR #8282, opened 2026-07-19
feat(metrics): auth + https for release-service metrics

yftacherzog:
release-service now supports authentication. This change utilizes this when deploying the release-service metrics service monitor, and also enables https for it.

Assisted-by: Cursor

PR #8324, opened 2026-07-21
feat(metrics): auth + https for integration-service metrics

yftacherzog:
integration-service now supports authentication. This change utilizes this when deploying the integration-service metrics service monitor, and also enables https for it.

Assisted-by: Cursor

Note: this is repeating the same pattern as https://github.com/konflux-ci/konflux-ci/pull/8282


avi-biton:
operator/pkg/kubernetes/component_metrics.go (pre-existing, surfaced by this PR)

Problem: `IsComponentMetricsScrapeResource()` does not recognize metrics-scraper as a metrics ServiceAccount. The `isMetricsReaderServiceAccountName` check only matches "metrics-reader" or names ending in "-metrics-reader". When `ComponentMetrics.Enabled` is false, the controller will still apply the metrics-scraper SA because it isn't classified as a scrape resource.

Mitigation: This is a pre-existing gap that also affects build-service and release-service (which already ship metrics-scraper SAs). Since the SA is harmless when unused (no token is minted for it), the impact is limited to a leftover SA when metrics are disabled. Worth a follow-up PR to add MetricsScraperServiceAccountName to the gating predicate.

_Posted by Avi's Cursor_

avi-biton (approved):

PR #8415, opened 2026-07-24
feat: Make konflux-ci release version configurable

tkdchen:
If konflux-ci operator install method is 'release', the release URL always points to the latest. This does not work for installation from an alternative version rather than the latest, e.g. v0.2.2-rc.9 or even an older version. The mismatch version may cause compatibility issue.

Here is an example, checkout to version (tag) v0.2.2-rc.9 and deploy a local instance, then the latest version of konflux-ci operator is installed, then the following error is reported:

```
Konflux in version "v1alpha1" cannot be handled as a Konflux: strict decoding error: unknown field
"spec.buildService.spec.pacWebhookInsecureSSL"
```

The expectation is the matched version is installed.

This commit introduces an optional environment variable `RELEASE_VERSION`. Installing the latest is still the default behavior. A specific version can be assigned on-demand.

### DIFF
--- a/scripts/deploy-local.env.template
+++ b/scripts/deploy-local.env.template
@@ -58,6 +58,13 @@ OPERATOR_IMAGE=""
 # See operator/config/samples/ for available configurations.
 # KONFLUX_CR=""
 
+# Release version used in conjunction with 'release' operator installation method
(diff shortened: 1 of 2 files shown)


tkdchen:
Trying the `local` method.

PR #8432, opened 2026-07-25
scripts: add Minikube as an alternative local deployment driver

FaisalAl-Rayes:
## Summary

- Adds `DEPLOY_DRIVER=minikube` support to `scripts/deploy-local.sh`, making it the single entry point for both Kind and Minikube local deployments
- Moves `scripts/minikube/setup-minikube-local-cluster.sh` up to `scripts/` to sit symmetrically alongside `scripts/setup-kind-local-cluster.sh`
- Removes `scripts/minikube/deploy-local.sh` thin wrapper — callers use `DEPLOY_DRIVER=minikube ./scripts/deploy-local.sh` directly
- Updates all internal path and comment references accordingly

## Test plan

- [ ] `./scripts/deploy-local.sh` (default Kind path) continues to work unchanged
- [ ] `DEPLOY_DRIVER=minikube ./scripts/deploy-local.sh` spins up a Minikube cluster and deploys Konflux
- [ ] `DEPLOY_DRIVER=invalid` exits with a clear error message
- [ ] Existing `DEPLOY_LOCAL_SKIP_KIND=1` env var still works (aliased to `DEPLOY_LOCAL_SKIP_CLUSTER`)

### DIFF
--- a/scripts/deploy-local.sh
+++ b/scripts/deploy-local.sh
@@ -3,18 +3,19 @@
 # Deploy Konflux for Local Development
 #
 # This script provides a one-command local development deployment of Konflux
-# on a Kind cluster. It's designed for LOCAL DEVELOPMENT CONVENIENCE ONLY.
+# on a Kind or Minikube cluster. It's designed for LOCAL DEVELOPMENT
+# CONVENIENCE ONLY.
 #
 # For production deployments on real clusters, see docs/operator-deployment.md
 #
 # What this script does:
-#  1. Creates a Kind cluster with proper configuration
+#  1. Creates a Kind/Minikube cluster with proper configuration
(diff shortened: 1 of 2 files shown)


yftacherzog:
Hi @FaisalAl-Rayes,
Thank you for exploring this!

I think that if we're going towards another deployment driver, we might want to check something more lightweight comparing to Kind/Minikube. k3s/k3d was suggested in the past, but was not explored thoroughly enough.

The reason being that with Kind, we're already on the edge of maxing the resources of the free GitHub runners. Not sure if minikube will take us one way or the other with that, but some more lightweight approaches might.

Whatever we pick, it needs to be something that runs in CI as well so we test it for regressions on every change.

PR #8433, opened 2026-07-25
ci: run e2e tests on k3s

yftacherzog:
We're stretching the resource limits with e2e tests based on kind. Trying to deploy on k3s instead, which is more lightweight and should be able to run on our CI.

Assisted-by: Cursor


yftacherzog:
closing for now.

PR #8442, opened 2026-07-26
fix(docs): close apply-konflux-cr code fence

### LINKED ISSUE #8441: Fix broken code fence in apply-konflux-cr docs
The bash block at the end of `operator/docs/content/docs/guides/apply-konflux-cr.md` closes with four backticks instead of three. Change ` ```` ` to ` ``` ` so the page renders correctly.

false200:
Fixes: #8441


yftacherzog (approved):

false200:
Here the e2e failure looks unrelated to this change as it's only a markdown fence fix in applykonfluxcr.md.

PR #8445, opened 2026-07-26
fix(docs): correct smee manifest path

### LINKED ISSUE #8444: Fix wrong smee manifest path in troubleshooting docs
# Description

The troubleshooting guide tells users to run `kubectl` against `./smee/smee-client.yaml`, but the file lives at `./dependencies/smee/smee-client.yaml`. Update those two commands so they point at the right path.

false200:
Fixes: #8444

### DIFF
--- a/operator/docs/content/docs/troubleshooting.md
+++ b/operator/docs/content/docs/troubleshooting.md
@@ -156,9 +156,9 @@ It may take a few minutes for the UI to become available again.
    verify the smee channel URL in the smee-client manifest, then redeploy it:
 
    ```bash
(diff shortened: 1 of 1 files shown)


yftacherzog (approved):
Thank you for raising the issue and creating the PR!
We have some issues in CI due to an upstream bug.
We'll get it merged once that is fixed and CI passes.

false200:
> Thank you for raising the issue and creating the PR! We have some issues in CI due to an upstream bug. We'll get it merged once that is fixed and CI passes.

Sure.

PR #8447, opened 2026-07-26
fix(#8417): non-viable deployment defaults

### LINKED ISSUE #8417: deploy-local.sh fails with strict decoding error when using default release operator method
Running SKIP_SECRETS=true ./scripts/deploy-local.sh from main fails at Applying Konflux configuration with a
  strict decoding error.

Logs: 
  ```
Error from server (BadRequest): error when creating "operator/config/samples/konflux_v1alpha1_konflux.yaml":
  Konflux in version "v1alpha1" cannot be handled as a Konflux: strict decoding error:
    unknown field "spec.integrationService.spec.snapshotGarbageCollector"
    unknown field "spec.namespaceLister.spec.cacheResyncPeriod"
    unknown field "spec.namespaceLister.spec.logLevel"
    unknown field "spec.ui.spec.dex.config.staticPasswords[0].groups"
    unknown field "spec.ui.spec.dex.config.staticPasswords[1].groups"
    unknown field "spec.ui.spec.proxy.reverseProxy"
```
 Root Cause
  
  The default `OPERATOR_INSTALL_METHOD=release` installs the latest released operator, but the sample CR on main (at
`  v0.2.2-rc.9-161-ge1fcfa16`) contains fields not yet in the released CRD schema.

  Steps to Reproduce

  1. `git clone https://github.com/konflux-ci/konflux-ci.git && cd konflux-ci`
  2. `cp scripts/deploy-local.env.template scripts/deploy-local.env`
  3. `SKIP_SECRETS=true ./scripts/deploy-local.sh`

  Either the sample CR should stay compatible with the released operator, or the deploy script should handle the mismatch.

### PR DESCRIPTION
**Pull request by yftacherzog**
The default deployment method for local deployments combines CR from local files with released image, which are likely to be incompatible. This commit changes the deployment methods to be more coherent and set the default to be based on a commit SHA instead an existing release.

Assisted-by: Cursor

Closes: #8417

### DIFF
--- a/scripts/deploy-local.sh
+++ b/scripts/deploy-local.sh
@@ -15,13 +15,14 @@
 #
 # Prerequisites:
 #  - kind, kubectl, podman (or docker)
-#  - kustomize (only for 'local' install method)
+#  - kustomize (only for 'checkout' / 'build' install methods)
 #  - Configuration file: scripts/deploy-local.env
 #
 # Usage:
 #   ./scripts/deploy-local.sh [konflux-cr-file]
 #
 # By default, uses operator/config/samples/konflux_v1alpha1_konflux.yaml
+# (or the matching sample from a GitHub release when using OPERATOR_INSTALL_METHOD=release).
 #
 # Example:
 #   cp scripts/deploy-local.env.template scripts/deploy-local.env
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: scripts/deploy-local.sh [hunk 1]
```diff
@@ -91,32 +199,109 @@ export SEGMENT_WRITE_KEY
...
+esac
+
+# Resolve Konflux CR
+# For release: use samples from the same release unless the caller set KONFLUX_CR /
+# passed a positional CR path (explicit override).
+KONFLUX_CR_EXPLICIT="${1:-${KONFLUX_CR:-}}"
+if [ "${INSTALL_METHOD}" = "release" ] && [ -z "${KONFLUX_CR_EXPLICIT}" ]; then
+    RELEASE_SAMPLES_DIR="$(fetch_release_samples "${OPERATOR_RELEASE}")"
+    # shellcheck disable=SC2064
+    trap 'rm -rf "${RELEASE_SAMPLES_DIR}"' EXIT
```
avi-biton: This unconditionally sets a new EXIT trap. If a future change adds another EXIT trap earlier in the script, this will silently overwrite it. Consider appending to any existing trap or using a cleanup function pattern. Not blocking, but worth noting.

_Posted by Avi's Cursor_

scripts/deploy-local.sh
```diff
@@ -55,6 +62,107 @@ set -euo pipefail
...
+    samples_dir="$(mktemp -d "${TMPDIR:-/tmp}/konflux-release-samples.XXXXXX")"
+    echo "Downloading release samples: ${url}" >&2
+    if ! curl -fsSL "${url}" | tar -xzf - -C "${samples_dir}"; then
+        echo "ERROR: Failed to download or extract release samples from ${url}" >&2
+        rm -rf "${samples_dir}"
+        exit 1
+    fi
+    # Samples may be archived with a leading ./ ; normalize to the extract dir.
+    echo "${samples_dir}"
+}
```
avi-biton, line 156: The function extracts into a temp dir and returns it, with a comment about normalizing ./-prefixed paths. But it doesn't actually normalize — it just returns the raw extract directory. If the tarball has a subdirectory (e.g. samples/konflux_v1alpha1_konflux.yaml instead of konflux_v1alpha1_konflux.yaml), resolve-konflux-cr.sh won't find the file. This depends on the tarball structure in releases, which the author presumably controls, so it's low risk but worth a defensive check or a comment documenting the expected tarball layout.

_Posted by Avi's Cursor_

avi-biton:
The [CONTRIBUTING.md](https://github.com/konflux-ci/konflux-ci/blob/main/CONTRIBUTING.md?plain=1#L112) file still refer to `OPERATOR_INSTALL_METHOD=build`.
The file should be updated

yftacherzog:
> The [CONTRIBUTING.md](https://github.com/konflux-ci/konflux-ci/blob/main/CONTRIBUTING.md?plain=1#L112) file still refer to `OPERATOR_INSTALL_METHOD=build`. The file should be updated

fixed

scripts/deploy-local.sh
yftacherzog: I changed the approach so we use a git-ignored path inside the repo and clean it on start -- no trap.

scripts/deploy-local.sh
yftacherzog: Should be better now 🤞

avi-biton (approved):

PR #8597, opened 2026-07-31
chore: read segment-bridge secret

Omeramsc:
The segment-bridge namespace creates a secret with the required segment source key per environment. this change make it so the secret is read by the cron-job.

A segment key supplied during the deployment is prefferred, so local deployments are unaffected.

Assisted-by: Cursor


operator/internal/controller/segmentbridge/konfluxsegmentbridge_controller_test.go
```diff
@@ -281,6 +282,157 @@ var _ = Describe("KonfluxSegmentBridge Controller", func() {
...
+			// followed by waitForSecret/Eventually on another owned resource) before calling this.
+			createExternalSecret := func(ctx context.Context, name, key, value string) {
+				secret := &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: segmentBridgeNamespace},
+					Data:       map[string][]byte{key: []byte(value)},
+				}
+				Expect(k8sClient.Create(ctx, secret)).To(Succeed())
+				DeferCleanup(func() {
+					_ = k8sClient.Delete(context.Background(), secret)
+				})
```
yftacherzog: Prefer `DeferCleanup(testutil.DeleteAndWait, k8sClient, secret)` here so cleanup waits until the Secret is gone.

That matches the helper used elsewhere in this PR (`createExternalSegmentKeySecret` in the UI controller tests) and other controller tests. Specs in this context reuse Secret names (e.g. `vault-segment-key`), so a non-blocking delete can leave a brief window where a later `Create` fails with `AlreadyExists`.

— Yftach's Cursor

Omeramsc: fixed

yftacherzog (approved):

PR #8667, opened 2026-08-04
feat(KFLUXVNGD-1169): Create trusted-ca ConfigMap on OpenShift in build-service and integration-service

avi-biton:
The trusted-ca ConfigMap is required by both build-service and integration-service deployments for custom CA trust. On non-OpenShift clusters, trust-manager creates it automatically via a Bundle resource. On OpenShift, trust-manager is skipped and the ConfigMap had to be created externally.

Add a shared helper (common.EnsureTrustedCAConfigMap) that creates the ConfigMap with the config.openshift.io/inject-trusted-cabundle label when the operator detects OpenShift. This brings the resource under operator lifecycle management, eliminating the need for external provisioning. OpenShift's cluster network operator populates the ConfigMap with the cluster CA bundle automatically.

The build-service controller already had ClusterInfo; the integration-service reconciler gains it here (wired in main.go). Non-OpenShift behavior is unchanged.

Follow-up: once this merges, update the operator reference downstream and remove the external trusted-ca ConfigMap creation from downstream deployments - the operator now manages it.

### DIFF
--- a/operator/internal/common/trusted_ca.go
+++ b/operator/internal/common/trusted_ca.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2025 Konflux CI.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/konflux-ci/konflux-ci/operator/pkg/clusterinfo"
+	"github.com/konflux-ci/konflux-ci/operator/pkg/tracking"
+)
+
+const (
+	// TrustedCAConfigMapName is the name of the ConfigMap that OpenShift's
+	// cluster network operator populates with the cluster-wide trusted CA bundle.
+	TrustedCAConfigMapName = "trusted-ca"
+
+	// OpenShiftInjectTrustedCABundleLabel is the label that triggers OpenShift's
+	// CA bundle injection into a ConfigMap.
+	OpenShiftInjectTrustedCABundleLabel = "config.openshift.io/inject-trusted-cabundle"
+)
+
+// EnsureTrustedCAConfigMap creates or updates the trusted-ca ConfigMap with the
+// OpenShift CA injection label in the given namespace. On non-OpenShift clusters
+// (or when clusterInfo is nil) this is a no-op. OpenShift's cluster network operator
+// automatically populates the ca-bundle.crt key when this label is present.
+func EnsureTrustedCAConfigMap(ctx context.Context, namespace string, tc *tracking.Client, ci *clusterinfo.Info) error {
+	if ci == nil || !ci.IsOpenShift() {
+		return nil
+	}
+	configMap := &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      TrustedCAConfigMapName,
+			Namespace: namespace,
+			Labels: map[string]string{
+				OpenShiftInjectTrustedCABundleLabel: "true",
+			},
+		},
+	}
+	if err := tc.ApplyOwned(ctx, configMap); err != nil {
+		return fmt.Errorf("failed to apply trusted-ca ConfigMap in %s: %w", namespace, err)
+	}
+	return nil
+}
--- a/operator/internal/common/trusted_ca_test.go
+++ b/operator/internal/common/trusted_ca_test.go
@@ -1,0 +1,314 @@
+/*
+Copyright 2025 Konflux CI.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/version"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	"github.com/konflux-ci/konflux-ci/operator/pkg/clusterinfo"
+	"github.com/konflux-ci/konflux-ci/operator/pkg/tracking"
+)
+
+type mockDiscoveryClient struct {
+	resources map[string]*metav1.APIResourceList
+}
+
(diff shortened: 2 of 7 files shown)


operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -177,11 +179,20 @@ func (r *KonfluxIntegrationServiceReconciler) Reconcile(ctx context.Context, req
...
 	}
 
+	// On OpenShift, create the trusted-ca ConfigMap with the injection label so
+	// the cluster network operator populates it with the cluster CA bundle.
+	// This runs after applyManifests to ensure the namespace exists.
+	if r.ClusterInfo != nil && r.ClusterInfo.IsOpenShift() {
+		if err := common.EnsureTrustedCAConfigMap(ctx, integrationServiceNamespace, tc); err != nil {
+			return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "ensure trusted-ca ConfigMap")
+		}
+	}
```
yftacherzog, line 198: Consider aligning the reconcile order with build-service: ensure the namespace exists, then create `trusted-ca`, then `applyManifests`.

That way the ConfigMap is present before Deployments that mount it are applied. Volumes are already `optional: true`, so the current order is not a hard failure — this is mainly for consistency and earlier availability of the CM.

— Yftach's Cursor

operator/internal/common/trusted_ca.go
```diff
@@ -0,0 +1,61 @@
...
+			Namespace: namespace,
+			Labels: map[string]string{
+				OpenShiftInjectTrustedCABundleLabel: "true",
+			},
+		},
+	}
+	if err := tc.ApplyOwned(ctx, configMap); err != nil {
+		return fmt.Errorf("failed to apply trusted-ca ConfigMap in %s: %w", namespace, err)
+	}
+	return nil
```
yftacherzog: Worth locking in the SSA contract that this helper must not claim `data` (so CNO-injected `ca-bundle.crt` survives re-apply).

A unit test that seeds `data.ca-bundle.crt`, calls `EnsureTrustedCAConfigMap` again, and asserts the key is still present would document that invariant. Leaving `Data` unset (as here) is the right shape; avoid adding an empty `Data` map later.

— Yftach's Cursor

operator/internal/common/trusted_ca_test.go
```diff
@@ -0,0 +1,133 @@
...
+		if cm.Namespace != "test-namespace" {
+			t.Errorf("expected namespace %q, got %q", "test-namespace", cm.Namespace)
+		}
+		val, ok := cm.Labels[OpenShiftInjectTrustedCABundleLabel]
+		if !ok {
+			t.Fatal("expected injection label to be present")
+		}
+		if val != "true" {
+			t.Errorf("expected injection label value %q, got %q", "true", val)
+		}
```
yftacherzog, line 129: Consider also asserting the ownership / component labels set by `ApplyOwned`. Lifecycle and orphan tracking depend on those labels, and the injection-label checks alone do not cover that path.

— Yftach's Cursor

operator/internal/controller/integrationservice/konfluxintegrationservice_controller_test.go
```diff
@@ -1497,3 +1499,127 @@ var _ = Describe("KonfluxIntegrationService Controller", func() {
...
+			ObjectMeta: metav1.ObjectMeta{Name: CRName},
+		}
+		Expect(k8sClient.Create(ctx, integrationService)).To(Succeed())
+	})
+
+	AfterEach(func() {
+		testutil.DeleteAndWait(ctx, k8sClient, integrationService)
+		testutil.DeleteAndWait(ctx, k8sClient, &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: "trusted-ca", Namespace: integrationServiceNamespace},
+		})
```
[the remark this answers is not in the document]
avi-biton: Not applicable here. `DeferCleanupParentAndChildren` is used when a test **asserts on cluster-scoped children** (e.g. the ClusterRole/ClusterRoleBinding drift-correction tests at line ~765 in the same file). Those tests need explicit cleanup because they verify specific cluster-scoped resources are created/restored, and envtest won't GC them reliably.

These trusted-ca tests only assert on a **namespace-scoped ConfigMap**. The cluster-scoped resources (ClusterRole, ClusterRoleBinding) are created as a side effect of `applyManifests` but are not relevant to the test assertions. They don't interfere because each test starts its own fresh manager via `startManagerWithClusterInfo`, and the AfterEach stops it.

This is the same pattern used by the existing SCC tests (lines 122-127) in the same file, which have been stable without `DeferCleanupParentAndChildren`.

---
*Posted by Avi's Cursor*

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -181,6 +181,14 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return errHandler.HandleWithReason(ctx, err, condition.ReasonNamespaceCreationFailed, "ensure namespace exists")
 	}
 
+	// On OpenShift, create the trusted-ca ConfigMap with the injection label so
+	// the cluster network operator populates it with the cluster CA bundle.
+	if r.ClusterInfo != nil && r.ClusterInfo.IsOpenShift() {
```
yftacherzog: Maybe this can go into the helper.

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
yftacherzog: Human says: worth it just for consistency -- if possible.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Done

operator/internal/common/trusted_ca_test.go
avi-biton: Done

operator/internal/common/trusted_ca.go
avi-biton: Done

operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
avi-biton: Done

yftacherzog:
Need to increase the coverage it seems

avi-biton:
> Need to increase the coverage it seems

The coverage was increased, however to keep on increasing it we need to test the error path which we do not do anywhere in the code.

yftacherzog:
> > Need to increase the coverage it seems
> 
> The coverage was increased, however to keep on increasing it we need to test the error path which we do not do anywhere in the code.

What do you mean by testing the error path?

avi-biton:
> > > Need to increase the coverage it seems
> > 
> > 
> > The coverage was increased, however to keep on increasing it we need to test the error path which we do not do anywhere in the code.
> 
> What do you mean by testing the error path?

We do not have tests for error paths such as:

```
	// Apply all embedded manifests
	if err := r.applyManifests(ctx, tc, buildService, webhookConfigMapName); err != nil {
		return errHandler.HandleApplyError(ctx, err)
	}
```

```
	webhookConfigMapName, err := r.reconcileWebhookConfig(ctx, buildService)
	if err != nil {
		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "reconcile webhook config")
	}
```
In build-service and all across the operator.
These are the same error paths as the new uncovered lines in this PR:
```
	// On OpenShift, create the trusted-ca ConfigMap with the injection label so
	// the cluster network operator populates it with the cluster CA bundle.
	if err := common.EnsureTrustedCAConfigMap(ctx, webhookConfigNamespace, tc, r.ClusterInfo); err != nil {
		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "ensure trusted-ca ConfigMap")
	}
```

yftacherzog:
> > > > Need to increase the coverage it seems
> > > 
> > > 
> > > The coverage was increased, however to keep on increasing it we need to test the error path which we do not do anywhere in the code.
> > 
> > 
> > What do you mean by testing the error path?
> 
> We do not have tests for error paths such as:
> 
> ```
> 	// Apply all embedded manifests
> 	if err := r.applyManifests(ctx, tc, buildService, webhookConfigMapName); err != nil {
> 		return errHandler.HandleApplyError(ctx, err)
> 	}
> ```
> 
> ```
> 	webhookConfigMapName, err := r.reconcileWebhookConfig(ctx, buildService)
> 	if err != nil {
> 		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "reconcile webhook config")
> 	}
> ```
> 
> In build-service and all across the operator. These are the same error paths as the new uncovered lines in this PR:
> 
> ```
> 	// On OpenShift, create the trusted-ca ConfigMap with the injection label so
> 	// the cluster network operator populates it with the cluster CA bundle.
> 	if err := common.EnsureTrustedCAConfigMap(ctx, webhookConfigNamespace, tc, r.ClusterInfo); err != nil {
> 		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "ensure trusted-ca ConfigMap")
> 	}
> ```

🤷

yftacherzog (approved):

gbenhaim (approved):

PR #8691, opened 2026-08-05
feat(segment-bridge):  RBAC for telemetry testing

Omeramsc:
Deploy a namespace-scoped Role and RoleBinding in segment-bridge so telemetry validation can read segment-bridge resources and optionally trigger CronJob runs. These permissions are required to perform maintenance and e2e tests for the telemetry feature against staging segment

Assisted-by: Cursor


yftacherzog (changes requested):
The `oamsalem` references should not have gone in. Correct?

Omeramsc:
> The `oamsalem` references should not have gone in. Correct?

Yes. closing it for now

PR #8703, opened 2026-08-05
feat(KFLUXVNGD-1179): add konflux_up Prometheus gauge for platform readiness

avi-biton:
Expose a `konflux_up` metric (gauge, 1 or 0) on the operator's `/metrics` endpoint. 
The value reflects the Konflux CR's aggregated Ready condition, which is True only when all component sub-CRs report Ready and the `cert-manager` dependency gate passes.

This is the foundation for Prometheus-based health monitoring of the Konflux deployment, enabling alerting rules to be layered on top.

### DIFF
--- a/operator/internal/controller/konflux/konflux_controller.go
+++ b/operator/internal/controller/konflux/konflux_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -47,6 +48,7 @@ import (
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/releaseservice"
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/segmentbridge"
 	uictrl "github.com/konflux-ci/konflux-ci/operator/internal/controller/ui"
+	"github.com/konflux-ci/konflux-ci/operator/internal/operatormetrics"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/clusterinfo"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/tracking"
 )
@@ -166,8 +168,14 @@ func (r *KonfluxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// Fetch the Konflux instance
 	konflux := &konfluxv1alpha1.Konflux{}
 	if err := r.Get(ctx, req.NamespacedName, konflux); err != nil {
+		if apierrors.IsNotFound(err) {
+			operatormetrics.SetKonfluxUp(false)
+		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+
+	// Update health metric on every exit path based on the final Ready condition.
+	defer func() { operatormetrics.SetKonfluxUp(konflux.IsReady()) }()
 
 	log.Info("Reconciling Konflux", "name", konflux.Name)
 
--- a/operator/internal/controller/konflux/konflux_controller_test.go
+++ b/operator/internal/controller/konflux/konflux_controller_test.go
@@ -21,6 +21,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	prometheustestutil "github.com/prometheus/client_golang/prometheus/testutil"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -48,6 +49,7 @@ import (
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/segmentbridge"
 	"github.com/konflux-ci/konflux-ci/operator/internal/controller/testutil"
 	uictrl "github.com/konflux-ci/konflux-ci/operator/internal/controller/ui"
+	"github.com/konflux-ci/konflux-ci/operator/internal/operatormetrics"
 	"github.com/konflux-ci/konflux-ci/operator/pkg/clusterinfo"
 )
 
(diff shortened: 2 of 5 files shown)


operator/internal/controller/konflux/konflux_controller.go
```diff
@@ -403,6 +404,9 @@ func (r *KonfluxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// Check cert-manager availability, set CertManagerAvailable condition, and override Ready if missing.
 	certManagerResult := r.checkCertManagerAvailability(ctx, konflux)
 
+	// Update health metric based on final Ready condition.
+	metrics.SetKonfluxUp(konflux.IsReady())
+
```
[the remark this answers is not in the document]
yftacherzog: Those kind of issues will probably be real issues. In the past, we implemented this metric as a recording rule. If we can do it reliably in code, then fine. If not, we can fall-back to a downstream recording rule.

yftacherzog (changes requested):

operator/internal/controller/konflux/konflux_controller.go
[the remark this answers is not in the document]
avi-biton: Handled

operator/internal/controller/konflux/konflux_controller.go
```diff
@@ -166,9 +167,14 @@ func (r *KonfluxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// Fetch the Konflux instance
 	konflux := &konfluxv1alpha1.Konflux{}
 	if err := r.Get(ctx, req.NamespacedName, konflux); err != nil {
+		// CR deleted or not found — platform is not ready.
+		metrics.SetKonfluxUp(false)
```
yftacherzog: `SetKonfluxUp(false)` runs for every `Get` error, not only NotFound. Transient failures (timeouts, connectivity, etc.) will flip `konflux_up` to `0` even when the last known Ready state was True.

Clear the gauge only on NotFound; leave the last known value unchanged on other `Get` errors:

```go
if err := r.Get(ctx, req.NamespacedName, konflux); err != nil {
	if apierrors.IsNotFound(err) {
		metrics.SetKonfluxUp(false)
	}
	return ctrl.Result{}, client.IgnoreNotFound(err)
}
```

— Yftach's Cursor

operator/internal/controller/konflux/konflux_controller_test.go
```diff
@@ -844,6 +846,42 @@ var _ = Describe("Konflux Controller", func() {
...
+	Context("konflux_up metric", func() {
+		const resourceName = "konflux"
+
+		It("should set konflux_up to 0 when the Konflux CR does not exist", func(_ context.Context) {
+			startManager(createTestClusterInfo())
+
+			// No CR created — reconciler receives NotFound and should set gauge to 0.
+			Eventually(func() float64 {
+				return prometheustestutil.ToFloat64(metrics.KonfluxUpGauge())
+			}).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Equal(float64(0)))
```
yftacherzog, line 877: This assertion passes from the gauge default (`0`) without exercising the NotFound path. With no Konflux CR, `For(&Konflux{})` does not enqueue a reconcile, so `SetKonfluxUp(false)` is never called.

Exercise the delete/NotFound path explicitly — e.g. seed `SetKonfluxUp(true)`, create then delete the CR (or call `Reconcile` for a missing name), then assert the gauge is `0`.

— Yftach's Cursor

yftacherzog (changes requested):
Two reliability issues around `konflux_up` update paths.

— Yftach's Cursor

operator/internal/controller/konflux/konflux_controller.go
avi-biton: Done

operator/internal/controller/konflux/konflux_controller_test.go
avi-biton: Done

operator/internal/controller/konflux/konflux_controller_test.go
```diff
@@ -844,6 +846,42 @@ var _ = Describe("Konflux Controller", func() {
 			}).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Succeed())
 		})
 	})
+
+	Context("konflux_up metric", func() {
+		const resourceName = "konflux"
+
+		It("should set konflux_up to 0 when the Konflux CR does not exist", func(_ context.Context) {
```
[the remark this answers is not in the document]
avi-biton: Done

operator/internal/controller/konflux/konflux_controller_test.go
```diff
@@ -844,6 +846,58 @@ var _ = Describe("Konflux Controller", func() {
...
 	})
+
+	Context("konflux_up metric", func() {
+		const resourceName = "konflux"
+
+		It("should set konflux_up to 0 when the Konflux CR is deleted", func(ctx context.Context) {
+			startManager(createTestClusterInfo())
+
+			// Seed gauge to 1 so we can verify it actually flips.
+			metrics.SetKonfluxUp(true)
```
yftacherzog: The seed happens before create/reconcile. After the Ready condition appears, the successful reconcile's `defer` already sets `konflux_up` to `0` (Ready is False without sub-controllers). The post-delete assertion then passes even if the NotFound path never clears the gauge.

Re-seed after reconcile settles, then delete:

```go
// ... wait for Ready condition ...

metrics.SetKonfluxUp(true)
Expect(k8sClient.Delete(ctx, cr)).To(Succeed())

Eventually(...).Should(Equal(float64(0)))
```

— Yftach's Cursor

yftacherzog (changes requested):
One remaining test-integrity issue on the NotFound path coverage.

— Yftach's Cursor

operator/internal/controller/konflux/konflux_controller_test.go
avi-biton: Great finding.
Fixed.

yftacherzog (approved):

operator/internal/operatormetrics/health.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var konfluxUp = prometheus.NewGauge(prometheus.GaugeOpts{
+	Name: "konflux_up",
+	Help: "Whether the Konflux platform is ready (1) or not (0). Aggregated from all component sub-CR readiness conditions.",
+})
```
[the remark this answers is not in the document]
avi-biton: Konflux operator is always deployed as a single replica (unlike the sub-components which have configurable Replicas fields), so the non-leader scraping concern doesn't apply.

PR #8752, opened 2026-08-10
test(KFLUXVNGD-1179): add e2e test for konflux_up ecosystem label wiring

avi-biton:
Add an e2e test that queries UWM Prometheus for
`konflux_up{service="konflux-operator"}` to verify the service ConstLabel survives Prometheus target-label collision resolution.

On failure, the test dumps all `konflux_up` labels and every time series from the konflux-operator namespace for diagnosis.

This test is expected to fail until the operator ServiceMonitor adds a metricRelabelings rule to restore the service label.


avi-biton:
This PR is only for tests.
Closing it.

PR #8842, opened 2026-08-13
fix(KFLUXSE-462): strip scheme from Segment apiHost served to the browser SDK

Omeramsc:
The Segment analytics-next SDK builds request URLs as
`${protocol}://${apiHost}`. The UI controller wrote the full
GetSegmentAPIURL() value (always https://...) into the browser-facing
segment-bridge-config Secret, producing malformed https://https://...
URLs and silently breaking user_login, user_logout, and
feedback_submitted events.

This change strip away the scheme only when populating the Secret served
at /segment/url. CronJob SEGMENT_BATCH_API construction is unchanged and still
uses the full URL with scheme

Assisted-by: Cursor

### DIFF
--- a/operator/pkg/segment/segment.go
+++ b/operator/pkg/segment/segment.go
@@ -19,6 +19,7 @@ package segment
 import (
 	"context"
 	"fmt"
+	"net/url"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


operator/pkg/segment/segment.go
```diff
@@ -90,6 +91,34 @@ func ResolveWriteKeySecretRef(
...
+//
+// If rawURL fails to parse, it is returned unchanged so callers fail loudly
+// downstream rather than silently losing data.
+func StripURLScheme(rawURL string) string {
+	u, err := url.Parse(rawURL)
+	if err != nil || u.Host == "" {
+		return rawURL
+	}
+	return u.Host + u.Path
+}
```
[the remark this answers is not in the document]
Omeramsc: fixed

shacharFridman (approved):
LGTM

operator/pkg/segment/segment.go
```diff
@@ -90,6 +91,34 @@ func ResolveWriteKeySecretRef(
...
+// in the browser before any network request is attempted.
+//
+// This helper must only be applied where the value is destined for the
+// browser SDK (the segment-bridge-config Secret's "url" key, served at
+// /segment/url). It must NOT be applied to the CronJob's SEGMENT_BATCH_API
+// construction, which correctly needs the scheme intact.
+//
+// If rawURL fails to parse, it is returned unchanged so callers fail loudly
+// downstream rather than silently losing data.
+func StripURLScheme(rawURL string) string {
```
hguemar: It bugs me but the function does a little more than strip the url scheme, it also drops query and fragment.
Maybe not an issue but it might be important to document it or rename (e.g: ExtractURLHostAndPath) to avoid someone else using the function incorrectly later.

hguemar (commented):
Code is correct, I would rename the function to avoid that in later code submissions it is used incorrectly assuming that it only strips URL Scheme.

operator/pkg/segment/segment.go
Omeramsc: Thanks for the review; Fixed.

avi-biton (approved):

PR #8843, opened 2026-08-13
feat: conformance tests can use pre-provisioned clusters

Troy876:
Adds support for running conformance tests against pre-provisioned clusters where namespaces are externally managed via GitOps. When `E2E_MANAGED_NAMESPACE` is set, the suite adapts setup and cleanup behavior to work within limited RBAC permissions and support parallel test execution.

## Pre-provisioned mode (`E2E_MANAGED_NAMESPACE` set)

- Detects pre-provisioned mode and adapts `setup-release.sh` at runtime via regex patches
- `patchECPForE2E` and `grantIntegrationRunnerJobRBAC` are non-fatal only in pre-provisioned mode
- Uses `|| true` for roleBinding commands and namespace creation instead of blanket `set -e` removal
- Verifies release testing prerequisites before starting release testing

## Parallel run isolation

- All resources (application, component, ECP, trusted-artifacts, release) are suffixed with a random per-run identifier to prevent collisions
- `release-pipeline` ServiceAccount secrets are managed via Go client (append/remove) instead of `kubectl apply` to avoid overwriting concurrent runs
- ECP is renamed per run (`ecp-<suffix>`) so cleanup of one run doesn't break another's release validation
- `trusted-artifacts` ImageRepository is renamed per run for full isolation

## ServiceAccount management

- `ensureReleasePipelineSA` creates the SA if it doesn't exist (first run in fresh namespaces)
- `ensureSASecret` appends push secrets with retry logic for ResourceVersion conflicts
- `pruneDanglingSASecrets` removes dangling secret references before each run
- `removeSASecrets` cleans up only the current run's secrets on teardown

## Cleanup

- Calls kube client API for individual resource cleanup instead of namespace deletion
- Non-fatal polling loop in `AfterAll` waits for release PipelineRun without aborting cleanup if it never appears
- Context cancellation check in cleanup polling to ensure graceful exit

## Environment variables

| Variable | Purpose | Default |
|----------|---------|---------|
| `E2E_REPO` | Override the component source repo | `testrepo` |
| `E2E_ITS_REPO` | Override the IntegrationTestScenario repo (should be public unless secrets are configured) | `testrepo` |
| `E2E_MANAGED_NAMESPACE` | Triggers pre-provisioned mode | — |
| `E2E_APPLICATIONS_NAMESPACE` | Sets the tenant namespace | — |

### DIFF
--- a/test/go-tests/tests/conformance/conformance.go
+++ b/test/go-tests/tests/conformance/conformance.go
@@ -29,6 +29,7 @@ import (
 	e2eConfig "github.com/konflux-ci/konflux-ci/test/go-tests/tests/conformance/config"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // conformanceCleanupBudget returns the max wall time to wait for AfterAll cleanup.
@@ -61,7 +62,9 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 
 	fw := &framework.Framework{}
 	var buildPipelineAnnotation map[string]string
-	var componentNewBaseBranch, gitRevision, componentRepositoryName, componentName string
+	var appName string
+	var componentNewBaseBranch, gitRevision, componentRepositoryName, componentName, releaseName string
+	var releaseRes setupReleaseResult
 
 	for _, appSpec := range e2eConfig.UpstreamAppSpecs {
 		appSpec := appSpec
@@ -79,22 +82,65 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 				userNamespace = fw.UserNamespace
-				managedNamespace = "default-managed-tenant"
-				klog.Info("conformance: namespaces ready", "user", userNamespace, "managed", managedNamespace)
+				managedNamespace = os.Getenv("E2E_MANAGED_NAMESPACE")
+				if managedNamespace == "" {
+					managedNamespace = "default-managed-tenant"
+				}
+				klog.Infof("conformance: namespaces ready user=%s managed=%s", userNamespace, managedNamespace)
 
-				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, util.GenerateRandomString(4))
+				suffix := util.GenerateRandomString(4)
+				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, suffix)
+				appName = fmt.Sprintf("%s-%s", appSpec.ApplicationName, suffix)
 				pacBranchName = constants.PaCPullRequestBranchPrefix + componentName
 				componentRepositoryName = utils.ExtractGitRepositoryNameFromURL(appSpec.ComponentSpec.GitSourceUrl)
 
-				gomega.Expect(runSetupRelease(appSpec.ApplicationName, componentName, userNamespace, managedNamespace)).To(gomega.Succeed())
-				gomega.Expect(patchECPForE2E(fw.AsKubeAdmin, "default", managedNamespace)).To(gomega.Succeed())
-				gomega.Expect(grantIntegrationRunnerJobRBAC(userNamespace)).To(gomega.Succeed())
+				if isPreProvisioned() {
+					client := fw.AsKubeAdmin.CommonController.KubeRest()
+					pruneDanglingSASecrets(context.Background(), client, managedNamespace)
+					gomega.Expect(ensureReleasePipelineSA(context.Background(), client, managedNamespace)).To(gomega.Succeed())
+				}
+
+				releaseName = fmt.Sprintf("release-%s", suffix)
+				var setupErr error
+				releaseRes, setupErr = runSetupRelease(appName, componentName, userNamespace, managedNamespace, releaseName)
+				gomega.Expect(setupErr).NotTo(gomega.HaveOccurred())
+				preProvisioned := isPreProvisioned()
+				if preProvisioned {
+					// Pre-provisioned setup uses intentionally mixed error handling:
+					// mutations (ensureSASecret, patchECP, grantRBAC) are non-fatal
+					// warnings because the cluster may already be partially configured,
+					// while verifications (verifyECPPatched, verifyReleasePrerequisites)
+					// are fatal Expects because tests cannot succeed without them.
+					client := fw.AsKubeAdmin.CommonController.KubeRest()
+					runIRNames := []string{componentName, releaseRes.TrustedArtifactsIR}
+					if err := ensureSASecret(context.Background(), client, managedNamespace, runIRNames); err != nil {
+						klog.Warningf("conformance: ensureSASecret failed (non-fatal): %v", err)
+					}
+					if err := patchECPForE2E(fw.AsKubeAdmin, releaseRes.ECPName, managedNamespace); err != nil {
+						klog.Warningf("conformance: patchECPForE2E failed (non-fatal in pre-provisioned env): %v", err)
+					}
+					if err := grantIntegrationRunnerJobRBAC(fw.AsKubeAdmin, userNamespace); err != nil {
+						klog.Warningf("conformance: grantIntegrationRunnerJobRBAC failed (non-fatal in pre-provisioned env): %v", err)
+					}
+					gomega.Expect(verifyECPPatched(fw.AsKubeAdmin, releaseRes.ECPName, managedNamespace)).To(gomega.Succeed())
+					if err := verifyPreProvisionedRBAC(fw.AsKubeAdmin, userNamespace); err != nil {
+						klog.Warningf("conformance: pre-provisioned RBAC check failed (non-fatal, tests may still pass): %v", err)
+					}
+					gomega.Expect(verifyReleasePrerequisites(fw.AsKubeAdmin, managedNamespace)).To(gomega.Succeed())
+				} else {
+					gomega.Expect(patchECPForE2E(fw.AsKubeAdmin, releaseRes.ECPName, managedNamespace)).To(gomega.Succeed())
+					gomega.Expect(grantIntegrationRunnerJobRBAC(fw.AsKubeAdmin, userNamespace)).To(gomega.Succeed())
+				}
 
 				buildPipelineAnnotation = build.GetBuildPipelineBundleAnnotation(appSpec.ComponentSpec.BuildPipelineType)
 			})
 
 			ginkgo.AfterAll(func() {
-				if strings.EqualFold(os.Getenv("E2E_SKIP_CLEANUP"), "true") || ginkgo.CurrentSpecReport().Failed() || strings.Contains(ginkgo.GinkgoLabelFilter(), upstreamKonfluxTestLabel) {
+				skipCleanup := strings.EqualFold(os.Getenv("E2E_SKIP_CLEANUP"), "true")
+				if !isPreProvisioned() {
+					skipCleanup = skipCleanup || ginkgo.CurrentSpecReport().Failed() || strings.Contains(ginkgo.GinkgoLabelFilter(), upstreamKonfluxTestLabel)
+				}
+				if skipCleanup {
 					return
 				}
 				klog.Info("conformance: cleaning up")
@@ -102,7 +148,14 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				ctx, cancel := context.WithTimeout(context.Background(), budget)
 				defer cancel()
 
-				_ = fw.AsKubeAdmin.CommonController.KubeInterface().CoreV1().Namespaces().Delete(ctx, managedNamespace, metav1.DeleteOptions{})
+				if isPreProvisioned() {
+					cleanupTenantResources(ctx, fw, userNamespace, appName, componentName, releaseName)
+				} else {
+					if err := fw.AsKubeAdmin.CommonController.KubeInterface().CoreV1().Namespaces().Delete(ctx, managedNamespace, metav1.DeleteOptions{}); err != nil {
+						klog.Warningf("conformance cleanup: delete managed namespace %s: %v", managedNamespace, err)
+					}
+				}
+
 				cleanupWithRetry(ctx, "delete PaC branch", func() error {
 					return fw.AsKubeAdmin.CommonController.GitHub.DeleteRef(ctx, componentRepositoryName, pacBranchName)
 				})
@@ -112,14 +165,37 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				cleanupWithRetry(ctx, "cleanup webhooks", func() error {
 					return build.CleanupWebhooks(ctx, fw, componentRepositoryName)
 				})
+
+				if isPreProvisioned() {
+					// The release PipelineRun is created asynchronously by the
+					// release-service and may not exist until ~50s after the
+					// last spec finishes. Poll until it appears, then sweep.
+					// This is non-fatal: if the test failed before reaching the
+					// release stage, no PipelineRun will ever appear and that's fine.
+					client := fw.AsKubeAdmin.CommonController.KubeRest()
+					appLabel := crclient.MatchingLabels{"appstudio.openshift.io/application": appName}
+					pollDeadline := time.Now().Add(60 * time.Second)
+					for time.Now().Before(pollDeadline) {
+						if ctx.Err() != nil {
+							klog.Warningf("conformance cleanup: context cancelled while waiting for release PipelineRun")
+							break
+						}
+						prList := &tektonapi.PipelineRunList{}
+						if err := client.List(ctx, prList, crclient.InNamespace(managedNamespace), appLabel); err == nil && len(prList.Items) > 0 {
+							break
+						}
+						time.Sleep(5 * time.Second)
+					}
+					cleanupManagedResources(ctx, fw, managedNamespace, appName, componentName, releaseName, releaseRes)
+				}
 			})
 
 			// --- Application & Component Setup ---
 
 			ginkgo.It("creates an application", ginkgo.Label(devEnvTestLabel, upstreamKonfluxTestLabel), func() {
-				createdApplication, createErr := fw.AsKubeAdmin.HasController.CreateApplication(appSpec.ApplicationName, userNamespace)
+				createdApplication, createErr := fw.AsKubeAdmin.HasController.CreateApplication(appName, userNamespace)
 				gomega.Expect(createErr).NotTo(gomega.HaveOccurred())
-				gomega.Expect(createdApplication.Spec.DisplayName).To(gomega.Equal(appSpec.ApplicationName))
+				gomega.Expect(createdApplication.Spec.DisplayName).To(gomega.Equal(appName))
 				gomega.Expect(createdApplication.Namespace).To(gomega.Equal(userNamespace))
 			})
 
@@ -127,7 +203,7 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				its := appSpec.ComponentSpec.IntegrationTestScenario
 				gomega.Eventually(func() error {
 					var createErr error
-					integrationTestScenario, createErr = fw.AsKubeAdmin.IntegrationController.CreateIntegrationTestScenario("", appSpec.ApplicationName, userNamespace, its.GitURL, its.GitRevision, its.TestPath, "", []string{})
+					integrationTestScenario, createErr = fw.AsKubeAdmin.IntegrationController.CreateIntegrationTestScenario("", appName, userNamespace, its.GitURL, its.GitRevision, its.TestPath, "", []string{})
 					return createErr
 				}, time.Minute*2, time.Second*5).Should(gomega.Succeed(), "timed out creating IntegrationTestScenario")
 			})
@@ -142,7 +218,7 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 			ginkgo.It(fmt.Sprintf("creates component %s (private: %t) from git source %s", appSpec.ComponentSpec.Name, appSpec.ComponentSpec.Private, appSpec.ComponentSpec.GitSourceUrl), ginkgo.Label(devEnvTestLabel, upstreamKonfluxTestLabel), func() {
 				componentObj := appservice.ComponentSpec{
 					ComponentName: componentName,
-					Application:   appSpec.ApplicationName,
+					Application:   appName,
 					Source: appservice.ComponentSource{
 						ComponentSourceUnion: appservice.ComponentSourceUnion{
 							GitSource: &appservice.GitSource{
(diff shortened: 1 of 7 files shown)


test/go-tests/tests/conformance/setup.go
```diff
@@ -43,12 +65,26 @@ func runSetupRelease(appName, componentName, tenantNS, managedNS string) error {
...
+		} else {
+			return fmt.Errorf("setup-release.sh: %w", err)
+		}
+	}
+	return nil
+}
+
+// namespaceExists returns true if the namespace already exists on the cluster.
+func namespaceExists(ns string) bool {
+	cmd := exec.Command(resolveKubectl(), "get", "namespace", ns, "--no-headers")
```
flacatus: lets try to use a kubeclient please somehow in the tests instead of exec.

test/go-tests/tests/conformance/setup.go
```diff
@@ -209,6 +245,41 @@ func dumpDiagnostics(hub *framework.ControllerHub, componentName, appName, names
...
+		klog.Warningf("conformance cleanup: delete PipelineRuns in %s: %v", ns, err)
+	}
+	if err := fw.AsKubeAdmin.HasController.DeleteAllImageRepositoriesInASpecificNamespace(ns, cleanupTimeout); err != nil {
+		klog.Warningf("conformance cleanup: delete ImageRepositories in %s: %v", ns, err)
+	}
+	if err := fw.AsKubeAdmin.TektonController.DeleteEnterpriseContractPolicy("default", ns, false); err != nil {
+		klog.Warningf("conformance cleanup: delete ECP in %s: %v", ns, err)
+	}
+	// Delete all ReleasePlanAdmissions via kubectl since the framework has no bulk list method.
+	cmd := exec.Command(resolveKubectl(), "delete", "releaseplanadmissions.appstudio.redhat.com", "--all", "-n", ns)
```
flacatus: dont use exec.

test/go-tests/tests/conformance/setup.go
```diff
@@ -209,6 +245,41 @@ func dumpDiagnostics(hub *framework.ControllerHub, componentName, appName, names
 	}
 }
 
+// cleanupManagedResources deletes test-created resources from the managed namespace
+// on pre-provisioned clusters where we cannot delete the namespace itself.
+func cleanupManagedResources(_ context.Context, fw *framework.Framework, ns string) {
```
flacatus: we dont need this if we remove component application

flacatus (approved):
left some comments. @yftacherzog if u have time can u review? We are going to run this tests in each ring

flacatus (approved):

test/go-tests/tests/conformance/setup.go
```diff
@@ -33,6 +51,45 @@ func runSetupRelease(appName, componentName, tenantNS, managedNS string) error {
...
+				name: "namespace-creation",
+				re:   regexp.MustCompile(`(?s)kubectl apply -f - <<EOF\napiVersion: v1\nkind: Namespace\n.*?\nEOF`),
+				repl: []byte(`echo "Namespace ${MANAGED_NS} already exists, skipping creation"`),
+			},
+			{
+				// The || true must go on the kubectl line (after <<EOF), NOT after the
+				// closing EOF marker -- placing it after EOF breaks the heredoc terminator.
+				name: "rolebinding-non-fatal",
+				re:   regexp.MustCompile(`(?m)(kubectl apply -f - <<EOF\napiVersion: rbac\.authorization\.k8s\.io/v1\nkind: RoleBinding)`),
+				repl: []byte("kubectl apply -f - <<EOF || true\napiVersion: rbac.authorization.k8s.io/v1\nkind: RoleBinding"),
```
avi-biton: **`test/go-tests/tests/conformance/setup.go` line 74 — `rolebinding-non-fatal` regex matches multiple locations**

The `rolebinding-non-fatal` regex:

```go
regexp.MustCompile(`(?m)(kubectl apply -f - <<EOF\napiVersion: rbac\.authorization\.k8s\.io/v1\nkind: RoleBinding)`)
```

matches **both** RoleBinding blocks in `setup-release.sh`:

- Step 4 (line 227): `authenticated-konflux-viewer` RoleBinding
- Step 10 (line 332): `release-pipeline-resource-role-binding` RoleBinding

This may be intentional (both should tolerate RBAC errors on pre-provisioned clusters), but the current match-count check (line 87) only detects **zero** matches — it can't distinguish one match from two. If the upstream script adds a third RoleBinding block, `|| true` would silently suppress errors on it too.

**Suggestion:** use `re.FindAllIndex` to validate the match count per patch and log a warning when it differs from the expected count:

```go
type scriptPatch struct {
    name          string
    re            *regexp.Regexp
    repl          []byte
    expectedCount int // 0 means "any number > 0"
}
```

This way, if upstream adds or removes a RoleBinding block, you'll get a clear signal instead of silent drift.

_Avi's Cursor_

test/go-tests/tests/conformance/setup.go
```diff
@@ -209,6 +328,177 @@ func dumpDiagnostics(hub *framework.ControllerHub, componentName, appName, names
...
+				pr.Finalizers = nil
+				if err := client.Update(ctx, pr); err != nil {
+					klog.Warningf("conformance cleanup: strip finalizers from PipelineRun %s/%s: %v", ns, pr.Name, err)
+				}
+			}
+			if err := client.Delete(ctx, pr); err != nil {
+				klog.Warningf("conformance cleanup: delete PipelineRun %s/%s: %v", ns, pr.Name, err)
+			}
+		}
+	}
```
avi-biton: **`test/go-tests/tests/conformance/setup.go` lines 331–351 — `cleanupManagedResources` strips finalizers from all PipelineRuns in the namespace**

On a shared pre-provisioned cluster, the managed namespace may contain PipelineRuns from other concurrent test runs. This code lists **all** PipelineRuns in the namespace, strips their finalizers, and deletes them — including runs that don't belong to this test execution.

Stripping finalizers bypasses controller-managed cleanup (e.g. Tekton Chains signing), which could leave other runs' artifacts unsigned or in an inconsistent state.

**Fix:** Filter PipelineRuns to only those owned by this test run. For example:

- Filter by the application label: `crclient.MatchingLabels{"appstudio.openshift.io/application": appName}`
- Or add a test-run-specific label during setup and filter on it during cleanup

The same concern applies to the `rpaList` cleanup below (lines 359–368) — `ReleasePlanAdmissionList` is also unfiltered and would delete RPAs created by other concurrent runs.

_Avi's Cursor_

test/go-tests/tests/conformance/conformance.go
```diff
@@ -79,30 +79,60 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+				preProvisioned := isPreProvisioned()
+				if preProvisioned {
+					if err := patchECPForE2E(fw.AsKubeAdmin, "default", managedNamespace); err != nil {
+						klog.Warningf("conformance: patchECPForE2E failed (non-fatal in pre-provisioned env): %v", err)
+					}
+					if err := grantIntegrationRunnerJobRBAC(fw.AsKubeAdmin, userNamespace); err != nil {
+						klog.Warningf("conformance: grantIntegrationRunnerJobRBAC failed (non-fatal in pre-provisioned env): %v", err)
+					}
+					verifyECPPatched(fw.AsKubeAdmin, "default", managedNamespace)
+					verifyPreProvisionedRBAC(fw.AsKubeAdmin, userNamespace)
```
avi-biton: **`test/go-tests/tests/conformance/conformance.go` lines 97–105 — `verifyECPPatched` and `verifyPreProvisionedRBAC` only warn, never fail**

After the non-fatal `patchECPForE2E` and `grantIntegrationRunnerJobRBAC` calls, the verification functions `verifyECPPatched` (setup.go:479) and `verifyPreProvisionedRBAC` (setup.go:462) only log warnings — they never return an error or assert. This means if the ECP is missing required exclusions, the suite continues and the release pipeline later fails with an opaque EC policy violation instead of a clear setup error.

Compare with `verifyReleasePrerequisites` on line 106, which correctly returns an error and is wrapped in `gomega.Expect`:

```go
gomega.Expect(verifyReleasePrerequisites(fw.AsKubeAdmin, managedNamespace)).To(gomega.Succeed())
```

**Suggestion:** Make `verifyECPPatched` and `verifyPreProvisionedRBAC` return errors and assert them the same way:

```go
gomega.Expect(verifyECPPatched(fw.AsKubeAdmin, "default", managedNamespace)).To(gomega.Succeed())
gomega.Expect(verifyPreProvisionedRBAC(fw.AsKubeAdmin, userNamespace)).To(gomega.Succeed())
```

This gives a clear, early setup failure pointing at the root cause rather than a confusing downstream policy violation or integration test failure.

_Avi's Cursor_

test/go-tests/tests/conformance/conformance.go
```diff
@@ -112,6 +142,15 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
 				cleanupWithRetry(ctx, "cleanup webhooks", func() error {
 					return build.CleanupWebhooks(ctx, fw, componentRepositoryName)
 				})
+
+				if isPreProvisioned() {
+					// The release PipelineRun is created asynchronously by the
+					// release-service and may not exist until ~50s after the
+					// last spec finishes. Wait so the sweep catches it.
+					klog.Info("conformance: waiting for late-arriving release PipelineRuns")
+					time.Sleep(15 * time.Second)
```
avi-biton: **`test/go-tests/tests/conformance/conformance.go` lines 146–153 — 15-second sleep may be insufficient for late-arriving PipelineRuns**

The comment on line 148 says PipelineRuns "may not exist until ~50s after the last spec finishes", but the sleep on line 151 is only 15 seconds. The assumption is that preceding cleanup operations (tenant resource deletion, PaC branch cleanup, webhook cleanup) add enough wall time to cover the gap — but if those operations complete quickly, the total elapsed time could be well under 50s, leaving orphaned PipelineRuns in the managed namespace.

**Suggestion:** Replace the fixed sleep with a short poll loop that waits for the expected release PipelineRun to appear before sweeping, e.g.:

```go
gomega.Eventually(func() bool {
    prList := &pipelinev1.PipelineRunList{}
    if err := client.List(ctx, prList, crclient.InNamespace(managedNamespace)); err != nil {
        return false
    }
    return len(prList.Items) > 0
}, 60*time.Second, 5*time.Second).Should(gomega.BeTrue(),
    "timed out waiting for release PipelineRun in managed namespace")
```

This is both more reliable and faster in the common case (no unnecessary sleep when the PipelineRun arrives early).

_Avi's Cursor_

avi-biton:
conformance/README.md not updated

The `README` describes namespace creation as unconditional (line 17: "Create a managed namespace") and does not mention E2E_MANAGED_NAMESPACE or pre-provisioned mode. The Configuration section should document the new env var.

_Avi's Cursor_

test/go-tests/tests/conformance/setup.go
Troy876: Added the expectedCount field to the scriptPatch struct and the loop now validates match counts and logs warnings when they differ from expected.

test/go-tests/tests/conformance/setup.go
Troy876: PipelineRuns are now filtered by appstudio.openshift.io/application label, and RPAs are deleted by their specific releaseName rather than listing all.

test/go-tests/tests/conformance/conformance.go
Troy876: verifyPreProvisionedRBAC is kept non-fatal intentionally since the namespace-scoped check can produce false negatives when ClusterRoles are bound at cluster level.

test/go-tests/tests/conformance/conformance.go
Troy876: Replaced the fixed sleep with a poll loop (60s timeout, 5s interval) filtered by the application label. Times out gracefully if no PipelineRun appears.

test/go-tests/tests/conformance/conformance.go
```diff
@@ -79,30 +80,74 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+					}
+					gomega.Expect(verifyECPPatched(fw.AsKubeAdmin, ecpName, managedNamespace)).To(gomega.Succeed())
+					if err := verifyPreProvisionedRBAC(fw.AsKubeAdmin, userNamespace); err != nil {
+						klog.Warningf("conformance: pre-provisioned RBAC check failed (non-fatal, tests may still pass): %v", err)
+					}
+					gomega.Expect(verifyReleasePrerequisites(fw.AsKubeAdmin, managedNamespace)).To(gomega.Succeed())
+				} else {
+					gomega.Expect(patchECPForE2E(fw.AsKubeAdmin, ecpName, managedNamespace)).To(gomega.Succeed())
+					gomega.Expect(grantIntegrationRunnerJobRBAC(fw.AsKubeAdmin, userNamespace)).To(gomega.Succeed())
+				}
```
avi-biton: Looks like there is an indentation issue in these lines

test/go-tests/tests/conformance/conformance.go
```diff
@@ -112,6 +157,25 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+				pollDeadline := time.Now().Add(60 * time.Second)
+				for time.Now().Before(pollDeadline) {
+					prList := &tektonapi.PipelineRunList{}
+					if err := client.List(ctx, prList, crclient.InNamespace(managedNamespace), appLabel); err == nil && len(prList.Items) > 0 {
+						break
+					}
+					time.Sleep(5 * time.Second)
+				}
+				cleanupManagedResources(ctx, fw, managedNamespace, appSpec.ApplicationName, componentName, releaseName)
+				}
```
avi-biton: Looks like there is an indentation issue in these lines

test/go-tests/tests/conformance/const.go
```diff
@@ -16,6 +16,10 @@ const (
 	// (Go duration string, e.g. 90s, 2m).
 	conformanceCleanupMaxDuration = 2 * time.Minute
 
+	// cleanupResourceTimeout is used by per-resource cleanup functions when
+	// framework methods require a wait budget.
+	cleanupResourceTimeout = 2 * time.Minute
```
avi-biton: Its defined but unused

test/go-tests/tests/conformance/conformance.go
```diff
@@ -79,30 +80,74 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
-				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, util.GenerateRandomString(4))
-				pacBranchName = constants.PaCPullRequestBranchPrefix + componentName
-				componentRepositoryName = utils.ExtractGitRepositoryNameFromURL(appSpec.ComponentSpec.GitSourceUrl)
-
-				gomega.Expect(runSetupRelease(appSpec.ApplicationName, componentName, userNamespace, managedNamespace)).To(gomega.Succeed())
-				gomega.Expect(patchECPForE2E(fw.AsKubeAdmin, "default", managedNamespace)).To(gomega.Succeed())
-				gomega.Expect(grantIntegrationRunnerJobRBAC(userNamespace)).To(gomega.Succeed())
+			suffix := util.GenerateRandomString(4)
+			componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, suffix)
+			appSpec.ApplicationName = fmt.Sprintf("%s-%s", appSpec.ApplicationName, suffix)
```
avi-biton: Although appSpec := appSpec copies the struct at line 68, this mutation inside BeforeAll means the original application name (from e2eConfig.UpstreamAppSpecs) is effectively lost for debugging purposes. If the spec were ever shared or the loop refactored, this would silently corrupt other iterations. A local variable would be safer.

Fix: Use a local appName := fmt.Sprintf(...) and pass it through, rather than mutating the copied struct's field.

_Avi's Cursor_

test/go-tests/tests/conformance/conformance.go
avi-biton: You can use Eventually for polling

test/go-tests/tests/conformance/conformance.go
Troy876: Should be fixed

test/go-tests/tests/conformance/conformance.go
Troy876: Should be fixed.

test/go-tests/tests/conformance/const.go
Troy876: Removed

test/go-tests/tests/conformance/conformance.go
Troy876: Replaced with a local appName variable so the copied struct stays untouched.

test/go-tests/tests/conformance/conformance.go
Troy876: I intentionally avoided gomega.Eventually here because this runs in AfterAll cleanup. If the assertion fails (PipelineRun never appears), Eventually would abort cleanup. The manual loop times out gracefully and proceeds to cleanupManagedResources regardless.

Let me know if that's an issue.

test/go-tests/tests/conformance/conformance.go
```diff
@@ -79,30 +81,74 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+					managedNamespace = "default-managed-tenant"
+				}
 				klog.Info("conformance: namespaces ready", "user", userNamespace, "managed", managedNamespace)
 
-				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, util.GenerateRandomString(4))
+					suffix := util.GenerateRandomString(4)
+				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, suffix)
+				appName = fmt.Sprintf("%s-%s", appSpec.ApplicationName, suffix)
 				pacBranchName = constants.PaCPullRequestBranchPrefix + componentName
 				componentRepositoryName = utils.ExtractGitRepositoryNameFromURL(appSpec.ComponentSpec.GitSourceUrl)
```
avi-biton: Wrong indentation

test/go-tests/tests/conformance/conformance.go
```diff
@@ -112,22 +158,41 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
+					appLabel := crclient.MatchingLabels{"appstudio.openshift.io/application": appName}
+					pollDeadline := time.Now().Add(60 * time.Second)
+					for time.Now().Before(pollDeadline) {
+						prList := &tektonapi.PipelineRunList{}
+						if err := client.List(ctx, prList, crclient.InNamespace(managedNamespace), appLabel); err == nil && len(prList.Items) > 0 {
+							break
+						}
+						time.Sleep(5 * time.Second)
+					}
+					cleanupManagedResources(ctx, fw, managedNamespace, appName, componentName, releaseName)
```
avi-biton: Wrong indentation

test/go-tests/tests/conformance/setup.go
```diff
@@ -33,6 +85,79 @@ func runSetupRelease(appName, componentName, tenantNS, managedNS string) error {
...
+	if preProvisioned {
+		klog.Infof("conformance: managed namespace %s is pre-provisioned (E2E_MANAGED_NAMESPACE set), adapting setup-release.sh", managedNS)
+
+		// Give per-run names to shared resources so concurrent test runs
+		// don't clobber each other. Each run owns its own resources and
+		// cleans them up independently.
+		taSuffix := strings.TrimPrefix(releaseName, "release-")
+
+		taName := "trusted-artifacts-" + taSuffix
+		scriptContent = []byte(strings.ReplaceAll(string(scriptContent), "trusted-artifacts", taName))
```
avi-biton: `strings.ReplaceAll` for "trusted-artifacts" replaces every occurrence in the entire script (comments, log messages, etc.), not just the resource name. 
Other replacements in the function use targeted strings.Replace with count=1. 

Suggestion: Use strings.Replace with the expected count, or a regex that targets only the resource definition context.

_Avi's Cursor_

test/go-tests/tests/conformance/setup.go
```diff
@@ -209,6 +507,269 @@ func dumpDiagnostics(hub *framework.ControllerHub, componentName, appName, names
...
+		for _, e := range e2eECPExclusions {
+			if !seen[e] {
+				missing = append(missing, e)
+			}
+		}
+		if len(missing) > 0 {
+			return fmt.Errorf("ECP %s/%s is missing required exclusions %v; EC validation will fail", managedNS, policyName, missing)
+		}
+	}
+	return nil
```
avi-biton, line 838: (verifyECPPatched) Problem: If all policy sources have Config == nil (e.g. patchECPForE2E failed entirely), the function skips every source via continue and returns nil — falsely reporting success. Suggestion: Track whether at least one source was actually verified; return an error if none were.

_Avi's Cursor_

test/go-tests/tests/conformance/conformance.go
```diff
@@ -79,30 +81,74 @@ var _ = ginkgo.Describe("[conformance]", ginkgo.Label(devEnvTestLabel, upstreamK
...
-				managedNamespace = "default-managed-tenant"
+				managedNamespace = os.Getenv("E2E_MANAGED_NAMESPACE")
+				if managedNamespace == "" {
+					managedNamespace = "default-managed-tenant"
+				}
 				klog.Info("conformance: namespaces ready", "user", userNamespace, "managed", managedNamespace)
 
-				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, util.GenerateRandomString(4))
+					suffix := util.GenerateRandomString(4)
+				componentName = fmt.Sprintf("%s-%s", appSpec.ComponentSpec.Name, suffix)
```
avi-biton: Wrong indentaion

Troy876: Fixed

test/go-tests/tests/conformance/setup.go
Troy876: Added a verified flag that tracks whether at least one source with a non-nil Config was processed. If none were, the function now returns an error instead of silently succeeding.

test/go-tests/tests/conformance/setup.go
Troy876: Replaced the single broad strings.ReplaceAll with four targeted replacements that only match functional occurrences (YAML name: fields, image paths, array entries, and kubectl get commands).

avi-biton (approved):

flacatus:
## Additional changes for private repo support

While validating this feature against stone-stage-p01, we found two additions that make the conformance tests work cleanly with private component repositories.

### 1. E2E_REPO env var — override the component source repo

test/go-tests/pkg/constants/constants.go:

```go
GITHUB_E2E_REPO_ENV string = "E2E_REPO" // #nosec

// DefaultIntegrationSRepo is the public repo used by the IntegrationTestScenario git resolver.
// If changed to a private repo, a secret must be created in the namespace and token/tokenKey
// params added to the resolver — see Konflux docs on accessing private repos in ITS.
DefaultIntegrationSRepo string = "testrepo"
```

test/go-tests/tests/conformance/config/scenarios.go — component uses E2E_REPO, ITS always uses the public DefaultIntegrationSRepo:

```go
GitSourceUrl: fmt.Sprintf("https://github.com/%s/%s",
    utils.GetEnv(constants.GITHUB_E2E_ORGANIZATION_ENV, constants.DefaultGitHubE2EOrganization),
    utils.GetEnv(constants.GITHUB_E2E_REPO_ENV, constants.DefaultIntegrationSRepo)),
// ...
GitURL: fmt.Sprintf("https://github.com/%s/%s",
    utils.GetEnv(constants.GITHUB_E2E_ORGANIZATION_ENV, constants.DefaultGitHubE2EOrganization),
    constants.DefaultIntegrationSRepo), // always public — git resolver has no credentials by default
```

Default is testrepo so existing behaviour is unchanged. Set E2E_REPO=testrepo-private to run the build against a private repo while the ITS pipeline is still fetched from the public testrepo.

### Validated on staging

- Component built from konflux-ci/testrepo-private (private)
- **26/26 specs passed** on stone-stage-p01

avi-biton (approved):

Troy876:
@avi-biton Can you add your comments to run the tests please?

flacatus:
@avi-biton I think the tests failures are not related with this PR. I see them everywhere failing

PR #8931, opened 2026-08-24
docs: add single-file verification commands to AGENTS.md

kelchen123:
Add single file verification for AgentReady.


yftacherzog (commented):
@kelchen123 What does this change provide?

amisstea (approved):

PR #8980, opened 2026-09-01
fix: clear BearerTokenFile to prevent in-cluster token override

flacatus:
When running inside a pod with a projected SA token, the Go HTTP transport re-reads the token file and overrides the static BearerToken from the KUBECONFIG. This causes 'failed to get server groups: Unauthorized' errors during API discovery when the projected token's OIDC audience doesn't match what the API server expects for certain operations.

Clear BearerTokenFile when a static BearerToken is already set from the kubeconfig, ensuring the static token is used consistently for all API calls including server group discovery.

### DIFF
--- a/test/go-tests/pkg/clients/kube/client.go
+++ b/test/go-tests/pkg/clients/kube/client.go
@@ -100,6 +100,16 @@ func NewAdminKubernetesClient() (*CustomClient, error) {
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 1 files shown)


yftacherzog (dismissed):

yftacherzog (approved):
@flacatus I'm good with the change.
What about fullsend comments? Do we need the same logic in other places as well for the issue you're trying to solve?

flacatus:
@yftacherzog i think we are good here!

PR #9006, opened 2026-09-02
fix(KONFLUX-15486): add write permission to the CI user

avi-biton:
add write permission to the CI user so it can comment on PRs


avi-biton:
Closed, not needed for now

PR #9016, opened 2026-09-03
feat: build conformance test image

Troy876:
This builds a conformance test image that contains the tests and dependencies for running conformance tests

Jira: https://redhat.atlassian.net/browse/KFLUXDP-1165


yftacherzog (changes requested):
I was not aware this is planned, so we'll need to evaluate this in the team to make sure no high-level issues with introducing this.

In the meantime, I would say that we should onboard the relevant component to downstream Konflux so this runs in CI before merging this.

We should also have in mind multi-arch builds and hermeticity from the start rather than introducing non-hermetic builds in a repo that currently builds hermetically (not sure what it will take).

test/go-tests/Containerfile
```diff
@@ -0,0 +1,54 @@
+# Build the conformance test binary.
+#
+# The test module uses a replace directive (../../operator), so the Docker
+# context MUST be the repository root — not test/go-tests/ alone.
+#
+# Build:  podman build -f test/go-tests/Containerfile -t conformance-tests .
+FROM registry.access.redhat.com/ubi9/go-toolset:1.26 AS builder
```
yftacherzog: Any specific reason to use ubi9 rather than ubi10?

test/go-tests/Containerfile
```diff
@@ -0,0 +1,54 @@
...
+# Runtime image — no Go toolchain, no source code.
+# -------------------------------------------------------------------
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+
+RUN microdnf install -y tar gzip && \
+    curl -sSL "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz" \
+      | tar xz -C /usr/local/bin oc kubectl && \
+    chmod +x /usr/local/bin/oc /usr/local/bin/kubectl && \
+    curl -sSL "https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-amd64" \
+      -o /usr/local/bin/jq && chmod +x /usr/local/bin/jq && \
```
yftacherzog: I don't like the idea of us downloading stuff from the internet. Need to come up with an alternative. Maybe can get this from task-runner (not sure if it's better/perfect either).

Regardless of how we get it, we need a way to have renovate track those dependencies so they don't go stale.

test/go-tests/Containerfile
Troy876: No, I'll change it to ubi10

test/go-tests/Containerfile
Troy876: Some suggestions from Cursor:

> Option A: Use task-runner as runtime base (has everything, but bloated with 20+ unused tools)
> 
> Option B: Use ubi10-minimal as base, COPY binaries from official Red Hat images (minimal, but need to find source image for jq)
> 
> Option C: Use ubi10-minimal as base, install oc/jq via RPM packages (standard, but may need entitlements)
> 
> Option D: Use ubi10-minimal as base, COPY just oc/kubectl/jq from task-runner (minimal image, no downloads, tools come from an image we already maintain)

yftacherzog: Using task-runner as base will be the easiest I suppose. I wonder what's the size difference between using ubi10-minimal + getting just the binaries we need (regardless of the way) and using task-runner.

Troy876: Using ubi10-minimal with just the binaries we need would be roughly 3x smaller (~455 MB vs ~1.41 GB). The task-runner carries ~900 MB of tools we don't need (aws, cosign, syft, buildah, oras, tkn, python, etc.).

yftacherzog: Sounds big :)
What's the reason the conformance image needs oc, jq, etc.?

Troy876: > Only kubectl is truly required (the test code shells out to it). oc is a nice-to-have fallback, jq can be dropped entirely.

yftacherzog: oc is something we can prefetch I think, but requires entitlements as you mentioned. I think that's probably the right way to have it implemented. No?

Troy876: Agreed, currently I'm testing to see if we can run conformance via the image so all going well this won't cause any issue.

yftacherzog: I see now. oc/kubectl are used for running and testing the onboarding scripts that users are expected to be using, so we indeed need either of them (need to test the scripts in the same way that users will use them when they fetch them from the cluster). We can maybe do something to make jq redundant, but if we do the work to have oc/kubectl prefetched, might not take much to have jq fetched as well.

yftacherzog: If it's easier now, we can try and go with copying the binaries from task-runner.

Regardless, a few things I think are worth doing still in this PR to avoid tech debt:
1. Enable multi-arch builds for the new image (following the build patterns for the operator image).
2. Enable hermetic builds for this image (following the build patterns for the operator image).
3. Test this image as part of the build. I think we should be able to do that by making the e2e tests that we're doing on kind in aws (with tekton) use the image instead of what those do at the moment.

Troy876: I believe we can make jq redundant so we can avoid it. Containerfile uses ARG TARGETOS/TARGETARCH for cross-compilation, following the operator pattern so multi arch builds should be enabled. prefetch-input configured for gomod prefetch. CLI binaries (oc, kubectl) copied from digest-pinned task-runner image at build time. And e2e should support images, so if an image isn't set it acts normally. When set, extracts pre-built binary from image (podman locally, kubectl in Tekton) and runs directly instead of go test from source.

.tekton/pipelines/operator-e2e/pipeline.yaml
```diff
@@ -127,6 +127,12 @@ spec:
         Optional extra arguments appended to conformance go test after fixed Ginkgo flags
         (e.g. -ginkgo.focus=Name).
       default: ""
+    - name: conformance-image
```
yftacherzog, line 130: Can we have it set to `quay.io/redhat-user-workloads/konflux-vanguard-tenant/conformance-tests:on-pr-{{revision}}` in the pac-triggered pipelinerun? Would it work?

Troy876: It appears so yes, however Cursor raises a timing issue; might be something, might be nothing:

> One timing note: conformance image build (conformance-tests-on-pull-request) and e2e pipeline run in parallel from same PR event. But e2e spends ~10min on Kind provisioning + Konflux deploy before reaching conformance step — conformance image build should finish well before then.
> 
> If timing ever races, kubectl run in extraction step will retry/wait for image pull (120s timeout on pod ready).

yftacherzog: It should not be an issue. Provisioning takes a long time usually.

scripts/operator-e2e/run-conformance-tests.sh
```diff
@@ -14,12 +23,44 @@ export MY_GITHUB_ORG="${GH_ORG:?GH_ORG required}"
...
+
+  if command -v podman &>/dev/null; then
+    CID=$(podman create "${CONFORMANCE_IMAGE}")
+    podman cp "${CID}:/usr/local/bin/conformance.test" "${TMPBIN}/conformance.test"
+    podman rm "${CID}" >/dev/null
+  elif command -v kubectl &>/dev/null; then
+    POD_NAME="conformance-extract-$$"
+    kubectl run "${POD_NAME}" --image="${CONFORMANCE_IMAGE}" --restart=Never \
+      --command -- sleep 300
+    kubectl wait --for=condition=Ready "pod/${POD_NAME}" --timeout=120s
```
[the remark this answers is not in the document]
yftacherzog: We should not have fallbacks.
We should find a way that works and stick to it.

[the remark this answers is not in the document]
Troy876: Removed fallback

avi-biton:
The Containerfile is a reasonable start for a shared conformance runner, but CI does not actually run the image.

`run-conformance-tests.sh` starts a pod with `--command -- sleep 300` (so the entrypoint never runs), `cat`s `/usr/local/bin/conformance.test` out, deletes the pod, and executes the binary on the Tekton host. `oc`/`kubectl` in the image are unused. That treats the image as a zip file, not as a test environment.

That is not a contract other teams can copy. They should be able to pull a pin and run the container against their cluster (`podman run` / Tekton step / Prow `from:`), with kubeconfig and `GITHUB_TOKEN` / `MY_GITHUB_ORG` injected.


_Avi's Cursor_

yftacherzog:
> The Containerfile is a reasonable start for a shared conformance runner, but CI does not actually run the image.
> 
> `run-conformance-tests.sh` starts a pod with `--command -- sleep 300` (so the entrypoint never runs), `cat`s `/usr/local/bin/conformance.test` out, deletes the pod, and executes the binary on the Tekton host. `oc`/`kubectl` in the image are unused. That treats the image as a zip file, not as a test environment.
> 
> That is not a contract other teams can copy. They should be able to pull a pin and run the container against their cluster (`podman run` / Tekton step / Prow `from:`), with kubeconfig and `GITHUB_TOKEN` / `MY_GITHUB_ORG` injected.
> 
> _Avi's Cursor_

@Troy876 @avi-biton @gbenhaim I think we should decided go/no-go on the general approach, and then onboard that component so the PR changes run in the PR's CI.
I think we should avoid proceeding troubleshooting until CI is testing the changes.

yftacherzog:
@Troy876 I think the next thing to do is to open an Mr to onboard this component to downstream konflux. Without it, we're shooting in the dark here.

Troy876:
@yftacherzog  So I got CI working and I've locally tested the build/push script and I've confirmed the image works for conformance tests. I was also able to fully remove the jq dependency from conformance tests. Let me know what you think and if we need to change anything. :smile:

yftacherzog:
> @yftacherzog  So I got CI working and I've locally tested the build/push script and I've confirmed the image works for conformance tests. I was also able to fully remove the jq dependency from conformance tests. Let me know what you think and if we need to change anything. :smile: 

Nice! I'll be off in the coming week so I'll let @avi-biton @flacatus and @gbenhaim review. Can you squash the commits?

Troy876:
> > @yftacherzog  So I got CI working and I've locally tested the build/push script and I've confirmed the image works for conformance tests. I was also able to fully remove the jq dependency from conformance tests. Let me know what you think and if we need to change anything. 😄
> 
> Nice! I'll be off in the coming week so I'll let @avi-biton @flacatus and @gbenhaim review. Can you squash the commits?

Yup, I believe it's just the 1 commit, do want me to resolve all the comments too?

yftacherzog:
> > > @yftacherzog  So I got CI working and I've locally tested the build/push script and I've confirmed the image works for conformance tests. I was also able to fully remove the jq dependency from conformance tests. Let me know what you think and if we need to change anything. 😄
> > 
> > Nice! I'll be off in the coming week so I'll let @avi-biton @flacatus and @gbenhaim review. Can you squash the commits?
> 
> Yup, I believe it's just the 1 commit, do want me to resolve all the comments too?

Cool! Was looking through the mobile app so did not see the commit count and relied on what I saw earlier 😀

flacatus (approved):

operator/upstream-kustomizations/cli/setup-release.sh
```diff
@@ -218,9 +218,16 @@ fi
...
+CONFORMA_POLICY_SPEC=$(kubectl get enterprisecontractpolicy "${CONFORMA_POLICY}" -n enterprise-contract-service \
+    -o jsonpath='{.spec}')
+kubectl apply -f - <<EOF
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: EnterpriseContractPolicy
+metadata:
+  name: ${CONFORMA_POLICY}
+  namespace: ${MANAGED_NS}
+spec: ${CONFORMA_POLICY_SPEC}
+EOF
```
avi-biton, line 230: The jq-based ECP rename in pre-provisioned mode no longer matches `setup-release.sh`.

This PR replaced the jq pipeline in `operator/upstream-kustomizations/cli/setup-release.sh` (lines 221–230) with a heredoc:

```bash
CONFORMA_POLICY_SPEC=$(kubectl get enterprisecontractpolicy ... -o jsonpath='{.spec}')
# ...
metadata:
  name: ${CONFORMA_POLICY}
spec: ${CONFORMA_POLICY_SPEC}
```

But `runSetupRelease` in `test/go-tests/tests/conformance/setup.go` (lines 122–126) still patches the old jq snippet:

```go
oldJQ := `.metadata.namespace = "'"${MANAGED_NS}"'"'`
newJQ := `.metadata.namespace = "'"${MANAGED_NS}"'" | .metadata.name = "` + ecpName + `"'`
scriptContent = []byte(strings.Replace(string(scriptContent), oldJQ, newJQ, 1))
scriptContent = []byte(strings.Replace(string(scriptContent), "policy: ${CONFORMA_POLICY}", "policy: "+ecpName, 1))
```

`strings.Replace` is a silent no-op when `oldJQ` is gone. The RPA line at `setup-release.sh:405` (`policy: ${CONFORMA_POLICY}`) still matches, so staging mode (`E2E_MANAGED_NAMESPACE` set) will create an ECP named `default` while pointing the ReleasePlanAdmission at `ecp-<suffix>`.

Kind e2e does not set `E2E_MANAGED_NAMESPACE`, so the green PAC run would not catch this. That is the path this image is being built for.

Please patch the new heredoc (`name: ${CONFORMA_POLICY}` → `name: <ecpName>`) instead of the removed jq, and fail the test if the ECP rename does not match rather than continuing.

_Posted by Avi's Cursor_

scripts/operator-e2e/run-conformance-tests.sh
```diff
@@ -14,12 +21,93 @@ export MY_GITHUB_ORG="${GH_ORG:?GH_ORG required}"
...
-  ${E2E_CONFORMANCE_GO_TEST_EXTRA_ARGS:-}
+if [[ -n "${E2E_CONFORMANCE_IMAGE:-}" ]]; then
+  echo "Running conformance tests from image: ${E2E_CONFORMANCE_IMAGE}"
+
+  POD_NAME="conformance-runner-$$"
+  CREDS_SECRET="conformance-creds-$$"
+  KC_FILE="${KUBECONFIG:-$HOME/.kube/config}"
+  # Namespace pinned explicitly (kubeconfig default may not exist on target cluster);
+  # creds passed via Secret, not env literals, to keep tokens out of `kubectl get pod -o yaml`;
+  # --overrides is a JSON merge patch, so every container field must be declared explicit.
```
avi-biton, line 32: `--overrides` is likely dropping the container args.

The comment at `scripts/operator-e2e/run-conformance-tests.sh:32` is correct: `kubectl run --overrides` is a JSON merge patch, so every container field has to be declared on the object. The override at lines 47–69 sets `name`, `image`, `env`, and `volumeMounts`, but not `args`. The flags at lines 72–76 are passed as `kubectl run ... --`:

```bash
    --overrides="$(cat <<OVERRIDES
{
  "spec": {
    "containers": [{
      "name": "${POD_NAME}",
      "image": "${E2E_CONFORMANCE_IMAGE}",
      "env": [ ... ],
      "volumeMounts": [ ... ]
    }],
```

```bash
    -- \
    -ginkgo.vv \
    -test.timeout 45m \
    -test.count 1 \
    ${E2E_CONFORMANCE_GO_TEST_EXTRA_ARGS:-}
```

JSON merge patch replaces the `containers` array, so those `--` args are typically discarded. The suite can still pass because `ENTRYPOINT` runs `conformance.test` with defaults, which is probably what Kind just did. Documented `E2E_CONFORMANCE_GO_TEST_EXTRA_ARGS` (`-ginkgo.focus`, etc.) would not be applied.

Please put `args` on the container in the overrides JSON (including the extra-args expansion) instead of relying on `kubectl run --`.

_Avi's Cursor_

operator/upstream-kustomizations/cli/setup-release.sh
Troy876: Removed dead jq lines (oldJQ/newJQ/strings.Replace)
Added fatal bool field to scriptPatch; loop now returns error if a fatal patch misses
Added ecp-rename patch targeting name: ${CONFORMA_POLICY} in the heredoc, with fatal: true

scripts/operator-e2e/run-conformance-tests.sh
Troy876: Pre-computes EXTRA_ARGS_JSON shell fragment before kubectl run
Adds "args": [...] to container in --overrides JSON (includes extra args)
Removes -- -ginkgo.vv ... trailing args (were silently dropped by the merge patch)

Issue #8837, opened 2026-08-13
Move cluster trusted-ca Bundle to KonfluxCertManager with Konflux CR opt-in


yftacherzog:
## Summary

On non-OpenShift clusters (Kind/local), Tekton flows that resolve pipelines or tasks from GitHub via the in-cluster git resolver can fail with:

```
Reason=CouldntGetPipeline
fatal: unable to access 'https://github.com/.../':
  error setting certificate file: /etc/ssl/certs/ca-bundle.crt: exit status 128
```

This happens because `dependencies/tekton-config/tekton-config.yml` configures `tekton-pipelines-remote-resolvers` to set `GIT_SSL_CAINFO=/etc/ssl/certs/ca-bundle.crt` and mount a `trusted-ca` ConfigMap, but the ConfigMap is often never populated during operator-based bootstrap.

Related: #8818

## Root cause

Three pieces are misaligned on the operator / Kind deploy path:

1. **Tekton config** expects a cluster-wide `trusted-ca` ConfigMap (optional mount, but `GIT_SSL_CAINFO` is always set).
2. **`trust-bundle.yaml`** (trust-manager `Bundle` CR) distributes `trusted-ca` ConfigMaps cluster-wide, but today lives under **internal registry** manifests (`operator/upstream-kustomizations/registry/` and legacy `dependencies/registry/`).
3. **Operator bootstrap** (`deploy-local.sh` / `.tekton/tasks/deploy-konflux`) runs `deploy-deps.sh` with `SKIP_INTERNAL_REGISTRY=true`, so the Bundle is never applied unless `internalRegistry.enabled: true` (side effect of the registry reconciler).

The `Bundle` is not registry-specific — it merges public CAs and `konflux-ca-secret` for cluster-wide use (Tekton git resolver, builds, etc.). Tying it to internal registry is the wrong ownership model.

### 1. Konflux CR field

Add an explicit field under `spec.certManager` (name TBD, e.g. `distributeClusterCABundle`) to control whether the operator applies the trust-manager `Bundle` that populates cluster-wide `trusted-ca` ConfigMaps.

**Not** tied to `internalRegistry.enabled` or `createClusterIssuer`.

### 2. Defaults when field is omitted

Apply platform-aware defaults in the reconciler (via `ClusterInfo`), not CRD zero-value:

| Platform | Default |
|----------|---------|
[...]

PR #9064, opened 2026-09-06
fix(operator): reconcile internal registry trust Bundle

AyushSrivastava1818:
## What does this PR do?

Adds reconciliation for the trust-manager `Bundle` managed by the `KonfluxInternalRegistry` controller.

The controller already applies the `trusted-ca` Bundle as part of the internal registry resources, but changes or deletion of the Bundle did not trigger reconciliation because the controller could not directly import the trust-manager Go API due to dependency/version compatibility constraints.

This change adds a dynamic `unstructured.Unstructured` watch for the `trust.cert-manager.io/v1alpha1` `Bundle` resource without introducing a new Go dependency.

## Why is this needed?

Without a Bundle watch, out-of-band changes such as:

* deletion of the managed Bundle;
* modification of its specification; or
* removal of its Konflux ownership labels

could leave the resource in a drifted or missing state until another event caused the `KonfluxInternalRegistry` to reconcile.

The controller now observes changes to the managed Bundle and restores the desired state.

## What changed?

* Added dynamic discovery and watch registration for the trust-manager `Bundle` GVK.
* Register the watch only when the Bundle resource is available in the cluster.
* Ignore status-only Bundle updates.
* Use controller ownership to ensure unrelated Bundles do not trigger reconciliation.
* Added envtest support for the trust-manager Bundle CRD.
* Updated third-party manifest extraction and verification scripts to keep the CRD synchronized.
* Added regression coverage for Bundle creation, deletion/recreation, spec drift, label drift, ownership filtering, and missing-CRD behavior.

## Testing

The following checks were run:

* `go test -v -count=1 ./operator/internal/controller/internalregistry/...`
* `go vet ./...`
* `git diff main...HEAD --check`

The internal registry controller test suite passes, including the new Bundle reconciliation tests.

## Notes

No new Go dependencies were introduced.

The implementation uses the existing dynamic CRD watch approach used elsewhere in the operator to avoid the trust-manager dependency compatibility issue.

### DIFF
--- a/.github/scripts/update-third-party-manifests.sh
+++ b/.github/scripts/update-third-party-manifests.sh
@@ -84,6 +84,11 @@ helm template trust-manager jetstack/trust-manager \
   --set 'defaultPackage.resources.limits.memory=250Mi' \
   > dependencies/trust-manager/trust-manager.yaml
 
+# Extract trust-manager CRDs for envtest (only CustomResourceDefinition documents)
+TRUST_CRD_DIR="operator/test/crds/trust-manager"
+mkdir -p "$TRUST_CRD_DIR"
+yq 'select(.kind == "CustomResourceDefinition")' dependencies/trust-manager/trust-manager.yaml > "$TRUST_CRD_DIR/trust-manager.crds.yaml"
+
 # prometheus-operator ServiceMonitor CRD (Kind envtest + deploy-deps prerequisite)
 PROM_TAG="${PROMETHEUS_OPERATOR_VERSION#v}"
 PROM_CRD_URL="https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v${PROM_TAG}/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml"
@@ -104,4 +109,5 @@ cp -f "$PROM_ENVTEST_CRD" "$PROM_DEPS_CRD"
 echo "Generated dependencies/cert-manager/cert-manager.yaml (cert-manager $CERT_MANAGER_VERSION)"
 echo "Generated $CRD_DIR/cert-manager.crds.yaml (extracted CRDs for envtest)"
 echo "Generated dependencies/trust-manager/trust-manager.yaml (trust-manager $TRUST_MANAGER_VERSION)"
+echo "Generated $TRUST_CRD_DIR/trust-manager.crds.yaml (extracted CRDs for envtest)"
 echo "Generated $PROM_ENVTEST_CRD (prometheus-operator ServiceMonitor CRD $PROMETHEUS_OPERATOR_VERSION)"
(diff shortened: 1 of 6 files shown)


yftacherzog:
I'm not sure it will make sense to address this issue before addressing https://github.com/konflux-ci/konflux-ci/issues/8837

AyushSrivastava1818:
@yftacherzog Thanks for the review. I’ve addressed the three findings:

* Updated the setup test so it actually covers the case where the trust-manager CRD is not available.
* Moved the Bundle GVK and helper function to follow the existing code organization.
* Updated the unrelated Bundle cleanup to use `testutil.DeleteAndWait`.

I also ran the internal registry controller tests, `go vet ./...`, and `git diff --check`; everything passes.

I’m aware of #8837/#8838 and the proposed move of the trust-manager Bundle to `KonfluxCertManager`. Since #8838 hasn’t merged yet, I’ve kept this PR focused on fixing the current reconciliation issue in `KonfluxInternalRegistry`.

If the preferred approach is to wait for #8838 first, I’m happy to adjust this work accordingly.

yftacherzog:
> If the preferred approach is to wait for #8838 first, I’m happy to adjust this work accordingly.

I'm not sure it's necessarily going to be #8838, but we're planning to have that addressed in the next week or two.
Trying to get this one here merged before that will cause some redundant work for sure.

cc @avi-biton

yftacherzog:
> @yftacherzog Thanks for the review. I’ve addressed the three findings:
> 
> * Updated the setup test so it actually covers the case where the trust-manager CRD is not available.
> * Moved the Bundle GVK and helper function to follow the existing code organization.
> * Updated the unrelated Bundle cleanup to use `testutil.DeleteAndWait`.
> 
> I also ran the internal registry controller tests, `go vet ./...`, and `git diff --check`; everything passes.
> 
> I’m aware of #8837/#8838 and the proposed move of the trust-manager Bundle to `KonfluxCertManager`. Since #8838 hasn’t merged yet, I’ve kept this PR focused on fixing the current reconciliation issue in `KonfluxInternalRegistry`.
> 
> If the preferred approach is to wait for #8838 first, I’m happy to adjust this work accordingly.

That other issue was just closed. You're welcome to look into this again to see what's now still relevant.

avi-biton:
@yftacherzog @AyushSrivastava1818 
This PR https://github.com/konflux-ci/konflux-ci/pull/9091 got merged, it moves the bundle from the internal registry to cert-manager

AyushSrivastava1818:
Thanks for the review and guidance on this.

This PR was intended as an immediate fix for the trust-manager Bundle reconciliation gap in `KonfluxInternalRegistry`. Since then, the architectural change discussed in #8837/#8838 has landed in #9091 and moved `trusted-ca` Bundle ownership to `KonfluxCertManager`.

The merged implementation now provides the Bundle watch/self-healing behavior in the new owner, while also handling adoption of existing InternalRegistry-owned Bundles and cleanup when distribution is disabled.

As a result, the reconciliation changes in this PR are no longer needed and would be redundant with the new architecture.

I’m closing #9064 in favor of #9091. Thanks for the review and for pointing me toward the longer-term ownership change.

PR #9076, opened 2026-09-07
fix: restrict manager-role RBAC escalation to named ClusterRoles

yftacherzog:
Remove unscoped escalate on clusterroles and unscoped bind on clusterrolebindings from manager-role, which together allowed cluster-admin takeover if the operator SA was compromised. Scope escalate to explicit ClusterRole resourceNames for KonfluxRBAC, integration-service, and enterprise-contract reconcilers. Add manager-role contract tests against generated role.yaml and embedded manifests to catch both over-permissive and under-tight RBAC.

Assisted-by: Cursor


operator/internal/controller/integrationservice/konfluxintegrationservice_controller.go
```diff
@@ -132,7 +132,7 @@ type KonfluxIntegrationServiceReconciler struct {
...
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,resourceNames=integration-service-componentgroup-admin-role;integration-service-componentgroup-editor-role;integration-service-componentgroup-viewer-role;integration-service-integrationtestscenario-admin-role;integration-service-integrationtestscenario-editor-role;integration-service-integrationtestscenario-viewer-role;integration-service-manager-role;integration-service-metrics-auth-role;integration-service-metrics-reader;integration-service-nudgeconfig-admin-role;integration-service-nudgeconfig-editor-role;integration-service-nudgeconfig-viewer-role;integration-service-snapshot-garbage-collector;integration-service-tekton-editor-role;konflux-integration-runner,verbs=bind;escalate
```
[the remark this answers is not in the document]
yftacherzog: This is something that we will need to address later. The scope of the issue that this change is addressing is:

Before: stolen operator token → new cluster-admin ClusterRole + ClusterRoleBinding with any names.

After: stolen operator token → must hijack a whitelisted, operator-managed role (and then bindings) — cannot mint a generic backdoor role anymore; further hardening (scoped binding CRUD, admission policy) targets that remaining path.

operator/internal/rbac/embedded_clusterroles.go
```diff
@@ -0,0 +1,92 @@
...
+	"k8s.io/apimachinery/pkg/util/yaml"
+
+	"github.com/konflux-ci/konflux-ci/operator/pkg/manifests"
+)
+
+// EmbeddedClusterRoles returns ClusterRole names from a component's embedded manifest.
+func EmbeddedClusterRoles(component manifests.Component) ([]string, error) {
+	content, err := manifests.GetManifest(component)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
yftacherzog: done

gbenhaim (commented):

operator/internal/rbac/konflux_managed_clusterroles.go
```diff
@@ -0,0 +1,36 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rbac
+
+// KonfluxRBACManagedClusterRoles lists ClusterRoles applied by the KonfluxRBAC controller
+// from operator/pkg/manifests/rbac/manifests.yaml. Entry order is not significant; tests
+// compare sorted names against the embedded manifest and manager-role escalate rules.
+var KonfluxRBACManagedClusterRoles = []string{
```
gbenhaim: **Suggestion**: Drop this hardcoded list. It is a fourth copy of names that already live in the RBAC embedded manifest, the kubebuilder marker, and generated `role.yaml`.

`EmbeddedClusterRoles(manifests.RBAC)` already returns this set. `TestKonfluxRBACManagedClusterRolesMatchEmbeddedManifests` then requires the two to be equal, so the slice is not a filter — KonfluxRBAC applies the whole RBAC manifest. The drift tests exist only to keep the duplicate in sync.

`AllEmbeddedClusterRoles()` is the wrong substitute (it includes every component). Use `EmbeddedClusterRoles(manifests.RBAC)` only.

Concretely:
- Delete this file and `TestKonfluxRBACManagedClusterRolesMatchEmbeddedManifests`.
- Delete `TestManagerRoleAllowsEscalateOnKonfluxRBACManagedClusterRoles` and `TestManagerRoleKonfluxRBACEscalateNamesMatchManagedList`. `TestManagerRoleAllowsEscalateOnEmbeddedClusterRoles` already covers the RBAC component.
- In envtest cleanup, build children from `EmbeddedClusterRoles(manifests.RBAC)` instead of this slice.

Separately, please add a contract that every `escalate`/`bind` `resourceName` on `clusterroles` in `role.yaml` is in `AllEmbeddedClusterRoles()` plus known operator-owned names (`konflux-operator-metrics-reader`). The KonfluxRBAC list cannot catch extras such as `cluster-admin`; `TestManagerRoleDeniesEscalateOnArbitraryClusterRole` uses `evil-cluster-admin` and would still pass.

AGENTS.md
avi-biton: A third test was added, need to update the header

yftacherzog: done

operator/internal/rbac/konflux_managed_clusterroles.go
yftacherzog: Should be addressed now.

avi-biton (approved):

PR #9077, opened 2026-09-07
fix(KONFLUX-15487): remove dead impersonate and LSAR from manager-role

yftacherzog:
Proxy SA keeps impersonate/LSAR via konflux-proxy; manager only needs escalate on that ClusterRole to apply it. Operator Tests exercise UI reconcile; Operator E2E and Tekton e2e proxy tests catch deploy or runtime proxy/auth regressions.

Assisted-by: Cursor

### DIFF
--- a/operator/config/rbac/role.yaml
+++ b/operator/config/rbac/role.yaml
@@ -27,13 +27,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - groups
-  - users
-  verbs:
-  - impersonate
-- apiGroups:
-  - ""
-  resources:
   - namespaces
   - serviceaccounts
   - services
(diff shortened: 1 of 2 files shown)


gbenhaim (approved):
LGTM. Removing unused `impersonate` / LSAR from `manager-role` is the right least-privilege cut: the manager never used those verbs, they stay on `konflux-proxy`, and the UI reconciler still has named `bind`/`escalate` to apply that ClusterRole. In-cluster Operator E2E (`make deploy`) and Prow already exercised this path.

This is only a slice of KONFLUX-15487 — unscoped `escalate`, secrets CRUD, and webhook/SCC breadth are still open on that ticket.

PR #9176, opened 2026-09-13
Added support for automatic rotation of oauth2-proxy client secret in the operator.

### LINKED ISSUE #7024: Support automatic rotation of oauth2-proxy client secret in the operator
## Description

When the `oauth2-proxy-client-secret` Secret value changes, the Dex and proxy deployments don't pick up the new value because Kubernetes doesn't restart pods when a Secret referenced by `env.valueFrom.secretKeyRef` changes. The operator should support automatic rotation by detecting Secret content changes and triggering a deployment rollout.

## Current Behavior

The `CLIENT_SECRET` env var in the Dex deployment is sourced from the `oauth2-proxy-client-secret` Secret via `secretKeyRef`. If the Secret value is updated, existing pods continue using the stale value until manually restarted.

## Desired Behavior

The operator should detect changes to the `oauth2-proxy-client-secret` Secret and trigger a rolling restart of the affected deployments (Dex and proxy). This could be implemented by hashing the Secret content into a pod annotation (similar to how the Dex ConfigMap already uses content-based hash naming), so that Secret changes automatically trigger a new rollout.

## Context

- The operator already handles Dex config changes via content-hashed ConfigMap names (`hashedconfigmap`)
- The same pattern could be applied to Secrets: hash the Secret content into a deployment annotation, causing Kubernetes to roll out new pods when the hash changes
- This is the Kubernetes-native approach and provides rolling updates, auditability, and works regardless of what changed

_This issue was created by **Gal's Cursor**._

saikat709:
## Summary

Fixes #7024

When the `oauth2-proxy-client-secret` Secret is rotated, the `dex` and
`proxy` deployments continue using the stale value because Kubernetes
does not restart pods when a Secret is consumed via
`env.valueFrom.secretKeyRef`.

## Solution

Injected a content hash of the `oauth2-proxy-client-secret` as a
pod-template annotation on both deployments:

konflux.dev/oauth2-proxy-client-secret-hash:

When the Secret changes, the hash changes → Kubernetes sees a new pod
template spec → rolling restart is triggered automatically.
## Changes
- Added `oauth2ProxyClientSecretHashAnnotation` constant
- Added `reconcileOAuth2ProxyClientSecretHash()` — reads the Secret and
  hashes its data via `contenthash.Map()`
- Moved `ensureUISecrets` before `applyManifests` so the hash is
  readable on the same reconcile pass
- Passed hash through `applyManifests` → `applyUIDeploymentCustomizations`
  → injected into pod template annotations on both `dex` and `proxy`
- Added `mapOAuth2ProxyClientSecretToUI` mapper and a `Watches` entry in
  `SetupWithManager` so Secret changes trigger a reconcile
## Tests
- **Unit** (`TestApplyUIDeploymentCustomizations_ClientSecretHash`):
  verified annotation is set on proxy and dex, omitted when hash is
  empty, and updated when hash changes
- **Integration** (`Context("oauth2-proxy-client-secret rotation")`):
  full envtest coverage of initial annotation set, rotation triggering
  annotation update, and stability when Secret is unchanged


operator/internal/controller/ui/konfluxui_controller.go
```diff
@@ -877,6 +953,15 @@ func (r *KonfluxUIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			builder.WithPredicates(predicate.IgnoreStatusUpdatesPredicate)).
+		// Watch the oauth2-proxy-client-secret Secret so that rotations trigger a reconcile
+		// and the pod-template hash annotation is updated on both dex and proxy deployments.
+		Watches(&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(r.mapOAuth2ProxyClientSecretToUI),
+			builder.WithPredicates(
+				crpredicate.NewPredicateFuncs(func(o client.Object) bool {
+					return o.GetNamespace() == uiNamespace && o.GetName() == oauth2ProxyClientSecretName
+				}),
+			)).
```
yftacherzog, line 964: Is this required given that we have `Owns(&corev1.Secret{})` above?

yftacherzog:
We need patch code coverage to be above 80%. See https://github.com/konflux-ci/konflux-ci/pull/9176#issuecomment-5655390171

yftacherzog:
The envtest coverage is good for proving the annotation updates, but it doesn't verify the actual failure mode from #7024: Dex and oauth2-proxy continuing to serve traffic with stale `CLIENT_SECRET` env vars after rotation.

Could we add an E2E in `test/go-tests/proxy_test.go` that exercises the real auth path?

**Suggested flow:**
1. `ExtractToken(proxyClient)` + `expectProxyGETWithBearer(...)` — baseline
2. Patch `oauth2-proxy-client-secret` in `konflux-ui` with a new `client-secret` value
3. Wait for `dex` and `proxy` rollouts to complete (deployment status / new pods)
4. `ExtractToken` again (reads the updated secret from the API) + `expectProxyGETWithBearer` again

That would confirm both Dex (token grant) and oauth2-proxy (bearer proxying) picked up the new secret, not just that the pod-template annotation changed.

**Notes:**
- Gate behind an env var (e.g. `KONFLUX_PROXY_TEST_SECRET_ROTATION=1`) since it mutates cluster state
- On OpenShift CI (`KONFLUX_PROXY_AUTH=openshift`), use `obtainOpenShiftProxyIDToken` instead of the Dex password grant path
- `ensureUISecrets` only fills empty keys, so a patched value should persist across reconciles

It would make the fix much easier to trust in real clusters.

— Yftach's Cursor

operator/internal/controller/ui/konfluxui_controller.go
saikat709: Thanks. It's not required. Owns(&corev1.Secret{}) already handles secret updates because ensureUISecrets sets the owner reference on oauth2-proxy-client-secret. I'll remove this Watches block.

saikat709:
Hi, @yftacherzog!

I removed the redundant `Watches` block and helper since `Owns(&corev1.Secret{})` already handles secret updates. I also added unit tests for the hash reconciler and deployment customizations to fix codecov coverage, and added the E2E test in `proxy_test.go` gated behind `KONFLUX_PROXY_TEST_SECRET_ROTATION=1` to test the full auth rollout after secret rotation. 
Looking forward to a second review.

PR #9182, opened 2026-09-14
feat(KFLUXVNGD-1345): add trustedCA field to mount custom CA bundle

avi-biton:
The build-service controller-manager ships with a trusted-ca volume
that mounts an optional ConfigMap at /etc/ssl/certs/ca-custom-bundle.crt.
That extra-file hook already adds CAs to the image trust store without
replacing it. The remaining gap is which ConfigMap is mounted: the
volume hardcodes trusted-ca / ca-bundle.crt, which a trust-manager
Bundle in the build-service namespace often cannot own (name or key
clash).

Add an optional `trustedCA` field to `KonfluxBuildServiceConfigSpec`
that accepts a ConfigMap name and key in the build-service namespace.
When set, the controller points the existing extra-file volume at that
ConfigMap (optional: false). The user's key is projected as
ca-bundle.crt so the baked-in subPath still matches. The mount path is
unchanged; extra CAs are added and the image trust store is not
replaced. The extra file is exactly the referenced key, not a merge
with ConfigMap trusted-ca. On OpenShift the platform-injected bundle
is no longer mounted while trustedCA is set.

When omitted, the volume stays optional and points at ConfigMap
trusted-ca / key ca-bundle.crt.

subPath mounts do not pick up ConfigMap data changes, and Go's
SystemCertPool is loaded once per process. The controller Watches the
user-owned ConfigMap (Owns() does not see it) and stamps
konflux.konflux-ci.dev/trusted-ca-content-hash on the pod template so
a content change rolls the Deployment. pkg/hashedconfigmap cannot be
used here: it renames operator-owned objects, and this ConfigMap is
user-owned.

Key decisions:
- Keep the extra-file mount. Overlaying the UBI system bundle would
  replace public CAs and break TLS if the ConfigMap holds only extras.
- Put every CA that should appear in the extra file into the
  referenced ConfigMap. Public CAs from the image remain.
- The ConfigMap must live in the build-service namespace; this change
  does not mount cross-namespace objects.
- Contents must be PEM; the key name does not have to end in .pem.
- The TrustedCAConfigMap type lives in common_types.go for reuse
  by other components that have similar CA mount patterns.
- Name is a DNS-1123 subdomain (max 253 chars). Key is a Kubernetes
  ConfigMap key ([-._a-zA-Z0-9]+, max 253 chars).
- optional: false is set intentionally — if the operator explicitly
  declared a CA bundle, a missing ConfigMap or missing key keeps
  the pod Pending with a standard Kubernetes error. Reconcile still
  completes in those cases (orphan cleanup, status updates). Any
  other Get error fails the apply so server-side apply cannot prune
  the hash annotation and roll the controller-manager.
- A runtime filepath.Base check on the key provides defense-in-depth
  against path traversal (rejects "." and ".."), complementing the
  CRD-level regex validation.

Also includes:
- Extract FindContainer, FindVolume, and FindVolumeMount into
  pkg/kubernetes/ as shared helpers (using slices.IndexFunc),
  removing duplicates from testutil and controller packages.
  Migrated all callers across 8 controller packages and 15+
  test files.
- Unit tests for applyTrustedCAMount (omit, retarget, hash stamp
  and leftover removal, default ConfigMap name, other volumes
  preserved, invalid keys, special-character keys, missing volume,
  non-ConfigMap volume, missing manager container, missing mount),
  plus trustedCABundleHash, lookupTrustedCAHash, and
  mapTrustedCAConfigMap.
- Envtest for omit, retarget, restore, A-to-B retarget, missing
  ConfigMap still reconciles, ConfigMap data change updates the
  hash annotation, and CRD validation of Name (DNS-1123, length)
  and Key (length, invalid ConfigMap keys).
- How-to guide: docs/content/docs/guides/custom-ca-bundle.md.

Assisted-by: Cursor + Grok 4.6


operator/internal/controller/buildservice/konfluxbuildservice_controller_test.go
```diff
@@ -267,6 +270,212 @@ var _ = Describe("KonfluxBuildService Controller", func() {
...
+
+		BeforeEach(func() {
+			startManagerWithClusterInfo(nil)
+			buildService = newBuildServiceCR()
+			Expect(k8sClient.Create(ctx, buildService)).To(Succeed())
+			Eventually(getDeployment).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Not(BeNil()))
+		})
+
+		AfterEach(func() {
+			testutil.DeleteAndWait(ctx, k8sClient, buildService)
```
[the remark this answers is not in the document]
avi-biton: Fixed

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -195,6 +209,12 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
 		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "reconcile webhook config")
 	}
 
+	// Validate that the trustedCA ConfigMap exists and contains the expected key
+	// before applying manifests with optional: false.
+	if err := common.ValidateTrustedCAConfigMap(ctx, r.Client, buildService.Spec.TrustedCA, webhookConfigNamespace); err != nil {
+		return errHandler.HandleWithReason(ctx, err, condition.ReasonConfigMapFailed, "validate trustedCA ConfigMap")
```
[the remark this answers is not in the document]
avi-biton: Fixed

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,26 @@ const (
 	ZapEncoderArg = "--zap-encoder"
 )
 
+// TrustedCAConfigMap references a ConfigMap that contains a PEM CA bundle.
+// The ConfigMap must exist in the component's namespace.
```
filariow, line 62: Idea for a future improvement: Should we provide an easy way to propagate the ConfigMap via Trust Bundle? I mean, we allow the user to point to a ConfigMap and configure a TrustBundle that will propagate the given ConfigMap to all the Konflux components that reference that ConfigMap.

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,26 @@ const (
...
+//
+// This type is intentionally defined in common_types.go for reuse across
+// component specs (e.g. image-controller, integration-service) that have
+// similar CA mount patterns. When adding trustedCA support to another
+// component, embed a *TrustedCAConfigMap field in its config spec, call
+// common.ValidateTrustedCAConfigMap in the reconciler, and apply the mount
+// customization in the component's deployment overlay.
+type TrustedCAConfigMap struct {
+	// Name is the name of the ConfigMap containing the CA bundle.
+	// +kubebuilder:validation:MinLength=1
```
filariow, line 75: `Name` should be a valid [DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names). We could also add:


```suggestion
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9]$`
```

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,26 @@ const (
...
+	Name string `json:"name"`
+	// Key is the key within the ConfigMap that contains the CA bundle in PEM format.
+	// Must be a plain filename without path separators or directory traversal sequences.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`
```
filariow: checking on a kwokctl cluster, I got the following error:

```
The ConfigMap "mycm" is invalid: data[a-./]: Invalid value: "a-./": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```
should we use that regex instead?


```suggestion
	// +kubebuilder:validation:Pattern=`^[-._a-zA-Z0-9]+$`
```

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -195,6 +209,12 @@ func (r *KonfluxBuildServiceReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+
```
filariow: should we avoid to check if the ConfigMap exists at all?

If the ConfigMap doesn't exist at creation time, the Pod won't progress the `Pending` state and it will show the error about the missing ConfigMap, which is a standard kubernetes error easy to debug.

Also if the ConfigMap is deleted for whatever reason, the reconciliation won't progress this block. Like it won't Cleanup Orphans at L262, not update Component Statuses

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +442,81 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	if spec == nil {
+		return nil
+	}
+	key := spec.Key
+	// Defense-in-depth: the CRD schema regex (^[a-zA-Z0-9][a-zA-Z0-9._-]*$) is
+	// stricter, but this runtime check guards against bypasses (e.g. direct API
+	// calls that skip webhook validation).
+	if filepath.Base(key) != key || key == "." || key == ".." {
+		return fmt.Errorf("invalid CA bundle key %q: must be a plain filename without path separators or traversal sequences", key)
+	}
```
filariow: IIUC, kubernetes is already protecting against `..`, cf https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#keytopath-v1-core

> path is the relative path of the file to map the key to. May not be an absolute path. May not contain the path element '..'. May not start with the string '..'.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +442,81 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	if vol == nil || vol.ConfigMap == nil {
+		return fmt.Errorf("trusted-ca ConfigMap volume not found on deployment %s", deployment.Name)
+	}
+	vol.ConfigMap.Name = spec.Name
+	vol.ConfigMap.Items = []corev1.KeyToPath{{
+		Key:  key,
+		Path: trustedCASystemBundleFile,
+	}}
+	vol.ConfigMap.Optional = ptr.To(false)
```
filariow, line 463: as we are using go 1.26, we should be able to use `new(false)`. `go fix ./...` should be able to fix this automatically.


```suggestion
	vol.ConfigMap.Optional = new(false)
```

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +442,81 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	var mount *corev1.VolumeMount
+	for i := range manager.VolumeMounts {
+		if manager.VolumeMounts[i].Name == trustedCAVolumeName {
+			mount = &manager.VolumeMounts[i]
+			break
+		}
+	}
+	if mount == nil {
+		return fmt.Errorf("trusted-ca volume mount not found on container %q", buildManagerContainerName)
+	}
```
filariow: If you prefer, this can be simplified as follows.

```suggestion
	i := slices.IndexFunc(manager.VolumeMounts, func(m corev1.VolumeMount) bool {
		return m.Name == trustedCAVolumeName
	})
	if i == -1 {
		return fmt.Errorf("trusted-ca volume mount not found on container %q", buildManagerContainerName)
	}
	mount := &manager.VolumeMounts[i]
```

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +442,81 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	mount.MountPath = trustedCASystemMountPath
+	mount.SubPath = ""
+	mount.ReadOnly = true
+	return nil
```
filariow, line 477: this change is based on side-effect. Personally I find it not intuitive to understand.
I'd suggest changing the approach with a retrieve, update, insert or at least documenting the behavior in a comment

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +442,81 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+
+// mapTrustedCAConfigMap maps a ConfigMap event to the singleton build-service
+// reconcile request when the ConfigMap matches the currently-referenced
+// trustedCA name. Owned ConfigMaps (which have our owner label) are skipped
+// because they are already handled by the Owns(&corev1.ConfigMap{}) watch.
+func (r *KonfluxBuildServiceReconciler) mapTrustedCAConfigMap(ctx context.Context, obj client.Object) []reconcile.Request {
+	// Skip owned ConfigMaps — they are handled by the Owns() watch.
+	if _, ok := obj.GetLabels()[constant.KonfluxOwnerLabel]; ok {
+		return nil
+	}
```
filariow: should we check the owner reference instead? something like


```suggestion
	owned, err := controllerutil.HasOwnerReference(obj.GetOwnerReferences(), obj, r.Scheme)
	if err != nil || owned {
		// err is unexpected, as it's caused by scheme resolution errors.
		// ConfigMaps are registered in the Schema.
		return nil
	}
```

operator/api/v1alpha1/common_types.go
avi-biton: We already provide a way to deploy bundles using trust-manager:

https://github.com/konflux-ci/konflux-ci/blob/main/operator/config/samples/konflux_v1alpha1_konfluxcertmanager.yaml#L14-L19

We can use this method or similar in the future.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Good point - the Kubernetes KeyToPath.Path validation does reject .. and absolute paths. However, the filepath.Base check here is on the Key field (the ConfigMap data key used in Items[].Key), not on the projected Path (which we hardcode to tls-ca-bundle.pem). Kubernetes doesn't apply the same traversal validation on the Key selector.

That said, after relaxing the CRD regex to ^[-._a-zA-Z0-9]+$ (per your earlier suggestion), the runtime check now has a concrete purpose: it catches bare . and .. which the regex allows but are unsafe as filenames. It's 2 lines of defense-in-depth so I'd prefer to keep it, but happy to drop if you feel strongly.

_Avi's Cursor_

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Although it is supported by go 1.26, the common use in the repository is `ptr.To(false)` so I will not update it in this PR, it may be a good TODO item.

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Good suggestion - using controllerutil.HasOwnerReference would be more robust than label matching for filtering owned ConfigMaps.

This is now moot though: per your earlier suggestion (removing the pre-flight ConfigMap validation), we dropped the mapTrustedCAConfigMap method and its associated Watches entirely. The controller no longer watches unowned ConfigMaps, so this filtering logic no longer exists.

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,26 @@ const (
...
+// This type is intentionally defined in common_types.go for reuse across
+// component specs (e.g. image-controller, integration-service) that have
+// similar CA mount patterns. When adding trustedCA support to another
+// component, embed a *TrustedCAConfigMap field in its config spec, call
+// common.ValidateTrustedCAConfigMap in the reconciler, and apply the mount
+// customization in the component's deployment overlay.
+type TrustedCAConfigMap struct {
+	// Name is the name of the ConfigMap containing the CA bundle.
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
avi-biton: resolved

operator/api/v1alpha1/common_types.go
avi-biton: Rules are now following RFC 1123

operator/api/v1alpha1/common_types.go
avi-biton: Handled

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: This validation was removed, we no longer check for the configmap

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Nice addition, updated

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: A comment was added

avi-biton:
@filariow I addressed all the comments, PTAL again :)

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
```diff
@@ -998,12 +975,294 @@ func TestApplyBuildServiceDeploymentCustomizations_ResourceMerging(t *testing.T)
...
+		g.Expect(vol).NotTo(gomega.BeNil())
+		g.Expect(vol.ConfigMap).NotTo(gomega.BeNil())
+		g.Expect(vol.ConfigMap.Name).To(gomega.Equal(trustedCAVolumeName))
+		g.Expect(vol.ConfigMap.Optional).NotTo(gomega.BeNil())
+		g.Expect(*vol.ConfigMap.Optional).To(gomega.BeTrue())
+		g.Expect(vol.ConfigMap.Items).To(gomega.ConsistOf(corev1.KeyToPath{
+			Key:  trustedCADefaultFileVolumePath,
+			Path: trustedCADefaultFileVolumePath,
+		}))
+	})
```
filariow, line 1032: IIUC how the build-service works, this extra file is never used. Maybe we should consider removing it entirely in a follow-up PR.

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
```diff
@@ -998,12 +975,294 @@ func TestApplyBuildServiceDeploymentCustomizations_ResourceMerging(t *testing.T)
...
+
+		vol := findTrustedCAVolume(deployment)
+		g.Expect(vol).NotTo(gomega.BeNil())
+		g.Expect(vol.ConfigMap).NotTo(gomega.BeNil())
+		g.Expect(vol.ConfigMap.Name).To(gomega.Equal("custom-ca-bundle"))
+		g.Expect(vol.ConfigMap.Optional).NotTo(gomega.BeNil())
+		g.Expect(*vol.ConfigMap.Optional).To(gomega.BeFalse())
+		g.Expect(vol.ConfigMap.Items).To(gomega.ConsistOf(corev1.KeyToPath{
+			Key:  "tls.pem",
+			Path: trustedCASystemBundleFile,
```
filariow: minor-cosmetic: reusing `spec` would make it more intuitive to understand the remapping. It applies also to the other test cases

```suggestion
		g.Expect(vol.ConfigMap.Name).To(gomega.Equal(spec.TrustedCA.Name))
		g.Expect(vol.ConfigMap.Optional).NotTo(gomega.BeNil())
		g.Expect(*vol.ConfigMap.Optional).To(gomega.BeFalse())
		g.Expect(vol.ConfigMap.Items).To(gomega.ConsistOf(corev1.KeyToPath{
			Key:  spec.TrustedCA.Key,
			Path: trustedCASystemBundleFile,
```

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
```diff
@@ -998,12 +975,294 @@ func TestApplyBuildServiceDeploymentCustomizations_ResourceMerging(t *testing.T)
...
+						}},
+						Containers: []corev1.Container{{Name: "other-container"}},
+					},
+				},
+			},
+		}
+		err := applyBuildServiceDeploymentCustomizations(deployment, spec, nil, "")
+		g.Expect(err).To(gomega.HaveOccurred())
+		g.Expect(err.Error()).To(gomega.ContainSubstring("container"))
+		g.Expect(err.Error()).To(gomega.ContainSubstring("not found"))
```
filariow: minor-cosmetic: this can be simplified as 


```suggestion
		g.Expect(err).To(gomega.MatchError(gomega.And(
			gomega.ContainSubstring("container"),
			gomega.ContainSubstring("not found"),
		)))
```

filariow (approved):
just some cosmetic feedback. Thanks for addressing my comments 😄

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
avi-biton: It a good follow up, for now I will leave it as it is to make sure nothing breaks

operator/config/crd/bases/konflux.konflux-ci.dev_konfluxbuildservices.yaml
```diff
@@ -398,6 +398,39 @@ spec:
...
+                  When set, the referenced ConfigMap replaces the image trust store at that
+                  path. The ConfigMap must exist in the build-service namespace and should
+                  contain a full PEM bundle (platform CAs plus any extras).
+                properties:
+                  key:
+                    description: |-
+                      Key is the key within the ConfigMap that contains the CA bundle in PEM format.
+                      Must be a valid ConfigMap key: alphanumeric characters, '-', '_' or '.'.
+                    minLength: 1
+                    pattern: ^[-._a-zA-Z0-9]+$
```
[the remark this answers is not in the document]
avi-biton: runtime already rejects those

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,31 @@ const (
...
+//
+// This type is intentionally defined in common_types.go for reuse across
+// component specs (e.g. image-controller, integration-service) that have
+// similar CA mount patterns. When adding trustedCA support to another
+// component, embed a *TrustedCAConfigMap field in its config spec and apply
+// the mount in that component's deployment overlay. Name and Key are validated
+// by the CRD schema; the overlay must also reject keys whose filepath.Base
+// does not equal the key (path separators, "." and ".."), matching the
+// build-service applyTrustedCAMount check.
+type TrustedCAConfigMap struct {
```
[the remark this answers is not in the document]
avi-biton: No need to rename it

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
avi-biton: Updated

operator/internal/controller/buildservice/konfluxbuildservice_customization_test.go
avi-biton: updated

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -344,6 +359,9 @@ func applyBuildServiceDeploymentCustomizations(deployment *appsv1.Deployment, sp
 		if err := buildBuildControllerManagerOverlay(spec, clusterInfo, webhookConfigMapName).ApplyToDeployment(deployment); err != nil {
 			return err
 		}
+		if err := applyTrustedCAMount(deployment, spec.TrustedCA); err != nil {
```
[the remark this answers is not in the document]
avi-biton: Update the PR description

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,31 @@ const (
...
+	// Must be a valid Kubernetes DNS-1123 subdomain (same rules as ConfigMap names).
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Name string `json:"name"`
+	// Key is the key within the ConfigMap that contains the CA bundle in PEM format.
+	// Must be a valid ConfigMap key: alphanumeric characters, '-', '_' or '.'.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[-._a-zA-Z0-9]+$`
+	Key string `json:"key"`
```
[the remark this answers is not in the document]
avi-biton: Added

gbenhaim:
@avi-biton what is the use case of this change? why would one want to override the system certificates included in the image of the build-service?

In addition, if there is a use case, I think that we would need the same change to other services as well, such as integration and release.

avi-biton:
> @avi-biton what is the use case of this change? why would one want to override the system certificates included in the image of the build-service?
> 
> In addition, if there is a use case, I think that we would need the same change to other services as well, such as integration and release.

The use case is described in https://redhat-internal.slack.com/archives/C0BP4GYQJPQ/p1788941150125799
Its about UBI store location.
Other services like integration are intentionally out of this PR and it is now mentioned in the PR description

gbenhaim (commented):

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +436,47 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	manager := kubernetes.FindContainer(deployment.Spec.Template.Spec.Containers, buildManagerContainerName)
+	if manager == nil {
+		return fmt.Errorf("container %q not found on deployment %s", buildManagerContainerName, deployment.Name)
+	}
+	mount := kubernetes.FindVolumeMount(manager.VolumeMounts, trustedCAVolumeName)
+	if mount == nil {
+		return fmt.Errorf("trusted-ca volume mount not found on container %q", buildManagerContainerName)
+	}
+	// mount points into the VolumeMounts slice; field assignments update the element in place.
+	mount.MountPath = trustedCASystemMountPath
```
gbenhaim: **Concern**: if the existing extra-file mount already works, remounting at the UBI system path is the wrong shape for this feature.

Go does not special-case `ca-custom-bundle.crt`, but it *does* scan `/etc/ssl/certs` and append regular files there. So `/etc/ssl/certs/ca-custom-bundle.crt` can already feed extra CAs without replacing the image trust store.

The real gap is **which ConfigMap** is mounted. On OpenShift the operator creates `trusted-ca` with `inject-trusted-cabundle` (CNO fills `ca-bundle.crt`). A trust-manager Bundle typically cannot own that object (name/key clash, or the source CA lives in another namespace), so users need to point the volume at a different ConfigMap in `build-service`. The baked-in volume still hardcodes `trusted-ca` / `ca-bundle.crt`.

`trustedCA` should retarget `volume.configMap.name` / `items[].key` (and `optional: false` is reasonable) while **keeping** `mountPath: /etc/ssl/certs/ca-custom-bundle.crt` and `subPath: ca-bundle.crt` so extra CAs append.

Overlaying `/etc/pki/ca-trust/extracted/pem` replaces the UBI store. That forces a full PEM bundle and will break public TLS if the trust-manager ConfigMap only contains the extra CA.

Suggested change: keep the extra-file mount; only swap the ConfigMap reference:

```go
vol.ConfigMap.Name = spec.Name
vol.ConfigMap.Items = []corev1.KeyToPath{{
    Key:  key,
    Path: trustedCADefaultFileVolumePath, // ca-bundle.crt
}}
vol.ConfigMap.Optional = ptr.To(false)
// leave mount.MountPath / SubPath unchanged
```

filariow: > Concern: if the existing extra-file mount already works, remounting at the UBI system path is the wrong shape for this feature.

Either I don't understand the assumption or I think it's wrong. As of today it's not working. 

> On OpenShift the operator creates trusted-ca with inject-trusted-cabundle (CNO fills ca-bundle.crt). A trust-manager Bundle typically cannot own that object (name/key clash, or the source CA lives in another namespace), so users need to point the volume at a different ConfigMap in build-service.

With the proposed change, the user will be able to define a ConfigMap with a different name from which to load the CA Bundle. Also, if trust-manager is used, it can be configured [to inject default CAs](https://cert-manager.io/docs/trust/trust-manager/) already:

```yaml
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  name: my-org.com  # The bundle name will also be used for the target
spec:
  sources:
  # Include a bundle of publicly trusted certificates which can be
  # used to validate most TLS certificates on the internet, such as
  # those issued by Let's Encrypt, Google, Amazon and others.
  # NB: Unless you pin the package providing these CAs, the default can change!
  - useDefaultCAs: true
```

---

AIUI, The UBI10 base image already ships the `/etc/pki/tls/certs/ca-bundle.crt` CA Bundle. This is path by Go's runtime and if the file is found, [Go skips the directory scan](https://go.dev/src/crypto/x509/root_linux.go). So we need to overwrite, move, or remove it to enable scanning the directory.

```golang
// Possible certificate files; stop after finding one.
var certFiles = []string{
```

```bash
podman run --rm --name test-ca registry.access.redhat.com/ubi10/ubi-minimal:1787204571@sha256:07ab7c0a596785afa1de1963c1cac07f2b9df2e762b9d20d3433cff18ac1af18 bash -c 'ls /etc/pki/tls/certs/ca-bundle.crt'
/etc/pki/tls/certs/ca-bundle.crt
```

Alternatively we could use [SSL_CERT_DIR](https://pkg.go.dev/crypto/x509#SystemCertPool), which however have some implications on macOS and Windows.

> The environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default locations for the SSL certificate file and SSL certificate files directory, respectively. The latter can be a colon-separated list, or a semicolon-separated list on Windows. On platforms which have system APIs for certificate verification (macOS and Windows), setting SSL_CERT_FILE or SSL_CERT_DIR will prevent those APIs from being used, unless the x509sslcertoverrideplatform=0 GODEBUG setting is used. (This changed in Go 1.27.)

gbenhaim: @filariow On the Go loading model: finding a bundle file does **not** skip the directory scan.

Go does not detect the distro. `crypto/x509.loadOnDiskRoots` uses two hardcoded lists ([`root_linux.go`](https://go.dev/src/crypto/x509/root_linux.go) / [`root.go`](https://cs.opensource.google/go/go/+/master:src/crypto/x509/root.go)):

**1. Bundle files — first hit wins.** It tries these paths in order and `break`s out of *this loop only* on the first `ReadFile` success:

1. `/etc/ssl/certs/ca-certificates.crt`
2. `/etc/pki/tls/certs/ca-bundle.crt`
3. `/etc/ssl/ca-bundle.pem`
4. `/etc/pki/tls/cacert.pem`
5. `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`
6. `/etc/ssl/cert.pem`

On UBI10, (1) is usually missing so (2) wins. That file is typically a symlink to (5).

**2. Directory scan — still runs.** After the file loop it walks **both** `/etc/ssl/certs` and `/etc/pki/tls/certs` and appends every regular file (`readUniqueDirectoryEntries` only drops same-dir hashed `.0` symlinks so those certs are not parsed twice). `SSL_CERT_DIR`, if set, *replaces* that directory list; it does not turn the scan off.

So `ca-custom-bundle.crt` is never a default bundle path. The only way Go uses it is if that file actually shows up as a regular file in one of those two directories.

There is a recent performance change ([golang/go#80374](https://github.com/golang/go/pull/80374)) that *defers* the default directory scan until a verify lookup misses. That is still not “file found ⇒ directories ignored”: extra PEMs in the dir are still eligible, just lazily. Released Go that still scans eagerly will load them at startup.

Where this still does not help the extra-file mount: if kubelet `subPath` + UBI’s `/etc/ssl/certs` symlink means the file is **not** visible to `ReadDir` on those paths, Go never sees it and only the primary bundle counts. Overlaying `tls-ca-bundle.pem` is then the reliable way to change what Go reads as that first-hit file (via the `ca-bundle.crt` symlink). `SSL_CERT_DIR` would also change the dir list, including on platforms we do not want to special-case here.

If we have confirmed the extra-file is not visible in the image, I am fine keeping the remount. Retargeting the ConfigMap for trust-manager is still required either way.

_Gal's Cursor_

filariow: I couldn't validate the assumption that the directory scan is still happening if a file was matched locally. We should likely introduce a e2e test for this case.

avi-biton:
@gbenhaim @filariow 
I updated the PR to match Gal's suggestion and also updated the PR description to match it.
Please take a look :)

operator/api/v1alpha1/common_types.go
```diff
@@ -58,6 +58,32 @@ const (
...
+//
+// This type is intentionally defined in common_types.go for reuse across
+// component specs (e.g. image-controller, integration-service) that have
+// similar CA mount patterns. When adding trustedCA support to another
+// component, embed a *TrustedCAConfigMap field in its config spec and apply
+// the mount in that component's deployment overlay. Name and Key are validated
+// by the CRD schema; the overlay must also reject keys whose filepath.Base
+// does not equal the key (path separators, "." and ".."), matching the
+// build-service applyTrustedCAMount check.
+type TrustedCAConfigMap struct {
```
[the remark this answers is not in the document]
avi-biton: Its an infrastructure for later.
Dismissed

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +452,111 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+	if v, ok := cm.BinaryData[key]; ok {
+		return contenthash.String(string(v))
+	}
+	return ""
+}
+
+// mapTrustedCAConfigMap enqueues the singleton KonfluxBuildService when the
+// ConfigMap named in spec.trustedCA changes. The object is user-owned, so
+// Owns(&ConfigMap{}) does not see it.
+func (r *KonfluxBuildServiceReconciler) mapTrustedCAConfigMap(ctx context.Context, obj client.Object) []reconcile.Request {
```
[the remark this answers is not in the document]
avi-biton: Semantically, dismiss.

operator/config/crd/bases/konflux.konflux-ci.dev_konfluxbuildservices.yaml
```diff
@@ -398,6 +398,40 @@ spec:
...
+                  so the pod will not start until the ConfigMap exists. The mount path
+                  is unchanged.
+                properties:
+                  key:
+                    description: |-
+                      Key is the key within the ConfigMap that contains the CA bundle in PEM format.
+                      Must be a valid ConfigMap key: alphanumeric characters, '-', '_' or '.', max 253 chars.
+                    maxLength: 253
+                    minLength: 1
+                    pattern: ^[-._a-zA-Z0-9]+$
```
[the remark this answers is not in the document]
avi-biton: Added a CELL rule

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
```diff
@@ -418,6 +451,104 @@ func buildBuildControllerManagerOverlay(spec konfluxv1alpha1.KonfluxBuildService
...
+		return "", nil
+	}
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(ctx, types.NamespacedName{Name: spec.Name, Namespace: webhookConfigNamespace}, cm); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	return trustedCABundleHash(cm, spec.Key), nil
```
gbenhaim, line 549: Returning `("", nil)` on `NotFound` (and `trustedCABundleHash` doing the same for a missing key) is not equivalent to a Get error that aborts apply.

`applyTrustedCAMount` deletes the hash annotation when `contentHash` is empty. Production objects are a fresh copy of the embedded manifests, so SSA then **prunes** the live `konflux.konflux-ci.dev/trusted-ca-content-hash`. The pod template changes → Deployment rolls → running pods are replaced → new pods stay Pending because the volume is `optional: false`.

That is the opposite of the docs (“ConfigMap deleted after successful deploy: The pod continues running with the previously mounted data”). A kubelet ConfigMap volume would keep the old `subPath` data on existing pods; this path actively kills them.

The comment below is right that a *non-NotFound* Get must not apply an empty hash. Missing CM/key after a successful mount should do the same: keep the live annotation (copy it from the current Deployment) and still retarget the volume as `optional: false` so *new* pods wait.

The unit test that expects an empty hash to strip the leftover annotation, and the missing-CM envtest that asserts the annotation is absent, currently encode prune-on-missing. Those should distinguish “never stamped” from “ConfigMap disappeared after a successful mount.”

— Gal’s Cursor

gbenhaim (commented):
### Follow-up: missing ConfigMap after a successful mount will roll running pods

`lookupTrustedCAHash` treats ConfigMap `NotFound` (and a missing key) as `("", nil)`. `applyTrustedCAMount` then omits `konflux.konflux-ci.dev/trusted-ca-content-hash`. Apply is from a fresh manifest copy, so SSA **prunes** the live annotation, the pod template changes, and the Deployment rolls.

That contradicts the docs table (“pod continues running with the previously mounted data”) and is harsher than a normal ConfigMap volume: kubelet would keep the old mount on existing pods. A trust-manager delete/recreate would bounce build-service.

Non-NotFound `Get` errors already skip apply to avoid this. Missing CM/key should keep the live annotation (read it off the current Deployment) and still retarget `optional: false` for *new* pods.

The empty-hash unit test and the missing-CM envtest (`NotTo(HaveKey(trustedCAHashAnnotation))`) currently lock in prune-on-missing. Worth an envtest that deletes the ConfigMap **after** a hashed mount and asserts the annotation is unchanged.

— Gal’s Cursor

operator/internal/controller/buildservice/konfluxbuildservice_controller.go
avi-biton: Great finding, fixed.

gbenhaim:
### Follow-up (non-blocking): share the overlay primitives, not `applyTrustedCAMount`

This does not need to land in this PR. Build-service and integration-service already ship the same extra-file mount (`trusted-ca` → `/etc/ssl/certs/ca-custom-bundle.crt`, `subPath: ca-bundle.crt`). When integration-service (and later consumers) grow a `spec.trustedCA` field, they should compose generic overlay options instead of copying `applyTrustedCAMount`.

`pkg/customization` should stay a generic overlay engine. Keep `lookupTrustedCAHash`, the ConfigMap watch, and “preserve the live hash if the ConfigMap disappears” on each reconciler — those need a client and ApplyOwned semantics.

**Do not change** `WithConfigMapVolumeUpdate`. It is name-only and silent-skips a missing volume (dex, webhook-config, quay-ca). Do not teach it `Items`/`Optional` or fail-loud.

**Do not add** `WithTrustedCAMount(*TrustedCAConfigMap)` here. `FromContainerSpec` is user resources/env; TrustedCA also carries rollout/SSA policy that will drift per consumer.

Suggested API:

```go
// ConfigMapVolumeOverride patches an existing ConfigMap volume in place.
// Unlike WithConfigMapVolumeUpdate, Apply fails if the volume is missing
// or is not a ConfigMap volume.
type ConfigMapVolumeOverride struct {
	Name     string
	Items    []corev1.KeyToPath // replace Items when non-nil
	Optional *bool              // set when non-nil
}

func WithConfigMapVolumeOverride(volumeName string, override ConfigMapVolumeOverride) PodOverlayOption
func WithRequiredVolumeMount(containerName, volumeName string) PodOverlayOption
func WithPodAnnotation(key, value string) PodOverlayOption
func WithoutPodAnnotation(keys ...string) PodOverlayOption
```

Store the override on a **second** `PodOverlay` field (do not overload `configMapVolumeUpdates`). Apply it after the existing name-only updates. `WithPodAnnotation` merges; `WithoutPodAnnotation` is for in-memory leftovers — fresh ApplyOwned copies already omit unset keys, which is what SSA needs on restore.

Consumer wiring (hash still computed by the reconciler and passed in):

```go
if spec.TrustedCA != nil {
    podOpts = append(podOpts,
        customization.WithConfigMapVolumeOverride("trusted-ca", customization.ConfigMapVolumeOverride{
            Name: spec.TrustedCA.Name,
            Items: []corev1.KeyToPath{{
                Key:  spec.TrustedCA.Key,
                Path: common.TrustedCAProjectedPath, // "ca-bundle.crt"
            }},
            Optional: ptr.To(false),
        }),
        customization.WithRequiredVolumeMount(managerContainerName, "trusted-ca"),
    )
    if hash != "" {
        podOpts = append(podOpts,
            customization.WithPodAnnotation(common.TrustedCAHashAnnotation, hash),
        )
    }
}
```

Shared constants (`trusted-ca`, projected path, annotation key) belong in `operator/internal/common` next to `TrustedCAConfigMapName`, not in `pkg/customization`. Image-controller’s quay CA path is a different volume/env pattern and should stay local.

Happy to take this in a follow-up PR.

— Gal’s Cursor

gbenhaim (approved):

PR #9198, opened 2026-09-15
feat(KONFLUX-9075): enable namespace-lister HTTPS metrics endpoint

yftacherzog:
Wire KonfluxNamespaceLister.spec.componentMetrics to deployment -enable-metrics toggling and expose :9100 on the Service with metadata.labels for future ServiceMonitor selection. Add namespace-lister-auth-delegator RBAC for authenticated metrics scrape. Envtests verify enable/disable, Service shape, and auth-delegator binding.

Previously we added the API knob; Now we make it affect the workload. Next steps:
- Add ServiceMonitor, scrape token reconciler, metrics-scraper RBAC, and UWM NetworkPolicy, with Kind metricsintegration coverage.
- Add OCP UWM CI and component-monitoring docs.
- The listener and Service are wired in this PR; scrape is inert until we later add NetworkPolicy for :9100.

Assisted-by: Cursor


operator/upstream-kustomizations/namespace-lister/service.yaml
```diff
@@ -11,3 +13,6 @@ spec:
   - name: http
     targetPort: 8080
     port: 8080
+  - name: metrics
+    targetPort: 9100
+    port: 9100
```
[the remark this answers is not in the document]
yftacherzog: Planned to be addressed in the next PRs.

operator/upstream-kustomizations/namespace-lister/rbac.yaml
```diff
@@ -34,3 +34,35 @@ rules:
   - roles
   - rolebindings
   verbs: ["get", "list", "watch"]
+---
```
avi-biton, line 37: Problem: docs/component-monitoring.md ("Repo layout", "Controller wiring checklist" step 5/9) states operator-owned metrics resources live in <component>/monitoring/, are skipped when componentMetrics.enabled: false, and are registered for orphan cleanup. This PR puts the metrics RBAC in the always-applied core set. With metrics disabled, the namespace-lister SA still holds cluster-wide create tokenreviews / create subjectaccessreviews — a privilege it has no use for — and toggling off never reclaims it. Note kubernetes.IsComponentMetricsScrapeResource would not match these names anyway (it matches the -metrics-reader suffix), so gating needs explicit handling rather than the shared helper.
Fix: Either move the two objects to a namespace-lister/monitoring/ overlay, gate apply on spec.ComponentMetrics.IsEnabled(), and add ClusterRole/ClusterRoleBinding to NamespaceListerCleanupGVKs + NamespaceListerClusterScopedAllowList; or state in the PR/commit why the auth-delegator is intentionally core (e.g. avoiding a 401-loop during toggle races) and leave a comment in rbac.yaml.

_Avi's Claude_

avi-biton:
Problem: 
`namespace-lister-allow-from-konfluxui` is the only ingress NetworkPolicy on podSelector: apps=namespace-lister, and it permits port 8080 only. Nothing can reach :9100 today, so the endpoint is enabled but unscrapable. The PR description does list the UWM NetworkPolicy as a follow-up, so this is a scope decision rather than a defect — worth stating in the PR body that the feature is inert until the follow-up lands.
Fix: 
No change required if the incremental plan is accepted; consider making the deferral explicit in the commit message.

_Avi's Claude_

operator/internal/controller/namespacelister/konfluxnamespacelister_controller_test.go
```diff
@@ -352,6 +445,36 @@ var _ = Describe("KonfluxNamespaceLister Controller", func() {
...
+				ObjectMeta: metav1.ObjectMeta{Name: crbNN.Name},
+			})).To(Succeed())
+
+			By("verifying the metrics-auth ClusterRoleBinding is recreated with ownership labels")
+			Eventually(func(g Gomega) {
+				crb := &rbacv1.ClusterRoleBinding{}
+				g.Expect(k8sClient.Get(ctx, crbNN, crb)).To(Succeed())
+				g.Expect(crb.Labels).To(HaveKey(constant.KonfluxOwnerLabel))
+				g.Expect(crb.RoleRef.Name).To(Equal(metricsAuthRoleName))
+			}).WithTimeout(testutil.EventuallyTimeout).WithPolling(testutil.EventuallyPolling).Should(Succeed())
```
avi-biton, line 517: **Test gap: the metrics-auth ClusterRole is never asserted**

This spec verifies the ClusterRoleBinding — including `roleRef.Name == namespace-lister-metrics-auth-role` — but nothing ever `Get`s the ClusterRole itself, so a dangling `roleRef` would pass. More importantly, the two rules the metrics auth filter depends on are unpinned.

Upstream namespace-lister builds its metrics server with `FilterProvider: filters.WithAuthenticationAndAuthorization` ([`main.go`](https://github.com/konflux-ci/namespace-lister/blob/main/main.go)), which calls TokenReview + SubjectAccessReview on every scrape. If `create` on `tokenreviews` (`authentication.k8s.io`) or `subjectaccessreviews` (`authorization.k8s.io`) is accidentally dropped from `upstream-kustomizations/namespace-lister/rbac.yaml`, the manifests still render, the deployment still starts, and every scrape fails with a 500 — no test catches it.

Suggested addition after line 476:

```go
By("verifying the metrics-auth ClusterRole grants the auth-delegator rules")
Eventually(func(g Gomega) {
    cr := &rbacv1.ClusterRole{}
    g.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: metricsAuthRoleName}, cr)).To(Succeed())
    g.Expect(cr.Rules).To(ContainElement(rbacv1.PolicyRule{
        APIGroups: []string{"authentication.k8s.io"},
        Resources: []string{"tokenreviews"},
        Verbs:     []string{"create"},
    }))
    g.Expect(cr.Rules).To(ContainElement(rbacv1.PolicyRule{
        APIGroups: []string{"authorization.k8s.io"},
        Resources: []string{"subjectaccessreviews"},
        Verbs:     []string{"create"},
    }))

Related, and worth a second spec: the metrics RBAC and the Service :9100 port are applied unconditionally, independent of componentMetrics. That looks deliberate, but no test locks it in — disables metrics listener when componentMetrics is disabled (L149) only checks container args, and exposes metrics port and selector label on Service (L176) only runs with metrics enabled. A spec asserting both survive enabled: false would pin the intent either way.
```

_Avi's Claude_

yftacherzog:
> Problem: `namespace-lister-allow-from-konfluxui` is the only ingress NetworkPolicy on podSelector: apps=namespace-lister, and it permits port 8080 only. Nothing can reach :9100 today, so the endpoint is enabled but unscrapable. The PR description does list the UWM NetworkPolicy as a follow-up, so this is a scope decision rather than a defect — worth stating in the PR body that the feature is inert until the follow-up lands. Fix: No change required if the incremental plan is accepted; consider making the deferral explicit in the commit message.
> 
> _Avi's Claude_

I updated the PR description to include it as next step

operator/upstream-kustomizations/namespace-lister/rbac.yaml
yftacherzog: *-metrics-auth-role is in core, same as in build-service/image-controller/etc.

Only scrape-contract resources (monitoring/ + IsComponentMetricsScrapeResource) are gated.

Next PRs will add the monitoring/ overlay (ServiceMonitor, metrics-reader, NP) and reconciler skip/cleanup for those.

operator/internal/controller/namespacelister/konfluxnamespacelister_controller_test.go
yftacherzog: added

avi-biton (approved):

PR #9244, opened 2026-09-17
chore(AGENTS.md): slim always-on AGENTS.md and move playbooks into skills

avi-biton:
AGENTS.md is injected on every agent turn. It mixed repo-wide rules with task playbooks (Kind/e2e setup, CI workflow catalog, CRD test recipes, Makefile encyclopedia, skills table). That billed the same large prefix whether the session was a one-file controller edit or an e2e debug.

Keep in AGENTS.md only what applies to most sessions: modules and `cd operator`, code style, controller-runtime invariants, package-local lint/vet, compressed test/RBAC notes, and pointers to skills. Add a short session-hygiene section (atomic changes, `/compact` or new chat, load only the matching skill).

The file is shorter, but the CI cap is unchanged:
`MAX_AGENTS_MD_LINES` in `.github/workflows/validate-agents-md.yaml` is still **300**. This rewrite does not lock the file at the new length; it can grow again up to that limit.

Move the playbooks into the skills that already own those tasks so behavior stays the same when that work starts:

- `ginkgo-testing`: CRD self-healing/drift conventions (labels, cleanup, `c.kind` vs `c.name`)
- `create-pr`: CI workflow list, golangci-lint single-semver-line rule, `make manifests generate` on the pre-PR checklist
- `local-dev-setup`: copy/source `test/e2e/e2e.env` and run `run-e2e.sh`

Token counts are tiktoken `cl100k` vs `HEAD`.

**Always-on (every turn): `AGENTS.md`**

| | Before | After | Change |
|---|---|---|---|
| Lines | 181 | 68 | -62% |
| Tokens | 3511 | 1450 | **-2061 (-59%)** |

This is a large cut to always-on *repo* context, and it hits every turn of every session. It is not an order-of-magnitude drop in total agent spend: a full turn still includes tool schemas, user rules, the skill catalog, and chat history. Against that total, ~2k tokens per turn is a noticeable slice of the prefix, not a 50% bill cut. Cached later turns are cheaper per token than the uncached-equivalent figures below.

A 20-turn session saves about 41k uncached-equivalent input tokens from `AGENTS.md` alone.

**On invoke only (skill body when that task runs)**

| File | Before | After | Change |
|---|---|---|---|
| `skills/ginkgo-testing/SKILL.md` | 931 | 1120 | +189 | | `skills/create-pr/SKILL.md` | 637 | 890 | +253 | | `skills/local-dev-setup/SKILL.md` | 2192 | 2293 | +101 |

Those extras are not in general sessions. A Ginkgo or create-PR session is still far cheaper than keeping the old encyclopedias in `AGENTS.md` on every turn. The always-on skill catalog was not shortened.

Assisted-by Cursor + Grok 4.6


AGENTS.md
```diff
@@ -1,181 +1,68 @@
...
-| [ginkgo-testing](skills/ginkgo-testing/SKILL.md) | Writing or reviewing Ginkgo tests — cleanup patterns, soft assertions |
-| [go-toolchain-upgrade](skills/go-toolchain-upgrade/SKILL.md) | `go.mod`/`go.sum`, Go pins, or `go.mod requires go` CI failures |
-| [pr-review](skills/pr-review/SKILL.md) | Reviewing human-authored PRs — upstream/downstream hygiene and related checks |
-| [create-pr](skills/create-pr/SKILL.md) | Opening PRs, fork `/allow <commit-sha>` behavior |
-| [debug-e2e-tests](skills/debug-e2e-tests/SKILL.md) | Triaging or investigating failed e2e / OpenShift CI runs |
-| [update-upstream-deps](skills/update-upstream-deps/SKILL.md) | Bumping upstream SHAs or editing `upstream-kustomizations/` (triggers manifest rebuild) |
-| [companion-pr-review](skills/companion-pr-review/SKILL.md) | MintMaker/Renovate parent or companion PRs — what to skip vs review lightly |
-| [local-dev-setup](skills/local-dev-setup/SKILL.md) | Local Kind / dev environment |
-| [dev-verify-loop](skills/dev-verify-loop/SKILL.md) | Iterative stop-rebuild-restart cycle for operator development |
+- Human PR review: `skills/pr-review/SKILL.md` (upstream/downstream hygiene).
```
yftacherzog: Should we load this just for human PRs? Suppose agents (e.g. fullsend) open a PR, I think it should be reviewed in the same way. No?

avi-biton: Updated both the SKILL and AGENTS

kelchen123 (approved):

PR #9245, opened 2026-09-17
fix(KONFLUX-15639): Stop applying Dex demo passwords on OpenShift

avi-biton:
FIND-004: the OpenShift install path applied Dex static users (user1@konflux.dev / user2@konflux.dev) whose bcrypt hash is the well-known string "password". On an internet-reachable cluster that grants tenant access with a published credential.

The ticket asked to keep staticPasswords behind a Kind-only kustomize overlay under dependencies/dex/ and set enablePasswordDB: false in the OCP overlay. That pattern is no longer valid: dependencies/dex was removed and Dex is configured from the Konflux CR, so there is no Dex kustomize overlay to split.

This change applies the same control at the CR layer instead:

- Add konflux-openshift.yaml with enablePasswordDB: false and no staticPasswords. OpenShift login stays the operator default.
- Use that CR as the default for human ./deploy-konflux-on-ocp.sh runs.
- Refuse CRs that contain staticPasswords: unless the caller sets ALLOW_DEV_STATIC_PASSWORDS=true or OpenShift CI sets OPENSHIFT_CI=true (Prow still uses the Kind/e2e samples).
- Document the OpenShift sample and that Kind/CI demo users must not be used on reachable clusters.

Kind and GitHub Actions are unchanged and still use the demo-user CRs.

Assisted-by Cursor + Grok 4.6


avi-biton:
Closing this PR.
This PR https://github.com/konflux-ci/konflux-ci/pull/9279 replaces it.

